PR #12, opened 2017-02-27
add advisory for hyper message splitting vulnerability

seanmonstar:


tarcieri:
Assigned RUSTSEC-2017-0002

PR #18, opened 2017-03-25
Advisory: hyper HTTPS MitM due to lack of hostname verification

tarcieri:
cc @seanmonstar

### DIFF
--- a/crates/hyper/RUSTSEC-0000-0000.toml
+++ b/crates/hyper/RUSTSEC-0000-0000.toml
@@ -1,0 +1,18 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "hyper"
+patched_versions = [">= 0.9.4"]
+references = ["RUSTSEC-2016-0001"]
+date = "2016-05-09"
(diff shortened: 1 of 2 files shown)


tarcieri:
@seanmonstar do the contents of this advisory look correct to you?

seanmonstar:
Hostname validation was never enabled on Windows (or enabled for 1 point version, and then disabled the next) since when enabled it couldn't validate anything. With 0.10, hyper left TLS for external libraries.

tarcieri:
Hmm, there is currently no way to express platform-specific constraints, although that'd be a nice feature.

For now we can just add that to the description. Does that sound good?

seanmonstar:
Sure!

erickt:
Hello! Should this old PR be closed or merged in?

tarcieri:
@erickt I was just looking at it! I was just updating the description to note it's Windows-only

tarcieri:
@seanmonstar mind taking a look? probably should've asked before merging, but...

PR #19, opened 2017-03-25
Advisory: rust-openssl hostname verification

tarcieri:
cc @sfackler

### DIFF
--- a/crates/openssl/RUSTSEC-0000-0000.toml
+++ b/crates/openssl/RUSTSEC-0000-0000.toml
@@ -1,0 +1,20 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "openssl"
+patched_versions = [">= 0.9.0"]
+date = "2016-11-05"
+url = "https://github.com/sfackler/rust-openssl/releases/tag/v0.9.0"
+title = "SSL/TLS MitM vulnerability due to insecure defaults"
+description = """
+All versions of rust-openssl prior to 0.9.0 contained numerous insecure defaults
+including off-by-default certificate verification and no API to perform hostname
+verification.
+
+Unless configured correctly by a developer, these defaults could allow an attacker
+to perform man-in-the-middle attacks.
+
+The problem was addressed in newer versions by enabling certificate verification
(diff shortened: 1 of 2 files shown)


tarcieri:
@sfackler do the contents of this advisory look correct to you?

sfackler:
The version number seems wrong - 1.0.2 was when hostname verification support was added to OpenSSL proper. The rust-openssl 0.9.0 release had both the bindings in that PR as well as a handrolled implementation when linking against 1.0.1 which were set up when using the `SslConnector` type which was added in that release as well.

sfackler:
The separate [openssl-verify](https://crates.io/crates/openssl-verify) crate provided the handrolled implementation in a way compatible with rust-openssl 0.7 and 0.8, which Hyper used before its 0.10 release.

tarcieri:
@sfackler should it be this then?

```toml
patched_versions = [">= 0.9.0"]
```

sfackler:
I think it might make sense to make the advisory more generally about the lack of good defaults before 0.9 - hostname verification is part of that, but so is enabling certificate verification in the first place, limiting cipher suites, setting up forward secrecy on the server side, etc.

tarcieri:
@sfackler can you take another look at this? I tried to update it per your suggestions

crates/openssl/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,18 @@
...
+description = """
+All versions of rust-openssl prior to 0.9.0 contained numerous insecure defaults
+including off-by-default certificate verification and no API to perform hostname
+verification.
+
+Unless configured correctly by a developer, these defaults could allow an attacker
+to perform man-in-the-middle attacks.
+
+The problem was addressed in newer versions by enabling certificate verification
+by default and exposing APIs to perform hostname verification.
```
sfackler: It's probably worth noting here that the good defaults are present when you use the `SslConnector` and `SslAcceptor` types, not the more low-level `SslContext` type.

tarcieri: Added, thanks!

PR #21, opened 2017-05-04
Advisory: base64 heap-based buffer overflow

AGWA:


frewsxcv:
@AGWA out of curiosity, are you willing to reveal how you discovered the buffer overflow?

AGWA:
@frewsxcv Just by looking at the code. I saw that base64 was using unsafe code to write into a buffer, so I went back to see how the buffer was being allocated, and saw that integer overflow was not being handled when calculating the buffer size.

tarcieri:
Assigned RUSTSEC-2017-0004 in 524d876a

PR #22, opened 2017-05-07
Advisory: cookie denial of service

erickt:
cc @SergioBenitez and @alexcrichton


tarcieri:
Assigned RUSTSEC-2017-0005 in fafc60c

PR #24, opened 2018-06-25
Advisory: An integer underflow in untrusted 0.6.1 and older which could lead to panic

oherrala:
Ping @briansmith, the author and maintainer of untrusted who has seen this advisory before submission.

Short description:

> A mistake in error handling in untrusted before 0.6.2 could lead to an integer underflow and panic if a user of the crate didn't properly check for errors returned by untrusted.
>
> Combination of these two programming errors (one in untrusted and another by user of this crate) could lead to a panic and maybe a denial of service of affected software.
>
> The error in untrusted is fixed in release 0.6.2 released 2018-06-21. It's also advisable that users of untrusted check for their sources for cases where errors returned by untrusted are not handled correctly.

### DIFF
--- a/crates/untrusted/RUSTSEC-0000-0000.toml
+++ b/crates/untrusted/RUSTSEC-0000-0000.toml
@@ -1,0 +1,41 @@
+[advisory]
+package = "untrusted"
+
+# Versions which were never vulnerable
+unaffected_versions = []
+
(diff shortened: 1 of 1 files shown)


tarcieri:
Assigned RUSTSEC-2018-0001 in 3c0458d

PR #26, opened 2018-06-29
Aribtrary filesystem writes in tar 0.4.15 and older

alexcrichton:
More details inside!

### DIFF
--- a/crates/tar/RUSTSEC-2018-0002.toml
+++ b/crates/tar/RUSTSEC-2018-0002.toml
@@ -1,0 +1,25 @@
+[[advisory]]
+id = "RUSTSEC-2018-0002"
+package = "tar"
+unaffected_versions = []
+patched_versions = [">= 0.4.16"]
+dwf = []
+url = "https://github.com/alexcrichton/tar-rs/pull/156"
+title = "Links in archives can overwrite any existing file"
+date = "2018-06-29"
+description = """
+When unpacking a tarball with the `unpack_in`-family of functions it's intended
+that only files within the specified directory are able to be written. Tarballs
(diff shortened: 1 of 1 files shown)


crates/tar/RUSTSEC-2018-0002.toml
```diff
@@ -0,0 +1,25 @@
...
+with hard links or symlinks, however, can be used to overwrite any file on the
+filesystem.
+
+Tarballs can contain multiple entries for the same file. A tarball which first
+contains an entry for a hard link or symlink pointing to any file on the
+filesystem will have the link created, and then afterwards if the same file is
+listed in the tarball the hard link will be rewritten and any file can be
+rewritten on the filesystem.
+
+This has been fixed in https://github.com/alexcrichton/tar/pull/156 and is
```
niklasf: s/tar/tar-rs/

tirkarthi:
I think this is towards the snyk report. Relevant zip-rs issue : https://github.com/mvdnes/zip-rs/issues/74

Ref : https://github.com/snyk/zip-slip-vulnerability

tarcieri:
@greysteil yes, however given the current way the `rustsec-client` crate works, the rest of the advisories are combined into `Advisories.toml` in the root of the repo, which is an array of advisories. The other TOML files aren't consumed directly.

I'm traveling right now but I'll fix it up and get it into `Advisories.toml` shortly.

(As a sidebar I should probably just get rid of `Advisories.toml` and consume this repo directly via a git crate)

greysteil:
👍 - I'm pulling in the individual files in Dependabot, but have updated my processing to handle formatting like this, too.

Enjoy your travels!

PR #54, opened 2018-08-25
Add advisory for Claxon 0.3.2 and 0.4.1

ruuda:
I just published Claxon 0.3.2 and 0.4.1 which fix https://github.com/ruuda/claxon/issues/10.


tarcieri:
Assigned RUSTSEC-2018-0004 in 575dc9a

Issue #25, opened 2018-06-27
Missing documentation


Shnatsel:
By looking at this repository I do not see any answers to the following questions:

1. How do I file an advisory? (Presumably by opening a PR, but it's not stated)
1. Under what conditions should I file an advisory? Is a panic eligible for an advisory? Memory leak? Math that calculates buffer sizes incorrectly but is not dangerous unless fed to `unsafe` by API consumer? I've already found all of these.
1. What do I do if the crate maintainers are not responding / fixing the issues? My experience with filing bugs has ranged from [best response I've seen yet](https://github.com/PistonDevelopers/inflate/pull/43#discussion_r197676499) to [pretty much ignored](https://github.com/RustAudio/lewton/issues/27).

The reason I'm asking is that I've spent the last several days fuzzing various popular library crates, so now I'm sitting on a pile of bugs, wondering what to do with them aside of fixing them and opening PRs in absence of action from maintainers.

Also, somewhat out of scope of this issue:

 - I have looked for a place to report security issues in Rust packages on crates.io and found nothing of use. There is [security page for Rust itself](https://www.rust-lang.org/en-US/security.html) that does not mention anything about crates.io packages. If I'd found something more serious than panics or OOMs that I got so far, I would not be able to report it _anywhere,_ as my experience with CVEs so far can be summed up as "don't even bother".
 - I have failed to find a guide for shipping security updates on crates.io. Do I yank the vulnerable versions? How do I announce that a certain update is a security update? What other steps do I need to take?
 - This is the first time I hear about cargo-audit, despite being security-inclined. Cargo quickstart guides did not mention it. Neither did the Nomicon mention any kind of actions to take after you've found and fixed a memory safety issue. The folks over at [rust-fuzz](https://github.com/rust-fuzz/trophy-case) make no mention of this either.
 - I kind of expected cargo to notify me about security updates for dependencies of my crate without me invoking obscure tools. It's not like I can run `cargo audit` on everything twice a day, even if I'd known about it. Yet the [readme for it](https://github.com/rustsec/cargo-audit) says it doesn't even ship with cargo by default?
 - As an aside, I hope something like email updates is set up for maintainers of crates on crates.io shipping with cargo.lock pointing to vulnerable versions and cargo.toml to vulnerable series, but I am not one of them, so I wouldn't know.

tarcieri:
Hi there! Some great questions, thanks for opening this issue.

[...]

PR #59, opened 2018-09-07
Denial of service with debug build of http-codec

niklasf:
This is mostly a case of
> When in doubt, please open a PR.

It's a minor issue, because only debug builds and builds with `-C overflow-checks=on` are affected. One reason to file an advisory would be that this is unlikely to get fixed: I contacted the author and he says he is no longer maintaining the crate.

### DIFF
--- a/crates/http-codec/RUSTSEC-0000-0000.toml
+++ b/crates/http-codec/RUSTSEC-0000-0000.toml
@@ -1,0 +1,18 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "http-codec"
+date = "2018-09-07"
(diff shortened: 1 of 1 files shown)


tarcieri:
Hi there, sorry for the delay reviewing this.

It sounds like a panic for debug builds and integer overflow for non-debug builds. Either way that does not seem good. I think this is worth including.

Shnatsel:
As per the criteria discussed [here](https://github.com/RustSec/advisory-db/issues/25#issuecomment-401108454) I'd rather not include otherwise harmless overflows and debug-mode panics, at least until a severity system is in place.

tarcieri:
Alright, I can agree with that

PR #60, opened 2018-09-17
Add advisory for yaml-rust

xfix:
Not sure if an abort is a valid vulnerability, but for me crashing the entire application with no possibility of recovery (it's abort, not a panic) sounds like denial of service.


tarcieri:
Assigned `RUSTSEC-2018-0006` in #64

PR #61, opened 2018-09-17
Add advisory for serde_yaml

xfix:
Not sure if an abort is a valid vulnerability, but for me crashing the entire application with no possibility of recovery (it's abort, not a panic) sounds like denial of service.

This is separate from #60, this is about YAML aliases, not very deep data structures. That said, #60 was discovered while trying to fix that one.


tarcieri:
Assigned `RUSTSEC-2018-0006` in #63

Issue #69, opened 2018-12-06
Unclear which keywords are available or one should use


gnzlbg:

tarcieri:
The keywords are intended to be freeform ala `Cargo.toml`, however the advisories could probably also benefit from some "categories" which have a fixed schema, perhaps mapping to the existing labels:

https://github.com/RustSec/advisory-db/labels

gnzlbg:
It would be nice if there was some list of keywords to choose from. I honestly have no idea which keywords to use.

PR #70, opened 2018-12-06
Add advisory for slice-deque

gnzlbg:


crates/slice_deque/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,63 @@
+[advisory]
+# Identifier for the advisory (mandatory). Will be assigned a "RUSTSEC-YYYY-NNNN"
+# identifier e.g. RUSTSEC-2018-0001. Please use "RUSTSEC-0000-0000" in PRs.
```
tarcieri: Can you remove the comments? Thanks.

gnzlbg: Done.

tarcieri:
Per #69, for some keyword precedent, there's this (`memory-exposure`): https://github.com/RustSec/advisory-db/blob/master/crates/smallvec/RUSTSEC-2018-0003.toml#L7

I'd probably suggest `memory-corruption` and `rce` as keywords here if you're interested in adding some. I'd originally wanted to surface these in https://rustsec.org but haven't had time yet.

gnzlbg:
@tarcieri I've added those two keywords (hope I did that correctly)

crates/slice_deque/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,25 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "slice_deque"
```
tarcieri: I noticed this a bit too late: RustSec goes by the "package" name (i.e. what appears on crates.io and `Cargo.lock`) as opposed to the crate name.

Looking at your Cargo.toml, the package name appears to be `slice-deque` (hyphen, not underscore):

https://github.com/gnzlbg/slice_deque/blob/master/Cargo.toml

I can fix this up for you when I assign the RUSTSEC identifier for this advisory.

gnzlbg: oh I missed that!

> I can fix this up for you when I assign the RUSTSEC identifier for this advisory.

as you wish, I can also send a PR that fixes this !

tarcieri: #71 fixes it

PR #73, opened 2018-12-06
Verify package names are correct in CI

### LINKED ISSUE #72: Add CI which validates that the package name is a valid package
Happened in #70 -- the package didn't match the one on crates.io

alex:
Fixes #72


Cargo.toml
```diff
@@ -16,3 +16,6 @@ name = "rustsec-advisory-db"
 gumdrop = "0.4"
 gumdrop_derive = "0.4"
 rustsec = "0.9"
+reqwest = "0.9"
+serde = "1.0"
+serde_derive = "1.0"
```
tarcieri: I saw there's also this crate, which uses the same dependencies as above:

https://crates.io/crates/crates_io_api

Might be a case of YAGNI but I can see potential other cases where we might want to lint things in advisories based on the current state of crates.io (e.g. do version ranges actually match any releases? should crates which match those ranges always be yanked?)

alex: I'm happy to switch, would you prefer that?

tarcieri: If you don't mind, I think that'd be nice, and I'm not too worried about the extra dependency

PR #75, opened 2018-12-09
A bug in crossbeam v0.4.0

stjepang:


Shnatsel:
According to the version specification here 0.3.2 is vulnerable, and that's the most popular version right now according to crates.io download statistics: https://crates.io/crates/crossbeam

Is 0.3.2 really affected?

crates/crossbeam/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,56 @@
+[advisory]
+# Identifier for the advisory (mandatory). Will be assigned a "RUSTSEC-YYYY-NNNN"
```
tarcieri: Can you please remove the comments? Thanks!

stjepang: Done.

crates/crossbeam/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,19 @@
...
+
+The flaw was corrected by wrapping elements inside queues in a
+`ManuallyDrop`.
+
+Thanks to @c0gent for reporting the issue.
+"""
+patched_versions = [">= 0.4.1"]
+unaffected_versions = ["< 0.4.0"]
+url = "https://github.com/crossbeam-rs/crossbeam-epoch/issues/82"
+keywords = ["concurrency", "memory-management"]
```
tarcieri: Though these keywords are freeform, there are a few other advisories tagged `memory-corruption` so I might advise that one here.

tarcieri:
Looks good, thanks!

tarcieri:
Assigned `RUSTSEC-2018-0009` in #76

PR #77, opened 2018-12-11
Request RUSTSEC for resolved UAF in OpenSSL

alex:


tarcieri:
This looks good but I'm going to try to address https://github.com/RustSec/rustsec-crate/issues/51 tonight. Otherwise I think this will be pretty disruptive.

alex:
@tarcieri Is this r4r now that that's been fixed?

tarcieri:
I need to release the `rustsec` crate, but I can do that shortly here

tarcieri:
All right, sorry that took so long, but I think we're good to go on this now

tarcieri:
Assigned `RUSTSEC-2018-0010` in #79

Issue #68, opened 2018-11-28
Collecting metadata for impacted functions in advisories


tarcieri:
RustPräzi is a crater-like tool which builds a call graph across all of the crates published to crates.io:

https://internals.rust-lang.org/t/prototype-dev-tool-rustprazi-a-tool-to-build-an-entire-call-graph-of-crates-io/8912

It would be interesting to use this tool, perhaps in conjunction with [crates-audit](https://gitlab.com/zachreizner/crates-audit), to generate a list of impacted crates based on this call graph.

To do that we'd need to collect structured information about vulnerable functions in advisories. Looking at our existing advisories, that information is often already there, but buried in an unstructured prose description, e.g.:

https://rustsec.org/advisories/RUSTSEC-2018-0003

> If an iterator passed to `SmallVec::insert_many` panicked in `Iterator::next`, destructors were run during unwinding while the vector was in an inconsistent state, possibly causing a double free (a destructor running on two copies of the same value).

I think it'd be good to come up with a structured format for this sort of information that can be fed into tools like RustPräzi to determine impacted users of vulnerable crates.

Inventitech:
Sorry to chime in on this after it's been merged.

One problem with the current design is that it does not allow for the specification of a version.

**Why could this be important?**
Some security bugs affect quite a broad change of versions of a crate, and as such, the vulnerability might itself be subject to refactoring, renames, etc. So just specifying a name in `affected_functions` is perhaps overly simplistic.

Example

v1 introduce vulnerable function `f1`
v2 rename `f1` to  `f2`
v3 introduce new function `f1`, which is not vulnerable
v4 fix vulnerability in `f2`

Just naming all affected_functions `f1` and `f2` is not a great solution, because `f1` would be considered vulnerable from v2 to v3, even though we know it isn't.

**What would the change proposal be?**
For each function, add a range of affected versions.
```
affected_functions = [ ["f1","<2"], ["f2",">2 <4"]]
```

**What are the consequences of the proposed change?**
* It makes the declaration of this array more complicated.
* It increases precision and resolves possible ambiguities. I don't know how often this would occur in practice.


Another question would be whether we would want to restrict this to functions, only. Couldn't a data structure have a vulnerability, too?

tarcieri:
Haha, no worries.

Modifying the existing `affected_functions` might be tricky, but we can just abandon it, especially as this is a good reason:

[...]

PR #82, opened 2018-12-20
Add new affected functions attribute to template

Inventitech:
Refs #68


tarcieri:
@Inventitech awesome, thanks!

There's presently no lint on the crate name which would be nice. We're scraping the crates.io API now, which returns the crate name for a given package as part of its results (otherwise it's `-` => `_`  I believe?)

I think we should at least validate the first path segment matches the crate name. That said, I can merge this for now.

cc @alex

Inventitech:
> @Inventitech awesome, thanks!
> 
> crate name for a given package as part of its results (otherwise it's `-` => `_` I believe?)

True!

PR #86, opened 2019-01-14
Update to 'rustsec' crate v0.11

tarcieri:


tarcieri:
FYI @Inventitech, I reverted #83, however this bumps the `rustsec` crate version to use v0.11 which should support the `affected_paths` syntax implemented here:

https://github.com/RustSec/rustsec-crate/pull/58

PR #90, opened 2019-03-03
Updated example in the readme for affected_paths name change

alex:


tarcieri:
@alex it's a bit more than that. This section is now a TOML table which contains a mapping of `VersionReqs` to affected paths. See some discussion on #68.

Having actually used this though, I'm wondering if we still aren't quite in the right place with this feature, but discuss that in #68.

alex:
Ugh, right. Just going to close this until/if I have the time to do it properly.

Issue #20, opened 2017-04-20
File CVEs for all vulnerabilities


tarcieri:
This is a tracking issue for which vulnerabilities have CVEs. We should file a CVE for every vulnerability in our database:

| Advisory Id                      | Request Filed  | CVE Assigned |
|-------------------------|---------------|---------------------------|
| [RUSTSEC-2016-0001] | ⛔️                    | ⛔️                                       |
| [RUSTSEC-2016-0002] | ⛔️                    | ⛔️                                       |
| [RUSTSEC-2017-0001] | ✅                    | ✅ [CVE-2017-1000168]  |
| [RUSTSEC-2017-0002] | ⛔️                    | ⛔️                                       |
| [RUSTSEC-2017-0003] | ⛔️                    | ⛔️                                       |
| [RUSTSEC-2017-0004] | ✅                    | ✅ [CVE-2017-1000430] |
| [RUSTSEC-2017-0005] | ⛔️                    | ⛔️                                       |
| [RUSTSEC-2018-0001] | ⛔️                    | ⛔️                                       |
| [RUSTSEC-2018-0002] | ⛔️                    | ⛔️                                       |
| [RUSTSEC-2018-0003] | ⛔️                    | ⛔️                                       |
| [RUSTSEC-2018-0004] | ⛔️                    | ⛔️                                       |

[RUSTSEC-2016-0001]: https://rustsec.org/advisories/RUSTSEC-2016-0001.html
[RUSTSEC-2016-0002]: https://rustsec.org/advisories/RUSTSEC-2016-0002.html
[RUSTSEC-2017-0001]: https://rustsec.org/advisories/RUSTSEC-2017-0001.html
[RUSTSEC-2017-0002]: https://rustsec.org/advisories/RUSTSEC-2017-0002.html
[RUSTSEC-2017-0003]: https://rustsec.org/advisories/RUSTSEC-2017-0003.html
[RUSTSEC-2017-0004]: https://rustsec.org/advisories/RUSTSEC-2017-0004.html
[RUSTSEC-2017-0005]: https://rustsec.org/advisories/RUSTSEC-2017-0005.html
[RUSTSEC-2018-0001]: https://rustsec.org/advisories/RUSTSEC-2018-0001.html
[RUSTSEC-2018-0002]: https://rustsec.org/advisories/RUSTSEC-2018-0002.html
[RUSTSEC-2018-0003]: https://rustsec.org/advisories/RUSTSEC-2018-0003.html
[...]

PR #93, opened 2019-04-28
Add advisory for ammonia

xfix:


tarcieri:
Curious what @alex and @Shnatsel think about this.

Though we usually avoid advisories for DoS, there were soundness issues around stack overflow specifically (and I think still are on some platforms).

alex:
I'm always meh on things that are DoS only :-)

Shnatsel:
Also meh from me. This is mostly DoS, although stack overflow is a soundness issue on some (relatively obscure) platforms.

It's probably valuable to record those as well, but on the other hand we don't really want to spam people with advisories, and we already kind of do - for example, there is a stack overflow in yaml parser used in clap crate without a semver-compatible upgrade path, but it's not actually exposed to untrusted input, yet everything that depends on clap gets a non-actionable advisory... which is a whole lot of crates.

I think this is precedent in favor of collecting severity information as recently mentioned [here](https://github.com/RustSec/advisory-db/issues/20#issuecomment-486298160).

Since `ammonia` is an HTML sanitizer which is by definition exposed to untrusted input, and there is reasonable expectation for it to be able to handle that without bringing down the entire process, I'm fine with merging this even without severity information in place.

dpc:
[Published in `cargo-crev`](https://github.com/dpc/crev-proofs/commit/82a9c1de91d844a4f288bc7f3129ff280197cd87)

xfix:
About yaml-rust vulnerability, yes, the issue isn't really exploitable with how clap is using it. In fact, pretty much no issue in yaml-rust would affect clap, due to its use of trusted YAML files. However, let's say there would be a serious vulnerability in yaml-rust, perhaps even leading to code execution - in my opinion, it should be posted, even if it doesn't affect clap, even if clap users will get audit errors that don't affect it.

xfix:
About the severity information, assuming a reasonable worst case (this is how it's typically done for libraries, CVSS isn't really designed with libraries in mind), this vulnerability ends up with CVSS Base Score of 7.5, which is high (network, low attack complexity, no privileges required, no user interaction, unchanged scope, no confidentiality impact, no integrity impact, and high availability impact).

Network is because usually ammonia would end up being used for web applications, due to being a HTML sanitization library. It can be used by publicly available websites, such as blogs.

Low attack complexity because the attack doesn't have a probability of working - it will work 100% of time, every time.

No privileges required because it's reasonable to allow a subset of HTML in blog comments, for instance. Even if commenting requires registration, this isn't a privilege as far CVSS is concerned, as registering an account on many sites is easy.

No user interaction, because it's not necessary for anyone else to do anything.

High availability impact because performing an attack crashes the server. Even assuming system restarting the service when it crashes, it's easy to perform the attack again. Exploiting the attack will drop all HTTP server connections, making the service pretty much unusable. Additionally, the attack in many cases is persistent (for instance for blog comments use-case), and visiting the page with an attack will crash the server again.

tarcieri:
Thanks for the feedback everyone. Based on what everyone said, unless anyone strongly objects I think we should go ahead and merge it.

Shnatsel:
Based on github reactions everyone involved is okay with that. Let's go ahead and merge this!

tarcieri:
Assigned `RUSTSEC-2019-0001` in #94

PR #97, opened 2019-05-15
Add protobuf out-of-memory vulnerability

gedigi:


tarcieri:
Assigned RUSTSEC-2019-0003 in https://github.com/RustSec/advisory-db/pull/99

PR #104, opened 2019-06-06
[portaudio] add build script RCE

mcginty:
Via: https://github.com/RustAudio/rust-portaudio/issues/144

This is, as of yet, unfixed.

### DIFF
--- a/crates/portaudio/RUSTSEC-0000-0000.toml
+++ b/crates/portaudio/RUSTSEC-0000-0000.toml
@@ -1,0 +1,15 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "portaudio"
+date = "2016-08-01"
+title = "HTTP download and execution allows MitM RCE"
+description = """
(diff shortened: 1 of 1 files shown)


Shnatsel:
Nice find. We'll need a linter for that.

crates/portaudio/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,15 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "portaudio"
+date = "2016-08-01"
```
tarcieri: 2016?

tarcieri: Oh wow, so it is...

tarcieri:
@mcginty thanks for retroactively filing this! Wonder how many others like this there are

tarcieri:
Assigned `RUSTSEC-2016-0003` in #105

Issue #106, opened 2019-06-09
Format string vulnerabilities (and more) in multiple safe curses wrapper crates.


thomcc:
I've noticed potentially exploitable security vulnerabilities in the following crates[0], which can be triggered by users of these libraries without writing unsafe code.

I'm filing an issue rather than a PR since I don't know that I have time to see this through to it's conclusion, especially since multiple crates and several functions are involved. The FAQ [indicates that this is okay](https://github.com/RustSec/advisory-db/blob/master/CONTRIBUTING.md#faq).

### `pancurses`

The crate https://crates.io/crates/pancurses exposes the functions `Window::printw` and `Window::mvprintw` as safe functions (I think this is all of them, but could be wrong), despite these passing an arbitrary rust `str` to into C code which expects to receive a printf-style format string. This can lead to a [format string vulnerability](https://www.owasp.org/index.php/Format_string_attack).

I filed https://github.com/ihalila/pancurses/issues/66 about this, shortly before filing this issue.

### `ncurses`

The crate https://crates.io/crates/ncurses exposes the functions `printw`, `mvprintw`, and `mvwprintw` as safe functions, which have the same issues described above.

In https://github.com/jeaye/ncurses-rs/issues/172 I raised this issue about `printw` (but missed that it wasn't just `printw`), which lead to that functions deprecation[1].  When writing this issue for the advisory-db, I noticed the existence of the other two functions. I don't see any other functions accepting format strings (the `scanw` functions, mercifully, are not exposed).

Additionally, while filing this bug, I noticed the following additional vulnerabilities or memory safety issues (I haven't filed issues in the `ncurses` crate's repository about these):

- [`instr`](https://github.com/jeaye/ncurses-rs/blob/2a5dc99ef40c8f836b60b8a5ae9814410fc3bbcb/src/lib.rs#L596) is exposed, which has a buffer overflow (the curses `instr` function is somewhat similar to the `gets` function from the C stdlib, in that writes as much text that it wants into a lengthless buffer pointer).
    - EDIT: I've noticed that `mvwinstr` has the same problem, just not as directly.

[...]

PR #107, opened 2019-06-15
Add advisory for ncurses

thomcc:
See #106.

### DIFF
--- a/crates/ncurses/RUSTSEC-0000-0000.toml
+++ b/crates/ncurses/RUSTSEC-0000-0000.toml
@@ -1,0 +1,22 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+
+package = "ncurses"
+date = "2019-06-15"
+
+title = "Buffer overflow and format vulnerabilities in functions exposed without unsafe"
+
+description = """
+`ncurses` exposes functions from the ncurses library which:
+
+- Pass buffers without length to C functions that may write an arbitrary amount of
+  data, leading to a buffer overflow. (`instr`, `mvwinstr`, etc)
+- Passes rust &str to strings expecting C format arguments, allowing a format
+  vulnerability (functions in the `printw` family).
+"""
+
+patched_versions = []
+
+url = "https://github.com/RustSec/advisory-db/issues/106"
+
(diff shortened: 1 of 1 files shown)


crates/ncurses/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,24 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+
+package = "ncurses"
+date = "2019-06-15"
+
+# Single-line description of a vulnerability (mandatory)
```
alex: Please remove these comments.

thomcc: Sorry, fixed.

crates/ncurses/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,22 @@
...
+
+title = "Buffer overflow and format vulnerabilities in functions exposed without unsafe"
+
+description = """
+`ncurses` exposes functions from the ncurses library which:
+
+- Pass buffers without length to C functions that may write an arbitrary amount of
+  data, leading to a buffer overflow. (`instr`, `mvwinstr`, etc)
+- Passes rust &str to strings expecting C format arguments, allowing a format
+  vulnerability (functions in the `printw` family).
```
8573: I'm not a member of @RustSec and I don't know at how knowledgeable of readers these descriptions are aimed, but I can imagine that a layperson might gloss over "format vulnerability", not knowing why they should care, for which reason I like that #108 explains that it "trivially allows writing arbitrary data to stack memory."

tarcieri: I agree this could be improved, however I will go ahead and merge the advisory as-is. Feel free to submit a follow up PR to improve the wording.

tarcieri:
Assigned `RUSTSEC-2019-0004` in #109

tarcieri:
Seems I screwed this up and `RUSTSEC-2019-0004` is already in use by `libp2p-core`. I have opened a second PR to re-assign it as `RUSTSEC-2019-0006`, and opened some issues about both detecting this in the future in CI as well as automating assigning of `RUSTSEC-*` identifiers to unassigned advisories:

https://github.com/RustSec/advisory-db/pull/110

PR #108, opened 2019-06-15
Add advisory for pancurses

thomcc:
See #106


tarcieri:
Assigned `RUSTSEC-2019-0005` in #109

PR #114, opened 2019-06-23
RUSTSEC-2019-0006: Use -0005's format vuln wording

8573:
As filed, advisory RUSTSEC-2019-0006 simply notes that certain
functions in the covered crate create a "format vulnerability". This
patch, following up on [an exchange of comments on GitHub][1], edits
advisory RUSTSEC-2019-0006 to summarize the risk introduced by a
format vulnerability, copying the wording of the associated advisory
RUSTSEC-2019-0005.

[1]: <https://github.com/RustSec/advisory-db/pull/107#pullrequestreview-250212575>


tarcieri:
Thanks for the clarification! (saw your original note on this, but thanks for following up)

PR #119, opened 2019-06-30
Add advisory for SmallVec issue #148

Shnatsel:
https://github.com/servo/rust-smallvec/issues/148


crates/smallvec/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,58 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+
+# Name of the affected crate (mandatory)
```
tarcieri: Please remove the comments

lucab:
Was a CVE already requested/assigned for this? If not, is anybody planning to request one? I'd be interesting in that and can offer some help if needed.

Shnatsel:
Help with a CVE would be great! Last time I tried to get a CVE through iwantacve.org I've waited for months and never heard back. It's been https://github.com/rust-lang/rust/issues/53566 and I still want one, by the way.

Do you think it's also a good idea to include https://github.com/servo/rust-smallvec/issues/149 in the same CVE? It has a very similar exploitation pattern. I've failed to determine the exact point where the bug was introduced, though.

lucab:
@Shnatsel I'd suggest to have an advisory for servo/rust-smallvec#149 too (the fixed-version is enough, as a start) and wait for @tarcieri to assign stable IDs for both.
I can then try to forward this to my company secteam for CVE bureaucracy, if nobody gets to that in a quicker way.

attritionorg:
This was just assigned CVE-2019-1010299.

Shnatsel:
@lucab the other smallvec issue is in as RUSTSEC-2019-0012, so please proceed with the CVE magic.

lucab:
I think @attritionorg has a quicker channel than me, so I'll wait for their feedback first in order to avoid getting duplicated IDs.

attritionorg:
@lucab I do not. MITRE has an unofficial policy to ignore any mail from me, on any topic, including CVE questions/disputes/assignments.

Shnatsel:
The assignment of CVE-2019-1010299 was my application via iwantacve.org from February finally coming through.

PR #129, opened 2019-07-20
actually memoffset also had an uninit-drop vuln

RalfJung:
This affects all versions ever published.

### DIFF
--- a/crates/memoffset/RUSTSEC-2019-0011.toml
+++ b/crates/memoffset/RUSTSEC-2019-0011.toml
@@ -5,7 +5,7 @@ date = "2019-07-16"
 title = "Flaw in offset_of and span_of causes SIGILL, drops uninitialized memory of arbitrary type on panic in client code"
 description = """
 Affected versions of this crate caused traps and/or memory unsafety by zero-initializing references.
-They also could leat to uninitialized memory being dropped if the field for which the offset is requested was behind a deref coercion, and that deref coercion caused a panic.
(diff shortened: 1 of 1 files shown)


crates/memoffset/RUSTSEC-2019-0011.toml
```diff
@@ -2,13 +2,13 @@
 id = "RUSTSEC-2019-0011"
 package = "memoffset"
 date = "2019-07-16"
-title = "Flaw in offset_of and span_of causes SIGILL, potential memory unsafety"
+title = "Flaw in offset_of and span_of causes SIGILL, drops uninitialized memory of arbitrary type on panic in client code"
 description = """
 Affected versions of this crate caused traps and/or memory unsafety by zero-initializing references.
+They also could leat to uninitialized memory being dropped if the field for which the offset is requested was behind a deref coercion, and that deref coercion caused a panic.
```
Shnatsel: s/leat/lead/

RalfJung: Thanks, fixed.

tarcieri:
Should this note the potential implications are memory disclosure and/or RCE?

RalfJung:
I don't have any experience with vulnerability reports -- no idea.^^

There are no known cases of this actually happening in practice.

tarcieri:
@RalfJung ok, will merge this as-is and we can potentially follow up on severity information

crates/memoffset/RUSTSEC-2019-0011.toml
kud1ing: "leat" => "lead"?

PR #131, opened 2019-08-26
Add CVE-2019-15541 in rustls example code

ctz:
This is a bit of a weird one because it's in example code. No amount of upgrading or downgrading will help for vulnerable downstream crates that copied and built upon the problematic example code. For that reason I've listed pre-patch versions as `unaffected_versions` -- but I'd appreciate your opinion on this.

### DIFF
--- a/crates/rustls/RUSTSEC-0000-0000.toml
+++ b/crates/rustls/RUSTSEC-0000-0000.toml
@@ -1,0 +1,23 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rustls"
+date = "2019-08-26"
+title = "Denial of service in example code"
+
+description = """
+Example code provided alongside rustls contained a denial of service condition
+when a socket was uncleanly closed with pending TLS data to send.
+
+This does not affect the core rustls library, but downstream users may have used
+this code as a basis so are advised to ensure they do not have similar bugs
+as a result.
+
+Because crate version numbers cover the library but not the examples, this advisory
+covers no versions.
+"""
+
+patched_versions = [">= 0.16.0"]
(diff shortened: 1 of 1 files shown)


tarcieri:
@ctz I've been hacking on the `rustsec` crate a bunch and I'm wondering if it might be worth incorporating a concept of an informational advisory that doesn't impact any particular crate versions and won't show up in `cargo-audit`, but allows you to leverage RustSec as a way to catalog and distribute it. What do you think about that?

ctz:
That sounds ideal to me.

tarcieri:
@ctz cool, I happen to be actively working on the `rustsec` crate right now so I will try to build that in as a first-class concept. Perhaps we could even figure out a way to make it warn but not error. It seems like an "informational advisory" might also be a good way to warn about things like unmaintained crates:

https://internals.rust-lang.org/t/tracking-unmaintained-crates-using-rustsec/10884

In the meantime we can do a hack like:

```
patched_versions = [">= 0"]
```

...to ensure it doesn't actually break anyone's build.

Shnatsel:
There are even more reasons not to surface this in cargo-audit, or for that matter anything that parses RustSec DB:

1. This info should be surfaced to maintainers of directly dependent crates only. If they're found vulnerable, they should get their own RustSec advisories for specific versions. This advisory if landed as-is would also create a non-actionable message to transitively dependent crates, which is a glaring false positive.
2. There is no way to make this message go away. You're stuck with it essentially forever.

Instead I would open bugs on bug trackers of prominent uses of rustls, and file a CVE to alert companies with their proprietary code.

ctz:
Going to close this, since it was likely an overreaction.

PR #135, opened 2019-08-31
Add hdr decoder use-after-free advisory

HeroicKatora:
As asked for here: https://github.com/image-rs/image/pull/985#issuecomment-526855953

### DIFF
--- a/crates/image/RUSTSEC-0000-0000.toml
+++ b/crates/image/RUSTSEC-0000-0000.toml
@@ -1,18 +1,8 @@
 [advisory]
-# Identifier for the advisory (mandatory). Will be assigned a "RUSTSEC-YYYY-NNNN"
-# identifier e.g. RUSTSEC-2018-0001. Please use "RUSTSEC-0000-0000" in PRs.
 id = "RUSTSEC-0000-0000"
-
-# Name of the affected crate (mandatory)
 package = "image"
-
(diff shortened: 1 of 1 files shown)


crates/image/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,70 @@
...
+unaffected_versions = ["< 0.10.2"]
+
+# URL to a long-form description of this issue, e.g. a GitHub issue/PR,
+# a change log entry, or a blogpost announcing the release (optional)
+url = "https://github.com/image-rs/image/pull/985"
+
+# Keywords which describe this vulnerability, similar to Cargo (optional)
+keywords = ["drop", "use-after-free"]
+
+# Vulnerability aliases, e.g. CVE IDs (optional but recommended)
```
alex: You can completely remove the unused options.

crates/image/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,70 @@
+[advisory]
+# Identifier for the advisory (mandatory). Will be assigned a "RUSTSEC-YYYY-NNNN"
```
alex: Can you please remove these comments?

HeroicKatora: Sure.

PR #138, opened 2019-09-01
Add advisory for string-interner

lo48576:
As asked for at <https://github.com/Robbepop/string-interner/issues/9#issuecomment-526857810>.

Some notes:

* I think the attacker can read some already dropped memory, not by direct access, but by brute-forcing.
  I don't know how I can say it in the title.
* The fix is [already merged](https://github.com/Robbepop/string-interner/pull/10), but not yet released.
* I sent [another patch](https://github.com/Robbepop/string-interner/pull/12) to add an additional test that checks there are no dangling pointers in such situations, but it is not yet merged.
  So I didn't write about it in the description.
* I'm not native English speaker, so feel free to point out if there are unnatural wordings.

### DIFF
--- a/crates/string-interner/RUSTSEC-0000-0000.toml
+++ b/crates/string-interner/RUSTSEC-0000-0000.toml
@@ -1,0 +1,20 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "string-interner"
+date = "2019-08-24"
+title = "Cloned interners may read already dropped strings"
+description = """
+Affected versions of this crate did not clone contained strings when an interner is cloned.
+Interners have raw pointers to the contained strings, and they keep pointing the strings which the old interner owns, after the interner is cloned.
+If a new cloned interner is alive and the old original interner is dead, the new interner has dangling pointers to the old interner's storage, which is already dropped.
+
+This allows an attacker to read the already freed memory.
+The dangling pointers are used by the interners to check a string is already interned.
+An attacker can do brute force attack to get the data pointed by the dangling pointer.
+
+The flaw was corrected by <https://github.com/Robbepop/string-interner/pull/10>.
+This patch implements `Clone` manually to the interner type, so that the internal raw pointers always point the strings owned by the same interner.
+"""
+patched_versions = [">= 0.7.1"]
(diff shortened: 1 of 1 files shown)


Shnatsel:
It seems that bruteforcing is not required of strings in both old and new interners are attacker-controlled, although I might be wrong.

> This allows an attacker to read the uninitialized memory.

I believe it would be more precise to say "This allows an attacker to read already freed memory". 

Other than that LGTM. The title is very nice - precise and to the point!

lo48576:
## About "uninitialized memory"
This is my mistake.
Fixed and pushed amended commit.

## About bruteforcing
Raw (dangling) pointers are not exposed, and **the data there is not directly returned** to the attacker.
The pointers are only used by the internal `HashMap::<InternalStrRef, _>::get()` and `HashMap::<InternalStrRef, _>::insert()`, through `<InternalStrRef as Hash>::hash()` and `<InternalStrRef as Eq>::eq()`.
Freed memory might be **filled with random data** ([at least when I tested](https://github.com/lo48576/string-interner/blob/5bb5485c8479b2bfb40d83758c0db56f8f466efe/TEST_RESULT.md)).

To know the (potentially random) data at the freed memory, an attacker use `StringInterner::get()` (or similar methods), and it returns `Some(_)` if the data at the freed memory is equal to the given string slice.
Therefore, an attecker should do brute force attack to the interner, until `get()` returns `Some(_)`.

This is the reason I think bruteforcing is necessary, but I might have overlooked more efficient way...

crates/string-interner/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,37 @@
+[advisory]
+# Identifier for the advisory (mandatory). Will be assigned a "RUSTSEC-YYYY-NNNN"
```
tarcieri: Please remove the comments

lo48576: Oops... Removed.

lo48576:
Removed comments and empty lines.

tarcieri:
Assigned `RUSTSEC-2019-0023` in #163

PR #146, opened 2019-09-02
File standard libary/core vulns (closes RustSec/cargo-audit#46)

tarcieri:
Files vulnerabilities in the standard library originally reported at:

https://groups.google.com/forum/#!forum/rustlang-security-announcements

Or otherwise collected at:

https://github.com/RustSec/cargo-audit/issues/46

The `rustsec` crate doesn't presently consume these, but I'd like to add
support ASAP.

### DIFF
--- a/rust/std/CVE-2018-1000810.toml
+++ b/rust/std/CVE-2018-1000810.toml
@@ -1,0 +1,99 @@
+[advisory]
+id = "CVE-2018-1000810"
+package = "std"
+date = "2018-09-21"
+title = "Buffer overflow vulnenrability in str::repeat()"
+description = """
+The Rust team was recently notified of a security vulnerability affecting
+the `str::repeat` function in the standard library. If your code does not
+use this function, it is not affected.
+
+## Overview
+
+This vulnerability is an instance of CWE-680: Integer Overflow to Buffer
+Overflow[1].
+
+The `str::repeat` function in the standard library allows repeating a
+string a fixed number of times, returning an owned version of the final
+string. The capacity of the final string is calculated by multiplying
+the length of the string being repeated by the number of copies. This
+calculation can overflow, and this case was not properly checked for.
+
+The rest of the implementation of `str::repeat` contains unsafe code
+that relies on a preallocated vector having the capacity calculated
+earlier. On integer overflow the capacity will be less than required,
+and which then writes outside of the allocated buffer, leading to
+buffer overflow.
+
+## Affected Versions
+
+While the `str::repeat` function has been in Rust since 1.16.0, this
+vulnerability was introduced into the standard library in pull
+request #48657 [2]. The pull request was merged on March 6, 2018 and
+was first part of the 1.26.0 stable released on May 10, 2018.
+
+As such, this vulnerability affects:
+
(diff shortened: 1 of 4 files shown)


rust/std/CVE-2018-1000810.toml
```diff
@@ -0,0 +1,103 @@
...
+[1]: https://cwe.mitre.org/data/definitions/680.html
+[2]: https://github.com/rust-lang/rust/pull/48657
+[3]: https://github.com/rust-lang/rust/pull/54397
+[4]: https://github.com/rust-lang/rust/pull/54398
+[5]: https://github.com/rust-lang/rust/pull/54399
+"""
+patched_versions = [">= 1.29.1"]
+unaffected_versions = ["< 1.26.0"]
+url = "https://groups.google.com/forum/#!topic/rustlang-security-announcements/CmSuTm-SaU0"
+categories = ["dos"]
```
alex: This is not just a DoS, it's a heap buffer overflow write

tarcieri: I originally added `rce` before noticing it's local-only(?), and the current coarse-grained categorization doesn't have a proper category to capture that:

https://github.com/RustSec/rustsec-crate/blob/master/src/advisory/category.rs#L48

But it hasn't shipped yet, so perhaps it should be changed to `code-execution` instead of `rce`

Shnatsel:
I do not see much value in doing this until we have a story around consuming these. Right now all it would do is require us to manually duplicate Rust CVEs in here and show them on rustsec.org. That sounds like we're signing up for manual work with little or no gain.

tarcieri:
@Shnatsel next step to me is to consume them in the database, but it's a chicken and egg problem writing the code to consume them if they're not in the database to begin with.

My first thought when I actually found the extended descriptions of each vuln was "Wow, that was actually kind of hard to find..."

tarcieri:
Another couple thoughts on this:

- All of these advisories are written in Markdown, but I don't think there exists a Markdown rendering of them anywhere?
- The `VecDeque` vulnerability (CVE-2018-1000657) never received such a writeup for whatever reason, despite being one of the most severe so far

tarcieri:
Merging this with the initial goals of being able to load these in the `rustsec` crate's database and provide more discoverable Markdown renderings of them on rustsec.org.

I think these are low volume and low effort to import to the point that they won't cause undue burden.

tarcieri:
These are now rendered on rustsec.org:

- [CVE-2018-1000810: std: Buffer overflow vulnenrability in str::repeat()](https://rustsec.org/advisories/CVE-2018-1000810.html)
- [CVE-2018-1000657: std: Buffer overflow vulnenrability in VecDeque::reserve()](https://rustsec.org/advisories/CVE-2018-1000657.html)
- [CVE-2018-1000622: rustdoc: Uncontrolled search path element vulnerability in rustdoc plugins](https://rustsec.org/advisories/CVE-2018-1000622.html)
- [CVE-2019-12083: std: Memory safety vulnerabilities arising from `Error::type_id`](https://rustsec.org/advisories/CVE-2019-12083.html)

PR #147, opened 2019-09-02
Undefined behavior in renderdoc crate

ebkalderon:
While updating the [renderdoc-rs](https://github.com/ebkalderon/renderdoc-rs) bindings a few months ago, I had identified several methods that were taking immutable reference to `self` but were actually making internally mutating [RenderDoc](https://renderdoc.org) API calls, which could potentially lead to undefined behavior.

This issue was corrected in the 0.5.0 release, and it was [recently suggested](https://github.com/ebkalderon/renderdoc-rs/issues/28#issuecomment-526860298) that I file a public advisory here, which was a very good idea.

I would appreciate any feedback in reporting this advisory properly, and I hope to report any such cases in a more timely manner in the future.

### DIFF
--- a/crates/renderdoc/RUSTSEC-0000-0000.toml
+++ b/crates/renderdoc/RUSTSEC-0000-0000.toml
@@ -1,0 +1,23 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "renderdoc"
+date = "2019-09-02"
+title = "Internally mutating methods take immutable ref self"
+description = """
+Affected versions of this crate exposed several methods which took `self` by
+immutable reference, despite the requesting the RenderDoc API to set a mutable
+value internally.
+
+This is technically unsound and calling these methods from multiple threads
+without synchronization could lead to unexpected and unpredictable behavior.
+ 
+The flaw was corrected in release 0.5.0.
+"""
+patched_versions = [">= 0.5.0"]
+url = "https://github.com/ebkalderon/renderdoc-rs/pull/32"
+keywords = ["undefined_behavior"]
+affected_os = ["linux", "windows"]
+affected_functions = [
+    "renderdoc::api::RenderDocV110::trigger_multi_frame_capture",
+    "renderdoc::api::RenderDocV120::set_capture_file_comments",
+]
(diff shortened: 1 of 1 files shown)


ebkalderon:
@Shnatsel I am also working on publishing a security advisory on GitHub as well. If you would like to join as a collaborator to review the advisory text, I would be happy to add you.

crates/renderdoc/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,23 @@
...
+
+This is technically unsound and calling these methods from multiple threads
+without synchronization could lead to unexpected and unpredictable behavior.
+ 
+The flaw was corrected in release 0.5.0.
+"""
+patched_versions = [">= 0.5.0"]
+url = "https://github.com/ebkalderon/renderdoc-rs/pull/32"
+keywords = ["undefined_behavior"]
+affected_os = ["linux", "windows"]
```
tarcieri: Is there some reason this advisory is actually exclusive to these OSes? If not, you should remove this.

ebkalderon: I listed those particular OSs because RenderDoc, and this crate by extension, does not compile on macOS or other platforms. Still, I wouldn't mind removing it nonetheless if it's unnecessary.

tarcieri: I'd suggest removing it in that case

ebkalderon: Will do, thanks.

ebkalderon: Okay, done! :+1:

ebkalderon:
Thanks for the review, @tarcieri!

tarcieri:
Assigned `RUSTSEC-2019-0018` in https://github.com/RustSec/advisory-db/pull/148

ebkalderon:
The corresponding GitHub security advisory has been published here: https://github.com/ebkalderon/renderdoc-rs/security/advisories/GHSA-4mmc-49vf-jmcp

PR #124, opened 2019-07-16
add memoffset issue

RalfJung:

PR #149, opened 2019-09-04
Undefined behavior in Rand

vks:


Shnatsel:
Is there any practical impact out of this? I'm not aware of unaligned accesses being exploitable in practice.

Does "provenance rules" mean "2 mutable pointers to the same thing"?

dhardy:
Considering that the code in question should have been thoroughly tested on all our CI platforms to produce correct results, I think it quite unlikely there was any exploitable bug.

vks:
> Does "provenance rules" mean "2 mutable pointers to the same thing"?

No, they mean in this case that `&mut slice[0] as *mut _` is a raw pointer that can only be used to access the first element. However, Rand used this pointer to access higher elements as well. `slice.as_mut_ptr()` has to be used instead.

RalfJung:
Does this refer to https://github.com/rust-random/rand/pull/780 or to https://github.com/rust-random/rand/pull/784?

The former (https://github.com/rust-random/rand/pull/780) is, to my knowledge, not de-facto UB with any existing Rust versions. Whether it is de jure UB is hard to say as Rust does not have an aliasing model at this point, but the code might become de jure and also de facto UB in the future. The provenance violation comes up in an experimental pointer provenance model for Rust (Stacked Borrows). While we do exploit provenance to some extend with `noalias`, I do not think rand violated `noalias` (but I will not make any definite statement about `noalias`; the LLVM spec is extremely fuzzy in that regard and likely inconsistent).

The latter (https://github.com/rust-random/rand/pull/784) *is* de-facto UB as we tell LLVM that those accesses are aligned. It has nothing to do with pointer provenance though. Whether this UB actually leads to bad code generation or is exploitable, I do not know. (This comes back to the question I raised a while ago: is UB [de facto or de jure] sufficient for an advisory?)

vks:
> Btw, it might be a good idea to also link to the issue and/or pull request so that one can get a better idea of what the flaw and fix look like. Not sure if the url field can be used multiple times?

I added a link to the changelog, with contains the issue numbers. I'm also not sure whether I can specify multiple URLs instead.

tarcieri:
We've had a few tough calls on whether something deserves an advisory in the past. In #124, the more we studied it, the more it looked exploitable. As I look at this one, I feel the opposite: with more analysis I'm less convinced there's anything exploitable here. My feeling on this is it isn't worth an advisory unless someone can present a credible threat (even if hypothetical and far fetched).

tarcieri:
Closing this as unexploitable. Please reopen or refile if there is a credible threat.

RalfJung:
Would it make sense to reopen this an an "informational (unsound)" advisory?

tarcieri:
@RalfJung reopened for discussion at least

vks:
I guess it would encourage people to upgrade Rand. Does RustSec have a policy for "informational" advisories?

crates/rand/RUSTSEC-0000-0000.toml
```diff
@@ -2,6 +2,7 @@
 id = "RUSTSEC-0000-0000"
 package = "rand"
 date = "2019-04-19"
+informational = "unsound"
 title = "Invalid pointer usage"
 description = """
 Affected versions of this crate violated pointer provenance rules, resulting in
```
RalfJung, line 8: Given that those are experimental, I am not sure if it is appropriate to file an advisory for them.

RalfJung: IOW, the authority on UB is https://doc.rust-lang.org/reference/behavior-considered-undefined.html. And whether rand violated "the pointer aliasing rules" is unclear as those are not properly defined yet. Models that allow the old rand code are conceivable.

In contrast, unaligned accesses are explicitly UB in that list.

vks: Are you saying that we should only file the advisory for `rand_core`?

RalfJung: Yeah, basically.

vks:
I rebased on master and removed the mention of pointer provenance rules.

crates/rand_core/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,15 @@
...
+id = "RUSTSEC-0000-0000"
+package = "rand_core"
+date = "2019-04-19"
+informational = "unsound"
+title = "Unaligned memory access"
+description = """
+Affected versions of this crate violated alignment when casting byte slices to
+integer slices, resulting in undefined behavior.
+
+The flaws were corrected by Ralf Jung and Diggory Hardy.
```
RalfJung, line 11:
```suggestion
The flaw was corrected by Ralf Jung and Diggory Hardy.
```

crates/rand_core/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,15 @@
...
+date = "2019-04-19"
+informational = "unsound"
+title = "Unaligned memory access"
+description = """
+Affected versions of this crate violated alignment when casting byte slices to
+integer slices, resulting in undefined behavior.
+
+The flaw was corrected by Ralf Jung and Diggory Hardy.
+"""
+patched_versions = [">= 0.4.2"]
```
RalfJung, line 13: This doesn't look like the right toml format (maybe it changed); see https://github.com/RustSec/advisory-db#advisory-format.

RalfJung: @vks that is probably also the cause of the CI failure, though the message is quite cryptic:
```
error: error loading advisory DB repo from .: RustSec error: parse error: missing field `versions` at line 1 column 1
```

PR #151, opened 2019-09-06
Add advisory for broken `blake2` impls

tarcieri:
BLAKE2b and BLAKE2s were implemented using the wrong block size. All versions of the `blake2` crate prior to v0.8.1 compute incorrect digests.

See: https://github.com/RustCrypto/MACs/issues/19


newpavlov:
>All versions of the blake2 crate prior to v0.8.1 compute incorrect digests

Only HMAC values. Digest values are correct.

tarcieri:
@newpavlov will push a fix, 1s

tarcieri:
Updated to note this only impacts HMAC-BLAKE2

PR #157, opened 2019-09-09
Upgrade to `rustsec` crate v0.13.0-alpha1; add linter

tarcieri:
Upgrades the `rustsec` crate to the latest alpha release and uses the new `rustsec::advisory::Linter` functionality to lint advisories currently in the database.

Several of them are using invalid keys and need to be updated.

Additionally, this corrects all advisories so they pass the linter.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,120 +1,204 @@
+//! RustSec Advisory DB Linter
+
 #![allow(clippy::never_loop)]
 
 use gumdrop::Options;
-use rustsec::{AdvisoryDatabase, Repository};
-use std::{env, process::exit};
+use rustsec::package;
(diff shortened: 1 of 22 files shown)


src/main.rs
```diff
@@ -1,120 +1,204 @@
...
+            writeln_error!(
+                &mut stdout,
+                "{} advisories contain errors!",
+                invalid_advisories
+            );
+            exit(1);
+        }
     }
 
-    // Check that each path in `affected_paths` starts with the crate name
```
tarcieri: Seems like to properly replace this we need to pass the crate name as fetched from https://crates.io into the linter so it can properly check the paths.

PR #160, opened 2019-09-15
Use-after-free in portaudio-rs and double free in linea

Phosphorus15:
Two unwind-safety problem was detected in crates mentioned above.


tarcieri:
Assigned `RUSTSEC-2019-0021` and `RUSTSEC-2019-0022` in #161 and #162

PR #164, opened 2019-09-18
RUSTSEC-2019-0022: Fix date

tarcieri:
Filed as 2017 instead of 2019


tarcieri:
We could definitely use a lint for this

PR #177, opened 2019-10-07
Fix escapes in hyper advisory

### LINKED ISSUE #159: hyper/RUSTSEC-2017-0002.toml has unquoted characters
The description contains the text:
> Serializing of headers to the socket did not filter the values for newline bytes (\r or \n),

The `\r` and `\n` are parsed as actual escaped cr/lf, and so the description is printed as:
```
 or alizing of headers to the socket did not filter the values for newline bytes (
),
which allowed for header values to split a request or response. People would not likely include
```

I'm not sure if this intended in the format of `description`, or if this is a bug in the `rustsec` crate. As far as I can tell, it isn't possible to print the description with `cargo audit`, so maybe this has been overlooked.

vks:
Fixes #159.

### DIFF
--- a/crates/hyper/RUSTSEC-2017-0002.toml
+++ b/crates/hyper/RUSTSEC-2017-0002.toml
@@ -6,7 +6,7 @@ date = "2017-01-23"
 url = "https://github.com/hyperium/hyper/wiki/Security-001"
 title = "headers containing newline characters can split messages"
 description = """
-Serializing of headers to the socket did not filter the values for newline bytes (\r or \n),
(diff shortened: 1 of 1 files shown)


tarcieri:
Let's see if that works...

tarcieri:
Well, it's better than it was, but `\r` and `\n` probably need to be escaped with backticks:

https://rustsec.org/advisories/RUSTSEC-2017-0002.html

vks:
I was not sure whether you are performing any markdown processing. Should I add the backticks?

tarcieri:
Yes please, we'll see if that fixes it.

Longer term the web site could probably benefit from a simpler rendering pipeline.

PR #178, opened 2019-10-07
Use backticks for escaped characters

vks:
Followup to #177.


tarcieri:
Looks good now, thanks! 👍 

https://rustsec.org/advisories/RUSTSEC-2017-0002.html

PR #181, opened 2019-10-07
Add unmaintained crate informational advisory: rust-crypto

tarcieri:
No releases since May 2016, no commits since September 2016, with 62 open issues and 37 open PRs.

Author is unresponsive:

https://github.com/DaGenix/rust-crypto/issues/440

Advisory includes a large list of maintained "successor" crates: `rust-crypto` was a kitchen sink of functionality, so the advisory contains a list of potential successor crates each with an algorithm-by-algorithm breakdown of what they support.

### DIFF
--- a/crates/rust-crypto/RUSTSEC-0000-0000.toml
+++ b/crates/rust-crypto/RUSTSEC-0000-0000.toml
@@ -1,0 +1,78 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rust-crypto"
+title = "rust-crypto is unmaintained; switch to a modern alternative"
+informational = "unmaintained"
+date = "2016-09-06" # last GitHub commit
+url = "https://github.com/DaGenix/rust-crypto/issues/440"
+unaffected_versions = ["> 0.2.36"] # last release
+description = """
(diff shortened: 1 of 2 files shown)


crates/rust-crypto/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,78 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rust-crypto"
+title = "rust-crypto is unmaintained; switch to a modern alternative"
+informational = "unmaintained"
+date = "2018-09-06" # last GitHub commit
+url = "https://github.com/DaGenix/rust-crypto/issues/440"
+unaffected_versions = ["> 0.2.36"] # last release
+description = """
+The `rust-crypto` crate has not seen a release or GitHub commit since 2016,
```
Shnatsel: Doesn't mesh with `date = "2018-09-06" # last GitHub commit` above

tarcieri: Good catch!

tarcieri: Fixed

PR #188, opened 2019-10-10
Flaw in CBOR deserializer allows stack overflow

pyfisch:


tarcieri:
As this is a CBOR crate whose raison d'etre is parsing potentially attacker-controlled data, I think the DoS here puts this particular vulnerability in-scope, but I'd like to at least briefly leave this open to hear if anyone disagrees. To me it seems pretty clear-cut.

PR #196, opened 2019-10-19
Update RUSTSEC-2019-0026.toml

kpp:
According to the change log it was introduced in 0.2.0

Versions 0.2.0-0.2.4 were yanked.


tarcieri:
Thanks!

PR #197, opened 2019-10-20
Add a flatbuffers unsound code advisory

nagisa:
A couple of observations:

* There are still no released versions with a fix.
* The issue does not strictly fall into any particular criteria listed in https://github.com/RustSec/advisory-db/blob/master/CONTRIBUTING.md#criteria, but given the nature of this library the input is likely to be some untrusted external input over network or similar. This gives an attacker an easy vector to exploit UB which could subsequently result in execution of unintended branches in code, cause panics, or circumvent authorization checks, etc.

### DIFF
--- a/crates/flatbuffers/RUSTSEC-2019-0024.toml
+++ b/crates/flatbuffers/RUSTSEC-2019-0024.toml
@@ -1,0 +1,17 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "flatbuffers"
+patched_versions = []
+unaffected_versions = ["< 0.4.0"]
+date = "2019-10-20"
+url = "https://github.com/google/flatbuffers/issues/5530"
+title = "Unsound `impl Follow for bool`"
+description = """
+The implementation of `impl Follow for bool` allows to reinterpret arbitrary bytes as a `bool`.
+
+In Rust `bool` has stringent requirements for its in-memory representation. Use of this function
+allows to violate these requirements and invoke undefined behaviour in safe code.
+"""
+
+[affected]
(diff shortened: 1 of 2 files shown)


crates/flatbuffers/RUSTSEC-2019-0024.toml
```diff
@@ -0,0 +1,17 @@
...
+title = "Unsound `impl Follow for bool`"
+description = """
+The implementation of `impl Follow for bool` allows to reinterpret arbitrary bytes as a `bool`.
+
+In Rust `bool` has stringent requirements for its in-memory representation. Use of this function
+allows to violate these requirements and invoke undefined behaviour in safe code.
+"""
+
+[affected]
+functions = { "<bool as flatbuffers::Follow>::follow" = [">= 0.4.0", "<= 0.6.0"] }
```
tarcieri: The `as` syntax is unfortunately not supported by our path parser. You'd either have to remove this section entirely, or just use `flatbuffers::Follow::follow` for now

tarcieri:
The filename of the advisory TOML file needs to be `RUSTSEC-0000-0000.toml`

crates/flatbuffers/RUSTSEC-2019-0024.toml
tarcieri: Went ahead and assigned this RUSTSEC-2019-0028 inline

nagisa:
Yeah, sorry, my bad for not getting to this sooner.

tarcieri:
@nagisa it's all good, it now has an assigned advisory ID and is ready to merge when the tests pass

tarcieri:
Assigned `RUSTSEC-2019-0028` in d520ed4

PR #200, opened 2019-11-04
rust-crypto: Add orion as alternative

brycx:
I'd like to propose that [orion](https://crates.io/crates/orion)(which I'm the author/maintainer of) be added to the list of alternative, maintained crates in the rust-crypto informational advisory.

### DIFF
--- a/crates/rust-crypto/RUSTSEC-2016-0005.toml
+++ b/crates/rust-crypto/RUSTSEC-2016-0005.toml
@@ -46,12 +46,12 @@ which algorithms you need:
   - Signature algorithms: Ed25519
   - Short-input PRF: SipHash24
 - [`orion`]:
-  - AEAD algorithms: ChaCha20Poly1305 (IETF version), XChaCha20Poly1305 (IETF version)
+  - AEAD algorithms: ChaCha20Poly1305 (IETF version), XChaCha20Poly1305
   - Digest algorithms: SHA-512, BLAKE2b
   - Key derivation: HKDF
   - MACs: HMAC, Poly1305
   - Password hashing: PBKDF2
-  - Stream ciphers: ChaCha20 (IETF version), XChaCha20 (IETF version)
+  - Stream ciphers: ChaCha20 (IETF version), XChaCha20
 
 [dalek-cryptography GitHub Org]: https://github.com/dalek-cryptography
 [RustCrypto GitHub Org]: https://github.com/RustCrypto
--- a/crates/rust-crypto/RUSTSEC-2016-0005.toml
+++ b/crates/rust-crypto/RUSTSEC-2016-0005.toml
@@ -45,6 +45,13 @@ which algorithms you need:
   - Public key encryption: NaCl "Box" (X25519 + XSalsa20Poly1305)
   - Signature algorithms: Ed25519
   - Short-input PRF: SipHash24
+- [`orion`]:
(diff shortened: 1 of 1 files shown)


Shnatsel:
The library README currently states:

> This library is **not suitable for production code** and **usage is at own risk.**

While the original rust-crypto crates did not provide particularly strong guarantees, it is nevertheless strange to suggest a library explicitly labeled as not suitable for production use as the migration path, especially since more mature solutions such as `ring` are available. Could you explain why `orion` is a good addition?

brycx:
> While the original rust-crypto crates did not provide particularly strong guarantees, it is nevertheless strange to suggest a library explicitly labeled as not suitable for production use as the migration path, especially since more mature solutions such as ring are available. Could you explain why orion is a good addition?

I totally understand this concern. The README states as such because the library hasn't been audited by a third party yet. IIRC most of the crates under `RustCrypto` have not been audited either and disclaimers for this have also been added to some of their newer crates. Though `orion` does not nearly see as much use in the ecosystem as crates under `RustCrypto` do, which is a clear disadvantage of course. But orion does apply extensive testing, or at least I like to think it does.

Currently some advantages of orion, as I see it, could be minimal use of unsafe code, a collected library if you don't like to pull in multiple dependencies, and (maybe? RustCrypto has seen some big updates recently) more supported pure-Rust implementations of common crypto algorithms. In cases when you need pure-Rust for e.g. WASM, orion could also be considered an alternative for that.

Of course, if you can and want to use wrappers around well-established libraries (libsodium/BoringSSL), that would probably be the best choice at the end of the day.

tarcieri:
Personally I'm ok with adding `orion`. Many of the `RustCrypto`-the-org crates have similar downsides (immature, pure Rust which makes assurances around constant time presently impossible due to LLVM). In as much as people have different requirements (e.g. WASM) I think it's nice for people to have options. All of that said I would still consider any of the previously mentioned options superior to `rust-crypto`-the-crate

crates/rust-crypto/RUSTSEC-2016-0005.toml
```diff
@@ -45,6 +45,13 @@ which algorithms you need:
   - Public key encryption: NaCl "Box" (X25519 + XSalsa20Poly1305)
   - Signature algorithms: Ed25519
   - Short-input PRF: SipHash24
+- [`orion`]:
+  - AEAD algorithms: ChaCha20Poly1305 (IETF version), XChaCha20Poly1305 (IETF version)
```
tarcieri: Quick point of clarification: there is no "IETF version" of XChaCha20/XChaCha20Poly1305.

All extant implementations of XChaCha20 and XChaCha20Poly1305 AEAD are built on the 96-bit nonce / 32-bit counter version of ChaCha20 (i.e. the "IETF" version), so there's no need to disambiguate with versions built on the legacy 64-bit nonce / 64-bit counter ChaCha20 ("djb" version). But it's also confusing to call it that because there are no published RFCs for such a construction (only `draft-arciszewski-xchacha`)

brycx: I was also unsure about adding "(IETF version)" to the eXtended nonce variant. I ended up trying to be as specific as possible.

> But it's also confusing to call it that because there are no published RFCs for such a construction (only draft-arciszewski-xchacha)

Good point, I hadn't taken that into account.

tarcieri:
@brycx thanks!

PR #204, opened 2019-11-21
Added RUSTSEC advisory for crust as an archived/unmaintained.

simlay:
Hi, I was looking into p2p crates and the crate with the highest number of downloads has been archived on github so, I thought it would be useful for anyone that uses it to know it's unmaintained.

Anyway, I think I followed the steps correctly. Let me know if there's anything I need to change.

Thanks!


tarcieri:
@simlay thank you!

PR #211, opened 2019-12-21
string-interner-0.6.4 also fixes RUSTSEC-2019-0023

basvandijk:
The fix https://github.com/Robbepop/string-interner/pull/10 released in 0.7.1 was also backported to the 0.6 release line in https://github.com/Robbepop/string-interner/pull/14 and released in 0.6.4.


tarcieri:
Thanks!

PR #217, opened 2020-01-09
Add advisory for hyperium/http/issues/352

Qwaz:
`HeaderMap::reserve()` used `usize::next_power_of_two()` to calculate the increased capacity. However, `next_power_of_two()` silently overflows to 0 if given a sufficently large number in release mode.

If the map was not empty when the overflow happens, the library will invoke `self.grow(0)` and start infinite probing. This allows an attacker who controls the argument to `reserve()` to cause a potential denial of service (DoS).

The flaw was corrected in 0.2.0 release of `http` crate.

### DIFF
--- a/crates/http/RUSTSEC-0000-0000.toml
+++ b/crates/http/RUSTSEC-0000-0000.toml
@@ -1,0 +1,79 @@
+# Before you submit a PR using this template, **please delete the comments**
+# explaining each field, as well as any unused fields.
+
+[advisory]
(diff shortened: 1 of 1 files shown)


kpcyrd:
Any plans to backport this to http 0.1?

Qwaz:
@kpcyrd The fix is available on v0.1.20, and the affected version range has been updated in [this PR](https://github.com/RustSec/advisory-db/pull/221).

PR #218, opened 2020-01-09
Add advisory for hyperium/http/issues/354,355

Qwaz:
Affected versions of this crate incorrectly used raw pointer, which introduced unsoundness in its public safe API.

[Failing to drop the Drain struct causes double-free](https://github.com/hyperium/http/issues/354), and [it is possible to violate Rust's alias rule and cause data race with Drain's Iterator implementation](https://github.com/hyperium/http/issues/355).

The flaw was corrected in 0.2.0 release of `http` crate.

### DIFF
--- a/crates/http/RUSTSEC-0000-0000.toml
+++ b/crates/http/RUSTSEC-0000-0000.toml
@@ -1,0 +1,78 @@
+# Before you submit a PR using this template, **please delete the comments**
+# explaining each field, as well as any unused fields.
+
+[advisory]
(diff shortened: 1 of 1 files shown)


roy-work:
For both of the linked issues, the commit that purports to fix the issue is tagged with 0.1.20 and 0.1.21; is it not correct that this flaw was corrected in the 0.1.20 release of http?

(Moving it to 0.2.0 means a semver breaking change for those of us upgrading.)

Qwaz:
I saw the new PR and it looked good. Thank you for correcting this.

PR #241, opened 2020-03-10
add redox-os-relibc-uaf

BurtonQin:
https://gitlab.redox-os.org/redox-os/relibc/issues/159
Affected versions of this crate would access memory through a dangling pointer after the pointed vector was dropped.
This is a use-after-free vulnerability and could allow an attacker to gain arbitrary code execution.
The flaw can be corrected by making the vector static to prevent dropping it.


tarcieri:
@BurtonQin did you mean to close this PR?

BurtonQin:
Yes, redox-os/relibc is not an app in crates.io

tarcieri:
Aah, no worries then!

PR #246, opened 2020-03-19
Add advisory for cbox

eduardosm:


Shnatsel:
@tarcieri how does specifying no affected versions work? Would this be rejected as invalid or assume all versions are vulnerable?

The tests have passed so I assume this works?

tarcieri:
Assigned `RUSTSEC-2020-0005` in #249

PR #247, opened 2020-03-19
fix: update patched version for 2019-0028

rockstar:
This patch updates the `RUSTSEC-2019-0028` advisory to show a patched version is available. The patch was added [in PR 5554](https://github.com/google/flatbuffers/pull/5554), and released with version `0.6.1`.


tarcieri:
Thanks!

Issue #248, opened 2020-03-23
Add CVSS metadata to advisories


djc:
It appears tools like Snyk keep track of [CVSS](https://www.first.org/cvss/) data. It would be nice if we could provide that for Rust code, too.

tarcieri:
It's already supported via the [`cvss` crate](https://github.com/RustSec/cvss.rs). Here's example usage in a vulnerability:

https://github.com/RustSec/advisory-db/blob/master/rust/std/CVE-2018-1000810.toml#L8

However, it looks like it never got added to the advisory template.

djc:
Right, and therefore not many advisories actually have it.

tarcieri:
It's also somewhat arcane to compute and not something I'd want to block collection of advisories on, but rather something I'd like for curators to add retroactively.

tarcieri:
I'll consider this mostly addressed by #250, but change the title to reflect that it'd still be nice if curators were to submit CVSS information retroactively.

pinkforest:
There isn't anything actionable here but I guess we are fishing for ideas how to retroactively collect this data.

Now that we have discussions -

I started a concrete discussion favor of this here:
https://github.com/rustsec/advisory-db/discussions/1350

Closing this as completed

PR #250, opened 2020-03-24
Mention CVSS field in template (see #248)

djc:


README.md
```diff
@@ -52,6 +52,10 @@ url = "https://github.com/mystuff/mycrate/issues/123"
 # "format-injection", "memory-corruption", "memory-exposure", "privilege-escalation"
 categories = ["crypto-failure"]
 
+# Optional: a Common Vulnerability Scoring System score. More information
+# can be found on the CVSS website, https://www.first.org/cvss/.
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H"
```
tarcieri: Since it's optional, I'd suggest commenting it out like the other optional fields

djc: (I took my queue from `categories`, which is also optional but not commented out.)

tarcieri: Aah, guess we aren't consistent about that. Oh well, thanks

tarcieri:
Thanks!

PR #251, opened 2020-03-24
bumpalo: Report memory exposure bug in realloc

fitzgen:


tarcieri:
Assigned `RUSTSEC-2020-0006` in https://github.com/RustSec/advisory-db/pull/252

PR #253, opened 2020-03-27
Report memory management error in `bitvec`

myrrlyn:
See myrrlyn/bitvec#55


Shnatsel:
Thanks for the report, and fancy meeting you here! Feels like just recently we've been debating lunar charts and their influence on morphology 😄 

Please clarify what the status of versions before 0.11 is, and then we can merge. It's also a good idea to yank affected versions from crates.io.

myrrlyn:
The affected functions did not exist before `0.11`. All affected versions are yanked.

tarcieri:
Assigned `RUSTSEC-2020-0007` in #254

PR #264, opened 2020-04-17
Mark lz4-compress as unmaintained

Shnatsel:


tarcieri:
Assigned `RUSTSEC-2017-0007` in #266

PR #267, opened 2020-04-23
Add advisory for rusqlite

thomcc:
For work I completed an audit of rusqlite's unsafe code. There were a number of issues for various issues, most of them in uncommonly used APIs, but they could be fairly severe nonetheless.

I'd... prefer not to split this out among N separate advisories. Mostly because I'm lazy, but I also don't really see the benefit it would give. (Aside from making the `description` better, I guess)

Note that while VTab/VTabCursor not being unsafe traits isn't listed in the `functions`, you need to use `create_module` in order to expose that so it should be fine.

### DIFF
--- a/crates/rusqlite/RUSTSEC-0000-0000.toml
+++ b/crates/rusqlite/RUSTSEC-0000-0000.toml
@@ -1,0 +1,25 @@
+
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rusqlite"
+date = "2020-04-23"
+title = "Various memory safety issues"
+url = "https://github.com/rusqlite/rusqlite/releases/tag/0.23.0"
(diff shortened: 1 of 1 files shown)


Shnatsel:
Seeing that there is no semver-compatible fixed version, it would be nice to also specify in which version every issue was introduced. However, seeing that there's a lot of them and you need to opt in to extra features to get to the problematic code, it might not be worth the trouble.

thomcc:
Okay, I’ll take a look.

tarcieri:
Will go ahead and merge this. Please open a separate PR if you'd like to add info about which versions were unaffected.

tarcieri:
Assigned `RUSTSEC-2020-0014` in #274

PR #268, opened 2020-04-23
Advisory for plutonium.

najamelan:
I think it's good that cargo-audit reports this. [Reported first at cargo-geiger](https://github.com/rust-secure-code/cargo-geiger/issues/101) where it was suggested this was more for cargo-audit.

### DIFF
--- a/crates/plutonium/RUSTSEC-0000-0000.toml
+++ b/crates/plutonium/RUSTSEC-0000-0000.toml
@@ -1,0 +1,15 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "plutonium"
+date = "2020-04-23"
+title = "Crate intended to hide unsafe use."
+url = "https://www.reddit.com/r/rust/comments/g5rsuh/show_me_the_most_illegal_rust_code_youve_ever/fo88z2d?utm_source=share&utm_medium=web2x"
+description = """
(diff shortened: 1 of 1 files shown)


najamelan:
The check fails on bad version, but I don't know how to put it better.

crates/plutonium/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,15 @@
...
+url = "https://www.reddit.com/r/rust/comments/g5rsuh/show_me_the_most_illegal_rust_code_youve_ever/fo88z2d?utm_source=share&utm_medium=web2x"
+description = """
+This crate allows calling unsafe functions without using the keyword "unsafe". It further
+deliberately makes this undetectable with cargo-geiger.
+
+In the API docs the author also states their intend to disable `#![forbid(unsafe)]`. No
+production code should ever have it in their dependency graph.
+"""
+[versions]
+patched = ["none"]
```
tarcieri, line 15: Try an empty array here?

```suggestion
patched = []
```

najamelan:
Don't know why I didn't think of that. Seems to have worked. Thanks.

tarcieri:
Assigned `RUSTSEC-2020-0011` in #271

Issue #207, opened 2019-12-12
`fake-static` needs advisory


Shnatsel:
https://crates.io/crates/fake-static is created as a demonstration of a soundness bug in rustc that lets you invoke undefined behavior in safe code. The entire point of the code is to obtain use-after-free without writing `unsafe`. No production code should ever depend on this.

PR #270, opened 2020-04-23
Add advisory for fake-static

eduardosm:
Closes https://github.com/RustSec/advisory-db/issues/207

### DIFF
--- a/crates/fake-static/RUSTSEC-0000-0000.toml
+++ b/crates/fake-static/RUSTSEC-0000-0000.toml
@@ -1,0 +1,17 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "fake-static"
+date = "2020-04-24"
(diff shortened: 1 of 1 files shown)


crates/fake-static/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,18 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "fake-static"
+date = "2020-04-24"
+title = "fake-static allows converting any reference into a `'static` reference"
+keywords = ["ssl", "mitm"]
```
MikailBag, line 6: these `keywords` are probably wrong

eduardosm: oops, I forgot to remove them from the template

tarcieri:
Assigned `RUSTSEC-2020-0013` in #273

PR #276, opened 2020-04-24
Remove security advisory about `plutonium`.

### LINKED ISSUE #275: RUSTSEC-2020-0011 is not a security vulnerability.
PR #268 added a security advisory related to `plutonium`, a crate that hides `unsafe` usage.

However, the security advisory does not report a security issue with the crate, or any defect with its intended functionality, but makes a value judgement about whether the crate's intended behavior is good.

Security advisories are not for "crates we don't like", they're for conveying information about defects in intended behavior.

hdevalence:
Closes #275 by reverting the RUSTSEC assignment.


tarcieri:
By policy we never remove vulnerabilities once assigned, however you can mark it as “obsolete: true” to perform the equivalent of yanking it

sunjay:
@tarcieri that is a reasonable policy in general, but do you think it still applies given that this was never an actual security vulnerability? It was also just created so removing it should not be a big issue. We can leave the number unused too if that's a concern. 

If you're really adamant on enforcing the rule then `obsolete` can be fine, but it seems like this is a special case that merits just removing it altogether. What do you think?

tarcieri:
@sunjay as a general principle we’re trying to follow how crates.io itself works: an immutable archive of vulnerabilities. As soon as we assign an identifier, it shouldn’t simply disappear: if there was a problem with the original assignment our goal is to document what happened, which is a generally recognized DFIR practice.

sunjay:
Okay. I get that. I think I was leaning towards special case because I don't want this to affect the integrity of the database. Obsolete is fine too and I think documenting what happened is very important. :)

tarcieri:
Dup of #280

PR #277, opened 2020-04-25
warn about CVE-2020-1967

lzutao:


crates/openssl-src/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,18 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "openssl-src"
+date = "2020-04-25"
+title = "Invalid or unrecognised signature algorithm is received from the peer"
+url = "https://www.openssl.org/news/secadv/20200421.txt"
+categories = ["crypto-failure", "denial-of-service"]
```
alex, line 7: I don't think it's correct to say this is a crypto-failure. While the bug happens in a crypto protocol, there's no actual issue with the cryptograpy.

lzutao: my mistake.

tarcieri:
Assigned `RUSTSEC-2020-0015` in #286

MikailBag:
Does it affect `openssl` crate users? If it does, which functions are affected?

lzutao:
Yes if users use the vendored feature: https://github.com/sfackler/rust-openssl/blob/e446d819e32fb9e9e1debf645b7f741c125a76e5/openssl/Cargo.toml#L19

PR #279, opened 2020-04-25
RUSTSEC-2020-0011: make wording more objective

CAD97:
I still personally believe this advisory should be marked obsolete with a note in the description along the lines of explaining how it got in and pointing people to use `cargo-crev` for this kind of concern.

However, this rewording of the advisory seems objectively better and more correct about what `plutonium` actually offers.


tarcieri:
Thanks!

PR #283, opened 2020-05-02
Add `cargo-deny` to the list of RustSec clients

Veetaha:
`cargo-deny` is an awesome tool, it seems to be the superset of `cargo-audit`. I think it is reasonable to mention it here along with `cargo-audit`.
cc @Jake-Shadle


tarcieri:
Thanks! `cargo-deny` is awesome!

PR #295, opened 2020-05-13
tempdir crate has been deprecated since 2018-02-13

oherrala:
The functionality of `tempdir` crate has been merged into `tempfile` which should be used instead.

### DIFF
--- a/crates/tempdir/RUSTSEC-0000-0000.toml
+++ b/crates/tempdir/RUSTSEC-0000-0000.toml
@@ -1,0 +1,15 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tempdir"
+date = "2018-02-13" # date when deprecation was announced on project README
+title = "`tempdir` crate has been deprecated; use `tempfile` instead"
+informational = "unmaintained"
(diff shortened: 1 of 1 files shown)


crates/tempdir/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,15 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tempdir"
+date = "2018-02-13" # date when deprecation was announced on project README
+title = "`tempdir` crate has been deprecated; use `tempfile` instead"
+informational = "unmaintained"
+url = "https://github.com/rust-lang-deprecated/tempdir/pull/46"
+description = """
+The [`tempdir`](https://crates.io/crates/tempdir) crate has been deprecated
```
lzutao, line 9: The link is needless as `cargo-audit` implies crate from crates.io.

oherrala: The links are in description field so they will provide helpful links for everyone browsing https://rustsec.org/advisories/

For example: https://rustsec.org/advisories/RUSTSEC-2020-0010.html

I can remove the links if that's wanted.

PR #306, opened 2020-05-28
Add advisory for internment

jsgf:
`internment` 0.3.12 has a race condition in ArcIntern::drop which can
result in use-after-free.

### DIFF
--- a/crates/internment/RUSTSEC-0000-0000.toml
+++ b/crates/internment/RUSTSEC-0000-0000.toml
@@ -1,0 +1,24 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+
+package = "internment"
+
+date = "2020-05-28"
+
+title = "Use after free in ArcIntern::drop"
+
+url = "https://github.com/droundy/internment/issues/11"
+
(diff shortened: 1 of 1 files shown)


crates/internment/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,26 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+
+package = "internment"
+
+# Disclosure date of the advisory as an RFC 3339 date (mandatory)
```
alex, line 6:
```suggestion
```

crates/internment/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,26 @@
...
+date = "2020-05-28"
+
+# Single-line description of a vulnerability (mandatory)
```
alex, line 9:
```suggestion
```

jsgf:
ping?

tarcieri:
Assigned `RUSTSEC-2020-0017` in #309

PR #308, opened 2020-06-01
Add advisory for traitobject

eduardosm:


tarcieri:
This seems like another candidate for an `unsound` informational advisory categorization, since it doesn't represent an immediate security vulnerability

PR #315, opened 2020-06-22
Add tiny-http HTTP Request Smuggling

snoopysecurity:
Further details can be seen here: https://github.com/tiny-http/tiny-http/issues/173


crates/tiny-http/RUSTSEC-2020-0001.toml
```diff
@@ -0,0 +1,20 @@
+[advisory]
+id = "RUSTSEC-2020-0001"
```
alex, line 2: This should be ID 0000

PR #317, opened 2020-06-23
make memoffset advisory informational

RalfJung:
To cause the drop of uninit memory clients had to panic in a `Deref` implementation. That sounds like a sufficiently "uncommon" library usage to me that this probably should be just informational.


tarcieri:
We need to update the `rustsec-admin` linter to allow `unsound` as an `informational` advisory type. Let me do that and cut another release (we've needed to do that for the new `assign-id` command anyway)

tarcieri:
@RalfJung can you rebase so it pulls in the new linter? /cc @alex

RalfJung:
Sure, done.

PR #318, opened 2020-06-29
Informational advisory for rust-lang-nursery/failure#336

Qwaz:
The current `Failure` API allows safe Rust code to implement malfunctioning `__private_get_type_id__` and cause type confusion when downcasting, which is an undefined behavior.

https://github.com/rust-lang-nursery/failure/issues/336

### DIFF
--- a/crates/failure/RUSTSEC-0000-0000.toml
+++ b/crates/failure/RUSTSEC-0000-0000.toml
@@ -3,14 +3,14 @@ id = "RUSTSEC-0000-0000"
 package = "failure"
 date = "2019-11-13"
 informational = "unsound"
-title = "__private_get_type_id__ can be overriden with safe Rust code"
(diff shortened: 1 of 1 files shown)


crates/failure/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,20 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "failure"
+date = "2019-11-13"
+informational = "unsound"
+title = "__private_get_type_id__ can be overriden with safe Rust code"
```
Shnatsel, line 6: I believe something like `Type confusion if __private_get_type_id__ is overriden` would communicate the impact more clearly.

Qwaz: PR updated :)

Shnatsel:
Merging. Thanks!

Issue #313, opened 2020-06-22
Clarify what it means for libraries to have a security vulnerabilities


Qwaz:
Continued from the discussion from #300.

The current version of [CONTRIBUTING.md](https://github.com/RustSec/advisory-db/blob/5049594b7539974e9fe794433d390a971cfd1042/CONTRIBUTING.md#criteria) doesn't specify what it means for libraries to have a security bug. Here, following the practice of NPM and other security advisories, I suggest an initial definition of "a library is considered to have a security bug if there is a known usage of its safe APIs which creates a security vulnerability." I very much appreciate a discussion and a feedback on this definition.

---

Repeating here from #300 with a slight revision:

Any reachable exploitable bug in a program qualifies as a security vulnerability. For executable files, the judgement is easy. If there is a program path that handles untrusted input (file, network socket, etc.) and triggers an exploitable bug, it should be considered as a security bug. However, for libraries, there is one more layer of indirection. A user of a library decide how to use APIs in their program, and whether or not a bug in API becomes an exploitable bug depends on how the user structure their program. To my best knowledge, such bugs are handled as security vulnerabilities by many existing security advisories such as CVE and NPM.

[Node.js Ecosystem Triage Guidelines](https://github.com/nodejs/security-wg/blob/master/processes/third_party_triage_guidelines.md) defines a library vulnerability as:

> if a documented or obvious way to use a library leads to an exploitable vulnerability in the correct and safe calling code, then those defects are also vulnerabilities. Some APIs are unsafe to use and are not vulnerabilities if they are clearly marked this way and if safe alternatives exist. An excellent example of this is dangerouslySetInnerHTML in React.

What's not described here is how NPM handles a potential security bug in *uncommon usage* of libraries. The answer is that they are also treated as security vulnerabilities, and NPM assigns different severity score depends on how likely those APIs will be used in a vulnerable way. CVE considers the similar aspect when assigning CVSS score, too.

For example, take a look at these two NPM security advisories:

* [SQL injection in connect-pg-simple](https://www.npmjs.com/advisories/1153)
* [SQL injection in sequelize](https://www.npmjs.com/advisories/1146)

They are both classified as SQL injection bugs. The first bug was categorized as low severity because the vulnerable variables are unlikely to be controlled by user input, while the second bug was categorized as high severity because there are common legit scenarios where the bug can be triggered by an attacker.

[...]

PR #320, opened 2020-06-29
Informational advisory for SergioBenitez/Rocket#1312

Qwaz:
The affected version of `rocket` contains a `Clone` trait implementation of `LocalRequest` that reuses the pointer to inner `Request` object. This causes data race in rare combinations of APIs if the original and the cloned objects are modified at the same time.

https://github.com/SergioBenitez/Rocket/issues/1312

### DIFF
--- a/crates/rocket/RUSTSEC-0000-0000.toml
+++ b/crates/rocket/RUSTSEC-0000-0000.toml
@@ -1,0 +1,20 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rocket"
+date = "2020-05-27"
+title = "`LocalRequest::clone` creates multiple mutable references to the same object"
+url = "https://github.com/SergioBenitez/Rocket/issues/1312"
+description = """
+The affected version of `rocket` contains a `Clone` trait implementation of
+`LocalRequest` that reuses the pointer to inner `Request` object.
+This causes data race in rare combinations of APIs if the original and the
+cloned objects are modified at the same time.
+"""
+
+[affected]
+# TODO: confirm that this path is correct
+functions = { "rocket::local::LocalRequest::Clone::clone" = ["< 0.4.5, >= 0.4.0"] }
+
+[versions]
+patched = [">= 0.4.5"]
+unaffected = ["< 0.4.0"]
--- a/crates/rocket/RUSTSEC-0000-0000.toml
+++ b/crates/rocket/RUSTSEC-0000-0000.toml
@@ -2,6 +2,7 @@
 id = "RUSTSEC-0000-0000"
 package = "rocket"
 date = "2020-05-27"
+informational = "unsound"
 title = "`LocalRequest::clone` creates multiple mutable references to the same object"
 url = "https://github.com/SergioBenitez/Rocket/issues/1312"
 description = """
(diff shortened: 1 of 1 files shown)


Qwaz:
I'm not sure how to represent `LocalRequest`'s `Clone` implementation as path. I put `rocket::local::LocalRequest::Clone::clone`, but this looks awkward?

8573:
I don't think `...::LocalRequest::Clone::...` is right — `Clone` is not in `LocalRequest`'s namespace. I think the proper way in Rust itself to be specific about selecting the `clone` method from the `Clone` trait rather than any other `clone` method of `LocalRequest` would be to write `<rocket::local::LocalRequest as Clone>::clone` (a [qualified path](https://doc.rust-lang.org/reference/paths.html#qualified-paths)), but, looking at [the `rustsec-crate` source code](https://github.com/RustSec/rustsec-crate/blob/6a66f37cb6838f54aeb86871479dc7063de304fa/src/advisory/affected.rs), I think the most specific path supported here is `rocket::local::LocalRequest::clone`.

---

Although @SergioBenitez said "I am happy to have a RustSec advisory filed", I wonder, should this advisory be `informational = "unsound"` rather than a full security advisory, given your remarks on "Is this a security bug?" in <https://github.com/SergioBenitez/Rocket/issues/1312>?

Qwaz:
I consider the severity of this bug is somewhere between low to medium according to [npm's scale](https://docs.npmjs.com/about-audit-reports#severity). There is an ongoing discussion in #313 to determine the boundary between informational advisories and security advisories, and we can decide which type of advisory to file based on where the discussion settle on. I'm also fine with filing an "unsound" informational advisory first and later considering an upgrade.

Thank you for your feedback on the path question. I'll wait and see if any maintainer can confirm that.

8573:
> I consider the severity of this bug is somewhere between low to medium according to [npm's scale](https://docs.npmjs.com/about-audit-reports#severity).

Okay, then I suppose I misunderstood the sentences—

> I believe the security impact of this bug is minimal. I think It is still worthwhile to file a RustSec advisory after the bug confirmation considering the people who want to be notified about unsound APIs.

—as having meant "This doesn't seem important as a security bug, but it should have an advisory as a soundness bug"; I apologize for the misunderstanding.  Personally I don't mind being notified about the bug regardless of which type it is, and, with your meaning clarified, I make no objection to (and, either way, my approval is not needed for) filing the advisory without the `informational` field.

Qwaz:
The bug was reported before the existence of unsound informational advisory, so I was not particularly relating the report to the unsound advisory at that time.

SergioBenitez:
The APIs exposed by the `local` module, including `LocalRequest`, are [documented](https://api.rocket.rs/v0.4/rocket/local/index.html) and intended primarily for and nearly exclusive use during testing. As a result, the possibility of the unsound implementation leading to a vulnerability should be nil.

Nevertheless, these APIs are not concretely restricted to testing environments only, and so there exists a possibility, even if remote, that the APIs find themselves in use in a vulnerable path. However, in such a path, a `LocalRequest` request has no utility being cloned, and even less so prior to being populated, as required to effectuate this bug. Consequently, again, the possibility of a vulnerability arising should be nil.

Certainly, the implementation was unsound. However, based on the reasoning above, no inherent security vulnerability can  arise from documented, expected, or probable use of the APIs. As such, I would advocate for publication as an informational advisory.

Qwaz:
Thank you for the comment, @SergioBenitez. I've fixed the function path and made the advisory informational.

Shnatsel:
Apologies for overlooking this PR. It's merged now.

PR #322, opened 2020-06-29
Attempt to fix failing assign-ids task

alex:


alex:
@tarcieri I think this is the fix, based on https://github.com/peter-evans/create-pull-request/issues/122

alex:
@tarcieri fix appeared to work! let's merge something and see if it works for real!

tarcieri:
Cool, sounds good

PR #328, opened 2020-07-04
Security advisory for bqv/ozone

Qwaz:
Ozone contains several memory safety issues including [out-of-bound access](https://github.com/bqv/ozone/blob/e21f948b0178ab305f644118f18d87a838c618e0/src/buffer.rs#L38-L48) and dropping of [uninitialized memory](https://github.com/bqv/ozone/blob/e21f948b0178ab305f644118f18d87a838c618e0/src/map.rs#L94-L101).

---

I'm not sure whether this crate is popular enough to file an advisory. Is there any guideline in that aspect?

### DIFF
--- a/crates/ozone/RUSTSEC-0000-0000.toml
+++ b/crates/ozone/RUSTSEC-0000-0000.toml
@@ -1,0 +1,12 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "ozone"
+date = "2020-07-04"
+title = "Ozone contains several memory safety issues"
+description = """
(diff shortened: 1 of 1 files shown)


alex:
We don't generally limit advisories based on crate popularity -- if no one is using the crate, then no one will be bothered by it :-)

Shnatsel:
LGTM. Thanks for all your recent contributions!

I'm not sure we're responding sufficiently timely on github. If you want to have a more direct channel with the RustSec maintainers, you can join the [Rust Zulip](https://rust-lang.zulipchat.com/#narrow/stream/146229-wg-secure-code). We can be found in `#wg-secure-code`

PR #332, opened 2020-07-09
tough: CVE-2020-15093 / GHSA-5q2r-92f9-4m49

iliana:
https://github.com/awslabs/tough/security/advisories/GHSA-5q2r-92f9-4m49

### DIFF
--- a/crates/tough/RUSTSEC-0000-0000.toml
+++ b/crates/tough/RUSTSEC-0000-0000.toml
@@ -1,0 +1,29 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tough"
+date = "2020-07-09"
+title = "Improper uniqueness verification of signature threshold"
+url = "https://github.com/awslabs/tough/security/advisories/GHSA-5q2r-92f9-4m49"
+aliases = ["CVE-2020-15093", "GHSA-5q2r-92f9-4m49"]
+references = ["CVE-2020-6174"]
+description = """
+The tough library, prior to 0.7.1, does not properly verify the uniqueness of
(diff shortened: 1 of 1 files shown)


crates/tough/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,30 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tough"
+date = "2020-07-09"
+title = "Improper uniqueness verification of signature threshold"
+url = "https://github.com/awslabs/tough/security/advisories/GHSA-5q2r-92f9-4m49"
+aliases = ["CVE-2020-15093"]
```
tarcieri, line 7: You can also add `”GHSA-5q2r-92f9-4m49”` here

crates/tough/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,30 @@
...
+CVE-2020-6174 is assigned to the same issue in the TUF reference
+implementation.
+
+If you have any questions or comments about this advisory, contact AWS Security
+at aws-security@amazon.com.
+"""
+
+[versions]
+patched = [">= 0.7.1"]
+unaffected = ["< 0.7.0"]
```
Shnatsel, line 30: On the surface this contradicts the statement above about versions prior to 0.7.1 being affected. I assume the relevant functionality wasn't implemented in earlier versions?

iliana: Oh I misread that word, let me fix.

PR #335, opened 2020-07-11
libpnet

0xdeafbeef:
It's not patched yet, and we are looking for somebody, who can help us.

### DIFF
--- a/crates/pnet/RUSTSEC-0000-0000.toml
+++ b/crates/pnet/RUSTSEC-0000-0000.toml
@@ -1,6 +1,6 @@
 [advisory]
 id = "RUSTSEC-0000-0000"
-package = "libpnet"
+package = "pnet"
 date = "2019-06-11"
 title = "Compiler optimisation for next_with_timeout in pnet::transport::IcmpTransportChannelIterator flaws to SEGFAULT"
 url = "https://github.com/libpnet/libpnet/issues/449"
--- a/crates/libpnet/RUSTSEC-0000-0000.toml
+++ b/crates/libpnet/RUSTSEC-0000-0000.toml
@@ -1,0 +1,16 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "libpnet"
+date = "2019-06-11"
+title = "Compiler optimisation for next_with_timeout in pnet::transport::IcmpTransportChannelIterator flaws to SEGFAULT"
+url = "https://github.com/libpnet/libpnet/issues/449"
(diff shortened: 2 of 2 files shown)


crates/pnet/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,18 @@
...
+id = "RUSTSEC-0000-0000"
+package = "pnet"
+date = "2019-06-11"
+title = "Compiler optimisation for next_with_timeout in pnet::transport::IcmpTransportChannelIterator flaws to SEGFAULT"
+url = "https://github.com/libpnet/libpnet/issues/449"
+categories = ["memory-corruption"]
+cvss = "CVSS:3.0/AV:L/AC:L/PR:H/UI:N/S:U/C:N/I:H/A:H"
+keywords = ["segfault"]
+description = """
+Affected versions of this crate were optimized out by compiler,
```
RalfJung, line 11: Is this a compiler bug or is your code triggering [Undefined Behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html)?

RalfJung: Ah looks like the bug was found: https://github.com/libpnet/libpnet/issues/449#issuecomment-663355987

crates/pnet/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,18 @@
...
+which caused dereference of uninitialized file descriptor which caused segfault.
+"""
+[affected.functions]
+"pnet::transport::IcmpTransportChannelIterator" = ["< 0.26.0"]
+
+[versions]
+patched = [">= 0.26.0"] # no patch exists yet
```
taiki-e, line 18: I don't know if it is included in this repo's guideline, but I do not recommend using a non-exist version in the `patched` field and `affected` section if it has not actually been fixed.
`pnet` released version [0.26.0](https://github.com/libpnet/libpnet/releases/tag/v0.26.0), which does not include a fix for the vulnerability (https://github.com/libpnet/libpnet/pull/455).

tarcieri: Good point. This should be removed.

mrmonday: This should be fixed in `v0.27.2`.

taiki-e: Filed #494

Issue #299, opened 2020-05-15
SmallVec < 0.6.13 is unsound


RalfJung:
https://github.com/servo/rust-smallvec/pull/162 fixes a soundness bug in SmallVec where `mem::uninitialized` was used incorrectly. This fix is included in 0.6.13, but previous versions are unsound. The latest version is 1.4, but given how widely used this crate is, it might still be worth having an advisory for the old broken versions.

The advisory should probably follow whatever policy ends up being in place for soundness bugs.

PR #337, opened 2020-07-24
add smallvec unsoundness

RalfJung:
Fixes https://github.com/RustSec/advisory-db/issues/299

### DIFF
--- a/crates/smallvec/RUSTSEC-0000-0000.toml
+++ b/crates/smallvec/RUSTSEC-0000-0000.toml
@@ -1,0 +1,18 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "smallvec"
+date = "2018-09-25"
+informational = "unsound"
+
+title = "smallvec creates uninitialized value of any type"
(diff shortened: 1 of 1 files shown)


crates/smallvec/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,18 @@
...
+
+description = """
+Affected versions of this crate called `mem::uninitialized()` to create values of a user-supplied type `T`.
+This is unsound e.g. if `T` is a reference type (which must be non-null and thus may not remain uninitialized).
+ 
+The flaw was corrected by avoiding the use of `mem::uninitialized()`, using `MaybeUninit` instead.
+"""
+
+[versions]
+patched = [">= 0.6.13"]
```
alex, line 18: Is this right? The commit that closes the issue looks like it's only on 1.0 (was it backported)?

RalfJung: Yes it was backported: https://github.com/servo/rust-smallvec/pull/180

Issue #345, opened 2020-07-31
Guidelines for GitHub Advisory Database


vks:
Does it make sense to provide guidelines for the GitHub Advisory Database? It could be mentioned in `CONTRIBUTING.md` as an optional step.

tarcieri:
Sure, what did you have in mind specifically?

One problem is while the `rustsec` crate supports GHSA identifiers as aliases, there's to my knowledge no way to map one of those identifiers to an advisory automatically.

vks:
I was wondering whether we should file a GitHub advisory for `rand_core` (RUSTSEC-2019-0035, which is informal). Then I thought that there could be some guidance here.

tarcieri:
Sure, and once you do, we can add an alias for its GHSA in the RUSTSEC advisory.

vks:
@tarcieri Ok, I opened a draft and added you as a collaborator. It's GHSA-mmc9-pwm7-qj5w (not published yet).

It seems this would be the first crates.io GHSA and that we can ask GitHub to file a CVE for us. (I think this could be mentioned in the RUSTSEC guidelines.) However, for an informational advisory it's maybe not necessary to assign a CVE?

tarcieri:
Up to you about a CVE. I'd recommend against it in this case.

Advisory looks good. I'd suggest opening a PR that modifies `RUSTSEC-2019-0035.toml` to list `GHSA-mmc9-pwm7-qj5w ` as an `alias`.

PR #346, opened 2020-08-01
Add alias for GHSA

vks:
Refs #345.


crates/rand_core/RUSTSEC-2019-0035.toml
```diff
@@ -1,5 +1,6 @@
 [advisory]
 id = "RUSTSEC-2019-0035"
+alias = "GHSA-mmc9-pwm7-qj5w"
```
tarcieri, line 3: Whoops sorry, this needs to be:

```suggestion
aliases = ["GHSA-mmc9-pwm7-qj5w"]
```

PR #349, opened 2020-08-14
Add advisory for rgb

Shnatsel:


Shnatsel:
@HeroicKatora I'd appreciate your input on this. Notably I'm filing this as an informational advisory and not an exploit.

Shnatsel:
I consider 0.5.4 the first one affected because the relevant methods seem to be introduced in this commit: https://github.com/kornelski/rust-rgb/commit/d11ef3989e727f8c73440f38051b9388540f5ad0

crates/rgb/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,21 @@
...
+date = "2020-10-14"
+title = "Allows viewing and modifying arbitrary structs as bytes"
+url = "https://github.com/kornelski/rust-rgb/issues/35"
+informational = "unsound"
+keywords = ["type confusion"]
+description = """
+Affected versions of rgb crate allow viewing and modifying data any type `T` wrapped in `RGB<T>` as bytes,
+and do not correctly constrain `RGB<T>` and other wrapper structures to the types for which it is safe to do so.
+
+If a type containing madding is wrapped in `RGB<T>` and similar wrapper structures,
```
HeroicKatora, line 13:
```suggestion
If a type containing padding is wrapped in `RGB<T>` and similar wrapper structures,
```

crates/rgb/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,21 @@
...
+viewing it as bytes may lead to exposure of contents of uninitialized memory.
+
+If a type containing a pointer is wrapped in `RGB<T>` and similar wrapper structures,
+modifying it as bytes may lead to derefericing of arbitrary pointers.
```
HeroicKatora, line 17:
```suggestion
modifying it as bytes may lead to dereferencing of arbitrary pointers.
```

crates/rgb/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,21 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rgb"
+date = "2020-10-14"
+title = "Allows viewing and modifying arbitrary structs as bytes"
+url = "https://github.com/kornelski/rust-rgb/issues/35"
+informational = "unsound"
+keywords = ["type confusion"]
+description = """
+Affected versions of rgb crate allow viewing and modifying data any type `T` wrapped in `RGB<T>` as bytes,
```
HeroicKatora, line 10:
```suggestion
Affected versions of rgb crate allow viewing and modifying data of any type `T` wrapped in `RGB<T>` as bytes,
```

Shnatsel:
Better now?

crates/rgb/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,23 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rgb"
+date = "2020-10-14"
```
lzutao, line 4: in the future?

Shnatsel: Gah, my brain must have melted from all the heat. Sorry. Will fix that now.

Shnatsel: Fixed in git master. Thanks for spotting!

PR #352, opened 2020-08-17
Suggest lz-fear as an alternative to lz4-compress

Shnatsel:


crates/lz4-compress/RUSTSEC-2017-0007.toml
```diff
@@ -9,6 +9,8 @@ description = """
 [According to the developers](https://gitlab.redox-os.org/redox-os/tfs/issues/89) this crate is no longer maintained.
 
 The suggested alternative is [`lz4-compression`](https://crates.io/crates/lz4-compression), a maintained fork of `lz4-compress`.
+
+See also [lz-fear](https://crates.io/crates/lz-fear) which is compatible with the reference LZ4 implementation in C, but not with lz4-compress.
```
tarcieri, line 13: Maybe make this into a bulleted list like some of the other advisories, e.g.

https://github.com/RustSec/advisory-db/blob/master/crates/spin/RUSTSEC-2019-0031.toml#L11-L16

Shnatsel: I didn't want to make it into a bulleted list because the two options are really quite distinct, and I don't want to put them on the same level. If another LZ4 crate pops up, I'll make a bulleted list.

PR #353, opened 2020-08-19
Missing sanitazion in mozwire allows local file overwrite of files ending in .conf

alexanderkjall:
fix committed but not released yet.

### DIFF
--- a/crates/mozwire/RUSTSEC-0000-0000.toml
+++ b/crates/mozwire/RUSTSEC-0000-0000.toml
@@ -1,0 +1,20 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "mozwire"
+date = "2020-08-18"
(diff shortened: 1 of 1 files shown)


Shnatsel (approved):
Thanks for the report! Looks good to me.

@tarcieri do you see any issues with merging advisories before the version mentioned as patched is actually released?

alexanderkjall:
It would maybe be good to add a category for path traversal problems?

tarcieri:
@Shnatsel the only problem is it gives people no recourse as to how to solve the vulnerability, but it wouldn't be the first time.

@alexanderkjall we already have local file disclosure. We could also add unrestricted file upload.

The problem with "path traversal" / "directory traversal" is it's ambiguous as to what capabilities it actually gives the attacker.

(Also this is where we should probably implement CWE, but that's a somewhat nontrivial endeavor)

Shnatsel:
0.4.1 is released, merging.

PR #360, opened 2020-08-25
StrcCtx deallocates a memory region that it doesn't own

Qwaz:
`StrcCtx` deallocate a memory region that it doesn't own when `StrcCtx` is created without using `StrcCtx::new`.
This can introduce memory safety issues such as double-free and use-after-free to client programs.

Original issue report: https://github.com/pigeonhands/rust-arch/issues/2


Shnatsel:
Merging. Thanks!

PR #362, opened 2020-08-25
Matrix::new() drops uninitialized memory

Qwaz:
`Matrix::new()` internally calls `Matrix::fill_with()` which uses `*ptr = value` pattern to initialize the buffer.
This pattern assumes that there is an initialized struct at the address and drops it, which results in dropping of uninitialized struct.

Original issue report: https://gitlab.com/dvshapkin/alg-ds/-/issues/1


Shnatsel:
Thanks!

PR #364, opened 2020-08-25
arr: Multiple security issues including data race, buffer overflow, and uninitialized memory drop

Qwaz:
`arr` crate contains multiple security issues. Specifically,

1. It incorrectly implements Sync/Send bounds, which allows to smuggle non-Sync/Send types across the thread boundary.
2. `Index` and `IndexMut` implementation does not check the array bound.
3. `Array::new_from_template()` drops uninitialized memory.

Original issue report: https://github.com/sjep/array/issues/1

### DIFF
--- a/crates/arr/RUSTSEC-0000-0000.toml
+++ b/crates/arr/RUSTSEC-0000-0000.toml
@@ -1,0 +1,16 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "arr"
+date = "2020-08-25"
+title = "Multiple security issues including data race, buffer overflow, and uninitialized memory drop"
+url = "https://github.com/sjep/array/issues/1"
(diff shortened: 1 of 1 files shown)


Shnatsel:
Dropping ununit memory seems to be a common theme in your recent reports. I wonder if there's a specific pattern that clippy or rustc could lint against?

Thanks again!

Qwaz:
I've been recently working on common undefined behavior bug detection for Rust crates. It is still preliminary, but the initial run on crates.io seems promising as the recent stream of PRs show. I'm expecting more PRs like these, and please leave a comment here if you have any feedback as a RustSec maintainer :)

https://rust-lang.zulipchat.com/#narrow/stream/146229-wg-secure-code/topic/reporting-preference

PR #366, opened 2020-08-26
chunky: Chunk API does not respect align requirement

Qwaz:
Chunk API does not respect the align requirement of types. Unaligned reference can be created with the API, which is an undefined behavior.

Original issue report: https://github.com/aeplay/chunky/issues/2


Shnatsel:
Thanks again!

PR #373, opened 2020-09-03
obstack: Obstack generates unaligned references

Qwaz:
Obstack generates unaligned references for types that require a large alignment.

Original issue report: https://github.com/petertodd/rust-obstack/issues/4


petertodd:
Fixed: https://github.com/petertodd/rust-obstack/issues/4#issuecomment-687337872, and released a new version with the fix: https://crates.io/crates/obstack/0.1.4

Shnatsel:
Thanks for the report, and thanks for the swift fix! I've marked 0.1.4 and later as fixed in the advisory. Merging now.

PR #381, opened 2020-09-06
sized-chunks: Multiple soundness issues in Chunk and InlineArray

Qwaz:
Chunk:

* Array size is not checked when constructed with `unit()` and `pair()`.
* Array size is not checked when constructed with `From<InlineArray<A, T>>`.
* `Clone` and `insert_from` are not panic-safe; A panicking iterator causes memory safety issues with them.

InlineArray:

* Generates unaligned references for types with a large alignment requirement.

Original issue report: https://github.com/bodil/sized-chunks/issues/11


Shnatsel:
Thanks!

PR #389, opened 2020-09-20
added CVE numbers

alexanderkjall:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-25791
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-25792
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-25793
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-25794
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-25795
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-25796

### DIFF
--- a/crates/sized-chunks/RUSTSEC-2020-0041.toml
+++ b/crates/sized-chunks/RUSTSEC-2020-0041.toml
@@ -1,5 +1,6 @@
 [advisory]
 id = "RUSTSEC-2020-0041"
+aliases = ["CVE-2020-25791", "CVE-2020-25792", "CVE-2020-25793", "CVE-2020-25794", "CVE-2020-25795", "CVE-2020-25796"]
 package = "sized-chunks"
 date = "2020-09-06"
 title = "Multiple soundness issues in Chunk and InlineArray"


Shnatsel:
Wow, that's a lot of CVEs!

alexanderkjall:
Yeah, the reason behind all these patches are that I discovered a process bug, Debian's security work is very much based on CVE numbers, but not all rust advisories have CVE numbers, so they are missed by Debian.

I wrote a quick hack to discover those ( https://github.com/alexanderkjall/deb-rust-sec ) and reported them to Debian, and then Debian applied for CVE's for them.

Maybe it would be a good idea that you applied for CVE's when you merge a new advisory? I guess there is more places than Debian that might miss security holes if they don't have a CVE.

Shnatsel:
Hmm, that makes sense. Right now the onus is on the reporter to apply for a CVE, but nobody does that in practice. Meanwhile DB maintainers can be easily overwhelmed by the volume. I wonder if we can become a CVE ID issuer, like DWF did?

I wonder, how does Debian apply for CVEs? Is it through iwantacve.org or do they have some other process?

alexanderkjall:
Here my knowledge becomes a bit shaky, but they seem to have it's own pool according to their FAQ: https://security-team.debian.org/security_tracker.html#cve-pool-from-debian

But I think these numbers are from mitre, at least that's where Salvatore pointed me, see here for reference: https://salsa.debian.org/security-tracker-team/security-tracker/-/merge_requests/70

Shnatsel:
Ah, so Debian is a CNA then. Interesting. That's definitely something worth looking into for us as well.

tarcieri:
> Maybe it would be a good idea that you applied for CVE's when you merge a new advisory?

We’ve tried in the past (#20) and it’s been very difficult for us to consistently be able to get CVEs. We’ve applied for our own CVE block only to never get a response.

Difficulty in obtaining CVEs consistently on similar projects is the reason why this database uses its own identifier format. We’d like for every advisory to have a CVE assigned but it’s been awful when it becomes a blocker to publishing an advisory.

If someone has access to a CVE block and can readily obtain them and wants to be “point” on getting CVE assignments, that’d be great.

Shnatsel:
Perhaps it's time to re-apply for being a CNA, now that RustSec has a proven track record and is backed by an official working group?

pabs3:
Until that happens, you could use the newly revived DWF: 

https://opensourcesecurity.io/2021/03/07/episode-261-dwf-is-back-welcome-to-community-powered-cve/

tarcieri:
I applied to DWF in 2017 and never heard back. Can try again, I guess...

setharnold:
Given how easy it is to become a CNA these days https://cve.mitre.org/cve/cna.html#become_a_cna I'd like to suggest taking that path instead. I'll admit the process of submitting CVEs is a bit heavier-weight than filling out MITRE's form, but the hard part is almost always the actual work: describing the flaw, describing affected versions, linking in fixes when they exist, etc. You'd have that same kind of work regardless of who actually does the submissions to the CVE database.

tarcieri:
Interesting. That might be worth taking a look at.

PR #390, opened 2020-09-21
Add advisory for atom crate

ammaraskar:
Upstream issue: https://github.com/slide-rs/atom/issues/13

It looks like this vulnerability was fixed in the Github code but a new release was never pushed to cargo with it. The code snippet in the upstream issue presents an `Rc` used across thread-boundaries. Since this library is meant to be for concurrent/multi-thread access, this mistake of accidentally sending across a non `Send` type seems like an easy one to make and can lead to memory corruption issues.

### DIFF
--- a/crates/atom/RUSTSEC-0000-0000.toml
+++ b/crates/atom/RUSTSEC-0000-0000.toml
@@ -1,0 +1,15 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "atom"
+date = "2020-09-21"
+informational = "unsound"
+title = "Unsafe Send implementation in Atom allows data races"
+url = "https://github.com/slide-rs/atom/issues/13"
+description = """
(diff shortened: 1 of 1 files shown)


crates/atom/RUSTSEC-0000-0000.toml
```diff
@@ -0,0 +1,14 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "atom"
+date = "2020-09-21"
+title = "Unsafe Send implementation in Atom allows data races"
```
tarcieri, line 5: I'd suggest adding `informational = "unsound"` since this advisory is not a security advisory in and of itself.

ammaraskar: Done.

Shnatsel:
All comments are addressed, merging.

PR #396, opened 2020-09-26
Add advisory for out-of-bounds read in array-queue.

ammaraskar:
Upstream issue: https://github.com/raviqqe/array-queue/issues/2

There's an indexing problem in the `pop_back()` function allowing a read from uninitialized or previously dropped memory. As demonstrated on the upstream issue, this can be leveraged for an arbitrary read and potentially write.

### DIFF
--- a/crates/array-queue/RUSTSEC-0000-0000.toml
+++ b/crates/array-queue/RUSTSEC-0000-0000.toml
@@ -2,15 +2,17 @@
 id = "RUSTSEC-0000-0000"
 package = "array-queue"
 date = "2020-09-26"
-title = "array_queue pop_back allows an out-of-bounds read."
+title = "array_queue pop_back() may cause a use-after-free"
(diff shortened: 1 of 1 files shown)


Shnatsel:
I'd rather describe it as a use-after-free in the title, but other than that looks good to me.

Thanks for reporting the issue, and for the detailed description in the upstream bug report!

Shnatsel:
I've added version specification to the advisory and changed the title to a use-after-free. Please let me know if this looks correct to you. If so, we'll merge.

I've also tried to specify the exact function affected, but looks like our linter won't allow that due to a bug. I'll address that separately.

ammaraskar:
Thank you for amending the title, I was torn between use-after-free or out-of-bounds-read but I think it's clearly the former, the latter is just a consequence.

Also, thank you for adding the unaffected version. This looks good to merge on my side.

PR #409, opened 2020-09-26
Add advisory for very old, unsound actix-web

Shnatsel:


Shnatsel:
cc @robjtede - adding an advisory for very unsafe actix-web back from 2018. No-one should be using it at this point

robjtede:
Seems about right but I don't feel able to comment further than that because I wasn't involved in the org at the time of the 0.x versions.

PR #411, opened 2020-09-27
Add misaligned-access soundness issue for dync crate

ammaraskar:
Upstream issue: https://github.com/elrnv/dync/issues/4

This is an unsoundness informational advisory. As shown under miri in the issue above, the `VecCopy` type in the `dync` crate allows for misaligned access as it treats everything as a `u8` underneath.


Shnatsel:
Thanks!

PR #422, opened 2020-10-06
Add patched version for atom crate.

ammaraskar:
As per https://github.com/slide-rs/atom/issues/13#issuecomment-704110075, version `0.3.6` of atom now contains the fix for the soundness issue.


Shnatsel:
Thanks!

PR #425, opened 2020-10-11
Add advisory for UB in crossbeam-channel 0.4.3

taiki-e:
The affected version of this crate's the `bounded` channel incorrectly assumes that `Vec::from_iter` has allocated capacity that same as the number of iterator elements. `Vec::from_iter` does not actually guarantee that and may allocate extra memory. The destructor of the `bounded` channel reconstructs `Vec` from the raw pointer based on the incorrect assumes described above. This is unsound and causing deallocation with the incorrect capacity when `Vec::from_iter` has allocated different sizes with the number of iterator elements.

Refs: https://github.com/crossbeam-rs/crossbeam/pull/533, https://github.com/crossbeam-rs/crossbeam/issues/539 (Both issues are mainly saying about crossbeam-queue, but crossbeam-queue including this UB, has not been released and only crossbeam-channel is affected.)

### DIFF
--- a/crates/crossbeam-channel/RUSTSEC-0000-0000.md
+++ b/crates/crossbeam-channel/RUSTSEC-0000-0000.md
@@ -1,0 +1,15 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "crossbeam-channel"
+date = "2020-06-26"
+url = "https://github.com/crossbeam-rs/crossbeam/pull/533"
+
+[versions]
+patched = [">= 0.4.4"]
+unaffected = ["< 0.4.3"]
+```
+
+# Undefined Behavior in bounded channel
+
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thanks for the report!

Is this purely theoretical at this point, or has deallocation of the incorrect size has actually been observed in practice? If it's the latter, I'd drop `informational = "unsound"`.

Last time I checked it didn't do that, but that was over a year ago and there have been some optimizations for small vector sizes, like the first allocation was always having size 4.

taiki-e:
> Is this purely theoretical at this point, or has deallocation of the incorrect size has actually been observed in practice? If it's the latter, I'd drop `informational = "unsound"`.

It has actually been observed in practice (https://twitter.com/khuey_/status/1311641831201857537, https://github.com/nervosnetwork/ckb/pull/2244), so I'll drop `informational` field.

> Last time I checked `from_iter` never over-allocated in practice, but that was over a year ago and there have been some optimizations for small vector sizes, like the first allocation was always having size 4.

If the number of elements is small, it seems that [0-7 extra capacity is allocated depending on the size of the elements](https://github.com/rust-lang/rust/blob/c38f001db5e16f1e0db47c5aed80394e9df5430a/library/alloc/src/raw_vec.rs#L408-L421). 
([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=5b713b9d4ff0733e6411988a55aabe1d))

Shnatsel:
Merging. Thanks!

Issue #282, opened 2020-04-29
`dirs` and `directories` crates unmaintained?


tarcieri:
Saw this here:

https://www.reddit.com/r/rust/comments/ga7f56/why_dirs_and_directories_repositories_have_been/

The `dirs` and `directories` (EDIT: and `dirs-sys`) crates have been archived on GitHub. ~~No one has been able to figure out why or heard back from the author~~ (edit: someone claims to have received an email from the author and the [situation seems bad](https://www.reddit.com/r/rust/comments/ga7f56/why_dirs_and_directories_repositories_have_been/foyqipp/))

Some stats on current usage:

- [`dirs`](https://github.com/soc/dirs-rs)
  - downloads: ~15,000/day
  - dependent crates: 523
- [`directories`](https://github.com/soc/directories-rs)
  - downloads: ~4,000/day
  - dependent crates: 116
- [`dirs-sys`](https://github.com/soc/dirs-sys-rs)
  - downloads: ~13,000/day
  - dependent crates: 3

I don't want to be too hasty in filing advisories as that seems rather disruptive and we should probably wait and see what's going on first and if there's some sort of handoff to new owners. But if they're really unmaintained, we should document that.

The [`home` crate](https://github.com/brson/home) is an alternative, but does not provide equivalent functionality (e.g. XDG directories, though it seems the [`xdg` crate](https://crates.io/crates/xdg) might fit the bill there)

lzutao:
I created @xdg-rs for maintaining those crates. Please join if you interest.

Currently we only have publishing permission for "dirs". 
I mailed and kindly asked Simon for ownership for other crates. He hasn't responded yet.
In worst case, we will have to publish the same crates under different names.

tarcieri:
Neat! Let us know how it works out.

Shnatsel:
A fork to maintain those crates has just been announced: https://www.reddit.com/r/rust/comments/gli4pc/directoriesnext_dirsnext_the_new_home_for/

ArekPiekarz:
Is dirs still unmaintained? Its repository is no longer read-only and there has been a new release on June 21.

tarcieri:
Looks like we can close this

tarcieri:
Update: we merged #430 that marks them as unmaintained again as the repos have been archived. It seems like `dirs-next` and `directories-next` are the successors.

PR #430, opened 2020-10-16
dirs and directories crates are unmaintained

melocene:
Both of these crates have had their repos archived and unarchived a few times which prompted the forks. The forked crates look to be drop in replacements and I have not run into issues with them so far.

### DIFF
--- a/crates/directories/RUSTSEC-0000-0000.md
+++ b/crates/directories/RUSTSEC-0000-0000.md
@@ -1,0 +1,16 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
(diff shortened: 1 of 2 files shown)


tarcieri:
That's unfortunate! See #282 for some previous discussion on these crates.

kpcyrd:
I'm really confused about the dirs situation, I remember when dirs-next was created because upstream was unresponsive, projects started depending on dirs-next instead, afterwards dirs became active again and released a 3.0.0, then archived the repo again(?). We're currently patching dirs-next dependencies back to dirs in debian because we don't know which one to track.

PR #433, opened 2020-10-22
Add advisory for pair of use-after-frees fixed in libpulse-binding v1.2.1

jnqnfe:


jnqnfe:
Revised to v3 format.

Shnatsel:
Thanks for filing it here! I see it still fails to parse though. It's the middle of the night in my timezone so I won't attempt to fix it right now - I'll only break it more.

crates/libpulse-binding/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,19 @@
...
+date = "2018-06-15"
+url = "https://github.com/jnqnfe/pulse-binding-rust/security/advisories/GHSA-ghpq-vjxw-ch5w"
+categories = ["memory-corruption"]
+
+[versions]
+patched = [">= 1.2.1"]
+
+[affected]
+functions = { "libpulse-binding::stream::Stream::get_format_info" = ["< 1.2.1"],
+              "libpulse-binding::stream::Stream::get_context" = ["< 1.2.1"] }
```
tarcieri, line 14: Try changing this to:

```suggestion
[affected.functions]
"libpulse-binding::stream::Stream::get_format_info" = ["< 1.2.1"]
"libpulse-binding::stream::Stream::get_context" = ["< 1.2.1"]
```

jnqnfe: Tried that, didn't work. Had a thought though - using underscores for the crate name instead of dashes, and that's fixed it! :)

PR #434, opened 2020-10-22
Add advisory for possible use-after-free fixed in libpulse-binding v2.5.0

jnqnfe:


jnqnfe:
Revised to v3 format.

Shnatsel:
Thanks!

PR #435, opened 2020-10-22
Add advisory for notable UB fix in libpulse-binding v2.6.0

jnqnfe:
I wasn't sure how to differentiate an UB entry from a security entry... The reporting guidelines clearly say that unsoundness/UB issues can be reported regardless of whether or not they are vulnerabilities. I think this one counts as a noteworthy issue.

### DIFF
--- a/crates/libpulse-binding/RUSTSEC-0000-0000.md
+++ b/crates/libpulse-binding/RUSTSEC-0000-0000.md
@@ -1,0 +1,19 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "libpulse-binding"
+date = "2019-03-10"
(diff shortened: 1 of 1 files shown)


crates/libpulse-binding/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "libpulse-binding"
+date = "2019-03-10"
+url = "https://github.com/jnqnfe/pulse-binding-rust/commit/7fd282aef7787577c385aed88cb25d004b85f494"
+categories = ["memory-corruption"]
+
```
tarcieri, line 8: > I wasn't sure how to differentiate an UB entry from a security entry

Add the following:

```suggestion
informational = "unsound"
```

jnqnfe: done :)

Issue #454, opened 2020-10-29
Add advisory entries for futures-rs


Qwaz:
I recently reported a Send/Sync bound error in futures-rs MappedMutexGuard (rust-lang/futures-rs#2239) and found few more soundness bugs in the futures-rs issue tracker. I don't have enough time to work on advisories for these bugs at the moment, so I'll leave the list here so that someone else can look into them.

* rust-lang/futures-rs#1654
* rust-lang/futures-rs#2050
* rust-lang/futures-rs#2091
* rust-lang/futures-rs#2206
* rust-lang/futures-rs#2239

BlackHoleFox:
The [PR fixing a `Send + Sync` trait issue on `ArcWake`](https://github.com/rust-lang/futures-rs/pull/1654) has a [comment](https://github.com/rust-lang/futures-rs/pull/1654#issuecomment-526862328) saying it was only present in pre-production releases and I confirmed that no currently available versions of the crate are available on crates.io missing the `Send + Sync` implementation on `ArcWake`. Does an advisory still need to be filed for it? Its been well over a year since.

Shnatsel:
No. If it never made it into a release we have no machine-readable way to specify the version range. And anyone using development snapshots already threw caution to the wind, so we might as well not bother.

Shnatsel:
@BlackHoleFox thanks for the high-quality advisories! Now that they're merged, I'm closing this.

PR #455, opened 2020-10-30
Add advisory for data race fix in futures-util

BlackHoleFox:
Adds an advisory to cover a [recently discovered data race](https://github.com/rust-lang/futures-rs/issues/2239) in `future-util`'s `MappedMutexGuard` API.

1/4 of https://github.com/RustSec/advisory-db/issues/454


alex:
Looks like the linter is not pleased. I think we're using the new markdown based format for new advisories.

alex (approved):

BlackHoleFox:
I saw the failure and took another look at the README to see that I completely missed the fact everything is markdown now. Thanks for the quick response though :)

PR #475, opened 2020-11-09
Unexpected panic in multihash `from_slice` parsing code

kirushik:
This has been fixed for a while now (see the [url from the report](https://github.com/multiformats/rust-multihash/pull/72)), but it only recently occurred to us that it can cause a remotely-triggered DoS in libp2p and other consumers.

So I put today into the `date` field; feel free to correct me, a fixed version of `multihash` has been released on August 29, 2020.

### DIFF
--- a/crates/multihash/RUSTSEC-0000-0000.md
+++ b/crates/multihash/RUSTSEC-0000-0000.md
@@ -2,7 +2,7 @@
 [advisory]
 id = "RUSTSEC-0000-0000"
 package = "multihash"
-date = "2020-11-09"
+date = "2020-11-08"
 url = "https://github.com/multiformats/rust-multihash/pull/72"
 categories = ["denial-of-service"]
 keywords = ["parsing", "panic", "untrusted data"]
--- a/crates/multihash/RUSTSEC-0000-0000.md
+++ b/crates/multihash/RUSTSEC-0000-0000.md
@@ -1,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "multihash"
+date = "2020-11-09"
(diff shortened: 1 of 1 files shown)


crates/multihash/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
...
+id = "RUSTSEC-0000-0000"
+package = "multihash"
+date = "2020-11-09"
+url = "https://github.com/multiformats/rust-multihash/pull/72"
+categories = ["denial-of-service"]
+keywords = ["parsing", "panic", "untrusted data"]
+
+[affected.functions]
+"digests::MultihashRefGeneric<T>::from_slice" = ["< 0.11.3"]
+"digests::MultihashGeneric<T>::from_bytes" = ["< 0.11.3"]
```
tarcieri, line 12: Can you prefix these paths with `multihash::` please? Thanks!

```suggestion
"multihash::digests::MultihashRefGeneric<T>::from_slice" = ["< 0.11.3"]
"multihash::digests::MultihashGeneric<T>::from_bytes" = ["< 0.11.3"]
```

tarcieri:
> So I put today into the date field; feel free to correct me, a fixed version of multihash has been released on August 29, 2020.

The date field is generally the date of discovery or when an initial issue was opened

crates/multihash/RUSTSEC-0000-0000.md
kirushik: Done; my bad, should've spent another second looking at other advisories before creating my own.

kirushik:
OK, I've updated the date to yesterday's, since to my knowledge it was when we realized the repercussions.
Anything else I need to adjust here?

tarcieri:
Looks good!

PR #477, opened 2020-11-09
RUSTSEC-2020-0068: remove parameters from affected functions

tarcieri:
It's breaking `cargo-deny` which hasn't updated to the new parser yet

### DIFF
--- a/crates/multihash/RUSTSEC-2020-0068.md
+++ b/crates/multihash/RUSTSEC-2020-0068.md
@@ -8,8 +8,8 @@ categories = ["denial-of-service"]
 keywords = ["parsing", "panic", "untrusted data"]
 
 [affected.functions]
-"multihash::digests::MultihashRefGeneric<T>::from_slice" = ["< 0.11.3"]
-"multihash::digests::MultihashGeneric<T>::from_bytes" = ["< 0.11.3"]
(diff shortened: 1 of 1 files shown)


roy-work:
Just FYI, I think this also broke `cargo audit`, too,
```
» cargo audit
error: error loading advisory database: parse error: parse error: omit parameters when specifying affected paths: 'MultihashRefGeneric<T>' for key `affected.functions` at line 13 column 1
```
Now, this was versions 0.12.0 & 0.12.1 of `cargo-audit`, which I isn't quite the latest version. I don't know what RustSec's supported versions of `cargo audit` are?

tarcieri:
@roy-work let me know if #477 fixed it

tarcieri:
@roy-work all better now?

roy-work:
Sorry, I guess I wasn't explicit enough. I think this PR fixed it, yes. But your message implied to me that the breakage was limited to `cargo-deny` (I don't think it was), and that once `cargo-deny` updated, that it might be safe to make the same change again. I don't think it would be, but conditional on what versions of `cargo-audit` RustSec supports.

tarcieri:
Older versions of `cargo-audit` will break too yes, since they're using the version of the `rustsec` crate that `cargo-deny` presently uses.

After `cargo-deny` updates we can file an informational advisory to warn users of older versions of `cargo-audit` that they need to upgrade.

PR #479, opened 2020-11-11
Advisories are now in markdown format

amousset:


tarcieri:
Thanks! Missed that one.

PR #482, opened 2020-11-17
Report soundness bug with concurrency in futures-intrusive

ammaraskar:
Upstream issue: https://github.com/Matthias247/futures-intrusive/issues/53

This is a soundness issue with a `Sync` trait that allows data races with the lock guard object. Very similar to one reported in the standard library's [own mutex guard](https://github.com/rust-lang/rust/issues/41622).


tarcieri:
Thanks! Assigned `RUSTSEC-2020-0072`

PR #483, opened 2020-11-18
Report soundness bug with lock_api

ammaraskar:
Upstream issues:

* https://github.com/Amanieu/parking_lot/issues/258
* https://github.com/Amanieu/parking_lot/issues/259

Pretty similar to https://github.com/RustSec/advisory-db/pull/455 where the Mapped Guard variants were not considering the mapped-to types when it came to the concurrency traits.


tarcieri:
Looks good, thanks!

Issue #429, opened 2020-10-15
Support multiple URLs in the URL field


Qwaz:
The current advisory format only allows a single entry in the URL field, but sometimes it is useful to include multiple URLs in advisories.

Examples:
* [RUSTSEC-2019-0034](https://rustsec.org/advisories/RUSTSEC-2019-0034.html) was an advisory for two bugs, hyperium/http#354 and hyperium/http#355. I included the links in the description section and left the URL field empty, because it only supports a single entry.
* [RUSTSEC-2019-0035](https://rustsec.org/advisories/RUSTSEC-2019-0035.html) used a [changelog](https://github.com/rust-random/rand/blob/master/rand_core/CHANGELOG.md#050---2019-06-06) URL that lists PRs and issue IDs patched in that version, but it would be better to include precise issue numbers that are related to the issue in the advisory itself. For instance, rust-random/rand#781 mentioned in the changelog is not directly related to the bug.
* [RUSTSEC-2019-0009](https://rustsec.org/advisories/RUSTSEC-2019-0009.html) put servo/rust-smallvec#148 in the URL section, while servo/rust-smallvec#149 is also relevant.

Is it too late to introduce this kind of breaking changes to the advisory format, or can we still do this as part of V3 migration (#414)?

tarcieri:
I would suggest adding a brand new `references` section in addition to the current `url` field.

The `references` section can be a list/array of URLs, but `url` still provides the most definitive information about a vulnerability.

Qwaz:
I like the name `references`. One question though, what would be the recommendation if there are multiple equally definitive URLs (as in RUSTSEC-2019-0009)? Put one of them in `url` and the other in `references`, leave `url` empty and put both of them in `references`, or something else?

tarcieri:
I think it's up to the advisory author to decide. If there's no definitive one, they could all just be `references`.

tarcieri:
It seems `references` is already used for a list of vulnerability IDs, and there's some usage of it in the current advisory DB, unfortunately.

BlackHoleFox:
Its not as fun as references, but maybe “sources”, as in “information sources”?

tarcieri:
@BlackHoleFox I had a similar thought, but `resources`

tarcieri:
Reflecting on this, I'd *really* like to use `references` for this, but that'd be a breaking change.

One possible way to do it:

- Rename the current `references` section to `related`. This is non-breaking.
[...]

PR #492, opened 2020-11-23
Rename `references` fields to `related`

tarcieri:
This frees up `references` to be used for tracking multiple URLs with additional information.

See also: RustSec/advisory-db#429


tarcieri:
Will follow up fixing the linter shortly, but I'd like to do this rename first

PR #497, opened 2020-11-29
Add advisory for branca

brycx:
A couple of days ago, I informed the maintainer of this crate  about this security issue. A fix has now been released.


tarcieri:
Though we sometimes frown on simple DoS vulns due to unexpected panics, given this is a panic processing a security credential which I assume can be triggered by an otherwise unauthenticated network attacker in many scenarios I think this is fine to publish an advisory for.

tarcieri:
Assigned `RUSTSEC-2020-0075` in #498

PR #503, opened 2020-12-02
Add advisory on net2 making invalid memory assumptions

faern:
Hello! This "vulnerability" is not exploitable right now. But it will cause invalid memory access when the standard library changes how `SocketAddr` is implemented. See https://github.com/rust-lang/rust/pull/78802 for details.

I'm aiming to follow this advisory up with very similar ones on `socket2`, `miow` and `mio` since they contain more or less the exact same problem. But I will start with one until we have agreed on the content and format since they will likely be very similar. I can add all the others to the same PR if you prefer that.

I was not sure what `categories` to apply to this, if any. Since it's a little bit unclear how this could be exploitable in the future. It more or less depends on what the standard library changes the memory representation into. But since it was optional I left it out.

### DIFF
--- a/crates/net2/RUSTSEC-0000-0000.md
+++ b/crates/net2/RUSTSEC-0000-0000.md
@@ -1,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "net2"
+date = "2020-11-07"
+url = "https://github.com/deprecrated/net2-rs/issues/105"
+keywords = ["memory", "layout", "cast"]
+
+[versions]
+patched = [">= 0.2.36"]
+```
+
+# `net2` invalidly assumes the memory layout of std::net::SocketAddr
+
+The [`net2`](https://crates.io/crates/net2) crate has converted `std::net::SocketAddr`
+instances into C `sockaddr` pointers simply by casting the pointer. This will cause
+invalid memory access if/when the standard library ever changes the implementation.
+No warnings or errors will be emitted once the change happens.
+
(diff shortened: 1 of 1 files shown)


faern:
I reworded some stuff. I can squash commits later after I get some feedback.

Do you think it's worth/reasonable to mention that changing the memory layout is not just hypothetically possible, but actually in the pipe to hopefully happen. To make a stronger statement about the importance of upgrading.

Shnatsel:
For issues that don't pose an immediate threat we usually use `informational = "unsound"`, I believe this would be a good fit here.

faern:
Thank you @Shnatsel. Yeah unsoundness is exactly what it is so that sounds good.

Shnatsel:
We have [a separate advisory for the deprecation](https://rustsec.org/advisories/RUSTSEC-2020-0016.html), so I'm not sure it makes sense to include the notice to migrate to this particular advisory. Other than that looks good. Thanks!

faern:
Ok. It just felt strange to say "Please upgrade" when in reality people should move away from the crate. But I removed that entire section. The advisory metadata already tells the user which version is safe if they really have to stay with `net2`

Shnatsel:
Thanks again!

faern:
@Shnatsel Thank you for the quick response! When I add the similar advisories for the other three crates, do you prefer them as a single PR or one per crate?

Shnatsel:
I'm not sure how CI will handle a PR for multiple crates, so let's do one per crate

PR #509, opened 2020-12-02
Add advisory on mio SocketAddr casting

faern:
Same as #507 

I realized one thing: When `patched` is filled in this way, it will only warn people with `0.7.x` versions in their lockfiles where `x<6`, right? It won't start alarming users of `mio 0.6` or other minor releases?

### DIFF
--- a/crates/mio/RUSTSEC-0000-0000.md
+++ b/crates/mio/RUSTSEC-0000-0000.md
@@ -9,6 +9,7 @@ informational = "unsound"
 
 [versions]
 patched = [">= 0.7.6"]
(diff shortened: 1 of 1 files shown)


faern:
This is the last advisory I have on `SocketAddr`. Thanks for the amazing response time so far @Shnatsel!

Shnatsel:
This way 0.6.x users will see warnings too. You need to specify `unaffected` field to exclude them.

faern:
Aaaah. My bad then. Fix coming up!

faern:
I probably forgot that since the first one, `net2`, was affected on **all** previous releases.

PR #523, opened 2020-12-08
Add advisory for data race in thex

ammaraskar:
Crate: https://crates.io/crates/thex

No upstream issue for this since the repository for this crate seems to have vanished. It used to be at https://github.com/krl/thex but it seems to have been deleted and there hasn't been a release for a long time. (Pinging @krl, not sure if this repo has moved but just fyi you can [yank](https://doc.rust-lang.org/cargo/commands/cargo-yank.html) old unmaintained crates)

Here's a quick little proof-of-concept that segfaults from safe rust code:
```rust
#![forbid(unsafe_code)]

use thex::Thex;
use std::rc::Rc;

fn main() {
    let rc = Rc::new(());
    let rc_clone = rc.clone();

    let thex = Thex::new(rc_clone);
    std::thread::spawn(move || {
        let smuggled_rc = thex.shared();

        println!("Thread: {:p}", *smuggled_rc);
        // Race the refcount with the main thread.
        for _ in 0..100_000_000 {
            smuggled_rc.clone();
        }
    });

    println!("Main:   {:p}", rc);
    for _ in 0..100_000_000 {
        rc.clone();
    }
}
```

Output:
```
Main:   0x559653d9cb50
Thread: 0x559653d9cb50

Terminated with signal 4 (SIGILL)
```

### DIFF
--- a/crates/thex/RUSTSEC-0000-0000.md
+++ b/crates/thex/RUSTSEC-0000-0000.md
@@ -1,0 +1,19 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "thex"
+date = "2020-12-08"
+categories = ["memory-corruption"]
+keywords = ["concurrency"]
+
+[versions]
+patched = []
+```
+
+# Thex<T> allows data races of non-Send types across threads
+
+`thex::Thex<T>` implements `Sync` for all types `T`. However, it is missing a
+bound for `T: Send`.
+
+This allows non-Send types such as `Rc` to be sent across thread boundaries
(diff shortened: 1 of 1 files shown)


tarcieri:
Thanks! Assigned `RUSTSEC-2020-0090` in #527.

It might be worth filing an unmaintained crate advisory as well.

ammaraskar:
Aah, looks like it just got yanked: https://crates.io/crates/thex

Should an unmaintained advisory still be filed?

tarcieri:
Yanking all releases is probably sufficient, especially as `cargo audit` warns for yanked crates.

PR #525, opened 2020-12-09
Add nanorand 0.5.0 RNG weakness

aspenluxxxy:


crates/nanorand/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,22 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "nanorand"
+date = "2020-12-09"
+url = "https://twitter.com/aspenluxxxy/status/1336684692284772352"
+categories = ["crypto-failure"]
+cvss = "CVSS:3.1/AV:L/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:N/E:U/RL:O"
```
tarcieri, line 8: Currently we only support the [CVSS 3.1 Base Metric Group](https://www.first.org/cvss/v3.1/specification-document).

Can you remove the environmental (`E`) component?

tarcieri: `RL` as well, but I think that should do it.

tarcieri:
Thanks! Assigned `RUSTSEC-2020-0089` in #526

PR #528, opened 2020-12-10
Add advisory for data race in aovec

ammaraskar:
Crate: https://crates.io/crates/aovec

Much like #523, the repo for this crate seems to 404 now so there's no upstream issue here. The vector provided by this crate implements `Send` and `Sync` unconditionally which allows for data races.

Here's a simple demonstration that segfaults safe Rust code using this crate:
```rust
#![forbid(unsafe_code)]

use aovec::Aovec;
use std::rc::Rc;

fn main() {
    let vec = Aovec::new(1);

    let rc = Rc::new(42);
    vec.push(rc.clone());

    std::thread::spawn(move || {
        println!("Thread: {:p}", vec[0]);
        for _ in 0..100_000_000 {
            vec[0].clone();
        }
    });

    println!("Main:   {:p}", rc);
    for _ in 0..100_000_000 {
        rc.clone();
    }
}
```

Output:
```
Main:   0x5639ef979b50
Thread: 0x5639ef979b50

Terminated with signal 4 (SIGILL)
```

### DIFF
--- a/crates/aovec/RUSTSEC-0000-0000.md
+++ b/crates/aovec/RUSTSEC-0000-0000.md
@@ -1,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "aovec"
+date = "2020-12-10"
+categories = ["memory-corruption"]
+keywords = ["concurrency"]
+
+[versions]
+patched = []
+```
+
+# Aovec<T> lacks bound on its Send and Sync traits allowing data races
+
+`aovec::Aovec<T>` is a vector type that implements `Send` and `Sync` for all
+types `T`.
+
+This allows non-Send types such as `Rc` and non-Sync types such as `Cell` to
+be used across thread boundaries which can trigger undefined behavior and
+memory corruption.


tarcieri:
I think we might consider filing this one as `informational = "unsound"`

Shnatsel:
@tarcieri IIRC we have sometimes treated this as a vulnerability in the past. Since it can lead to real data races and segfaults in real-world code, I'd prefer to surface this as an error. Especially since the fix is straightforward and can be released in a semver-compatible update. I believe it's best to reserve `informational = "unsound"` for issues that do not currently pose a memory safety threat in real-world code, e.g. #509 or #543

I understand the line between unsoundness and a vulnerability is not currently well defined, but I want to use this class of issues to establish some precedent.

tarcieri:
As it were, #533 was another data race

ammaraskar:
Hmm, I think for these sort of data races, it might make sense to have them not be `informational` if it's reasonably expected that a user of the library would write code that could cause the data race. For example in this case since this is a generic vector class, it's fairly reasonable and probably an easy-to-make mistake to create an instance with a non-Send/Sync type.

If it seems like the code is a bit too artificial and a reasonable user of the library wouldn't make a mistake like it then maybe `informational` would be better? As such I think #533 should also be marked non-`informational`.

PR #533, opened 2020-12-14
add advisory for data race in 'reffers'

JOE1994:
### reffers: Unsound: can make `ARefss` contain a !Send, !Sync object.

Original issue report: https://github.com/diwic/reffers-rs/issues/7

Commit that fixed the issue: https://github.com/diwic/reffers-rs/commit/6dd7ca0d50f2464df708975cdafcfaeeb6d41c66

Thank you for reviewing this PR :+1:


tarcieri:
Unfortunate there isn't a patched release.

Anyway, thanks for the advisory!

PR #536, opened 2020-12-18
File an http request smuggling advisory for async-h1 < 2.3.0

jbr:
Hi there and thanks for running this database! Here's my first attempt at filing an advisory, please let me know if I should make any changes. I'm not certain on a few fields of the CVSS so I didn't fill that in at all.


Shnatsel:
Looks good to me! I'll drop the commented out fields, and it will be good to go.

PR #543, opened 2021-01-01
cache: exposes internally used raw pointer

stbuehler:


Shnatsel:
Thank you for the report!

I am struggling to imagine how this could lead to a segfault in real-world code, unless a user of this crate is deliberately writing underhanded code. Can you think of any? If not, I'd prefer to add `informational = "unsound"` field to surface it as a warning instead of a hard error.

tik-stbuehler:
Yes, it seems this needs a user deliberately abusing it; otoh some people might focus on reviewing only crates using `unsafe` - and trust that other crates can't do "bad" things, so I disagree with the conclusion that this isn't a real problem.

Also I don't think the current quality of the crate deserves anything but a "don't ever use this" label.

If you insist on the `informational` label I'll probably have to open another one for https://github.com/krl/cache/issues/1, which I hope really qualifies for hard error :)

(Also I don't think this crate deserves much attention and wouldn't have bothered if it didn't use such "global" name.)

PR #545, opened 2021-01-04
Add advisory for interfaces2

dalance:
interfaces2 is not maintained now, and I took over the maintenance of the original interfaces crate.

https://github.com/aep/interfaces-rs


Shnatsel:
Thanks!

Issue #539, opened 2020-12-21
Add advisories for standard library soundness bugs


Qwaz:
We recently re-discovered rust-lang/rust#60977 and rust-lang/rust#78498 in [abi_stable crate](https://github.com/rodrimati1992/abi_stable_crates/issues/44). While investigating, we found a few more interesting bugs in the Rust issue tracker. I believe these bugs need advisories; I'll leave the list here so that someone else can look into them and file an advisory.

These are memory safety/soundness bugs in stable APIs.
* rust-lang/rust#79808
* rust-lang/rust#78498

---

These are bugs in unstable APIs, so probably no advisory is needed. (But they are still quite interesting bugs!)
* rust-lang/rust#60977
* rust-lang/rust#76367

alex:
Should be possible to submit them here: https://github.com/RustSec/advisory-db/tree/master/rust/std

Qwaz:
Adding one more to the list :)
https://github.com/rust-lang/rust/issues/80335

Qwaz:
Found one more old bug.
https://github.com/rust-lang/rust/issues/25842

Qwaz:
Heap buffer overflow in `read_to_end_with_reservation()` rust-lang/rust#80894

Qwaz:
Update:

We contacted the Rust security team (security@rust-lang.org) about the process of requesting CVEs for memory safety bugs in the Rust standard library. We got responses that suggest to request CVE numbers on our own for publicly discolsed bugs.

From Steve Klabnik:
> (...)
> 3. Usually, this group's job is to help triage incoming, sensitive
> bugs. Usually that has correlated with CVEs, and so we are the ones
> ending up filing them, but there's an interesting governance question
> here. Furthermore, the CVE system is kind of outside of the project
> generally, that is, they don't actually require that the project is
> the one to request a CVE, though they do ask that reporters talk to
> vendors first, as you're doing here, mostly to minimize duplicates.
>
> I would be interested in hearing what Josh and Pietro have to say
> here. Thanks for raising this; this is new territory for us, and so I
> don't have an immediate good answer.

From Josh Stone:
> (...)
> I agree that issues which are already public probably don't *need* to involve this group. I do think we would like to be aware of all Rust Project CVEs though, so we're not caught off guard.
> 
> If there are *new* security implications realized about an old bug fix, maybe we would still want private disclosure. I can think of a few cases of kernel bugs that were quietly fixed, and then made a huge stink about security later.

[...]

PR #561, opened 2021-01-13
Add advisory for rust-lang/rust#78498

Qwaz:
Advisory for:
String::retain allows safely creating invalid (non-utf8) strings when abusing panic (rust-lang/rust#78498)


Qwaz:
It seems that the linter doesn't understand `alloc::string::String::retain`. Should I just put `std::string::String::retain`?

tarcieri:
Yes. Right now the linter requires the path begins with the same crate as the advisory is filed against.

Alternatively you could file it against `alloc`

Qwaz:
Updated the PR to use `std`, considering that `alloc` became stable in [Rust 1.36.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1360-2019-07-04).

rust/std/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,24 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
```
tarcieri, line 3: Presently all of the vulns in `rust/` are tracked using CVE identifiers. Do you have a CVE ID for this?

Qwaz: No, I don't. I'm not familiar with the CVE request process :sweat_smile:

tarcieri: Perhaps we can ping the [Security Response WG](https://www.rust-lang.org/governance/wgs/wg-security-response) (cc @pietroalbini) and ask them to obtain a CVE for this issue.

Qwaz: #539 contains a list of standard library memory safety/soundness bugs that had observable misbehaviors & shipped to stable Rust.

pietroalbini: Can you send an email to security@rust-lang.org about it?

Qwaz:
I'll reopen the PR when we have CVEs

PR #565, opened 2021-01-17
lazy-init: Missing Send bound for Lazy (khuey/lazy-init#9)

niklasf:
Add an informational advisory for a long-standing soundness issue that has recently been fixed (https://github.com/khuey/lazy-init/commit/ab2cac8383a499be6700244cd2682f55c120f4a7).

### DIFF
--- a/crates/lazy-init/RUSTSEC-0000-0000.md
+++ b/crates/lazy-init/RUSTSEC-0000-0000.md
@@ -1,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "lazy-init"
+date = "2021-01-17"
+informational = "unsound"
+url = "https://github.com/khuey/lazy-init/issues/9"
+
+[versions]
+patched = ["> 0.4.0"]
+```
+
+# Missing Send bound for Lazy
+
+All current versions of this crate allow causing data races in safe code.
+
+The flaw will be fixed in the next release.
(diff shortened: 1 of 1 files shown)


tarcieri:
We haven't fully codified this yet, but we recently decided data race vulnerabilities shouldn't be classed as informational and instead should be treated as a regular security vulnerability /cc @Shnatsel

Shnatsel:
So far we've been categorizing them inconsistently, but I've argued that anything that may result in a memory safety violation and does not require exceedingly contrived code should be treated as a vulnerability, and missing `Send` bound usually qualifies.

@tarcieri what do we need to do to make this policy official?

tarcieri:
I think we can add some clarification here: https://github.com/RustSec/advisory-db/blob/master/CONTRIBUTING.md#criteria

niklasf:
I believe this particular crate would qualify, then, because the primary purpose is to safely provide lazy initialization in multi-threaded code.

Is this the proper categorization, then?

```patch
-informational = "unsound"
```

Shnatsel:
Yes, that sounds about right.

tarcieri (approved):

khuey:
In the future it would be good to give the package maintainer a heads up (so I can do things like release a new version) before pushing an advisory live ...

Shnatsel:
That's fair.

khuey:
It's taken care of now but in the future I'd like to not have to drop what I'm doing to immediately fix busted `cargo deny` checks :)

niklasf:
@khuey Apologies. I should have left a comment on the original issue.

PR #568, opened 2021-01-18
glsl-layout: double drop upon panic in 'fn map_array'

JOE1994:
# glsl-layout: double drop upon panic in 'fn map_array'

link to PR (explanation of bug && fix): https://github.com/rustgd/glsl-layout/pull/10

Thank you for reviewing this PR! :crab:


Shnatsel:
Looks good. Thanks for reporting!

PR #574, opened 2021-01-18
av-data: read from arbitrary address in safe API

JOE1994:
# av-data: `Frame::copy_from_raw_parts` can lead to segfault without `unsafe`

Original issue report: https://github.com/rust-av/rust-av/issues/136

Thank you for reviewing :crab:


Shnatsel:
Thanks!

I'd prefer to have an actionable advisory, i.e. one where there is a patched version to upgrade to. I've asked on the original issue to publish a new release. If this doesn't happen by tomorrow, I'll push this as is.

JOE1994:
The authors have published a new release containing the bug fix, and I'll update this advisory with the patched version number right away.

Shnatsel:
Thanks!

PR #575, opened 2021-01-18
xcb: Soundness issue with base::Error

Qwaz:
Original issue report: https://github.com/rtbo/rust-xcb/issues/93

### DIFF
--- a/crates/xcb/RUSTSEC-0000-0000.md
+++ b/crates/xcb/RUSTSEC-0000-0000.md
@@ -1,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "xcb"
+date = "2020-12-10"
+url = "https://github.com/rtbo/rust-xcb/issues/93"
+categories = ["memory-corruption"]
+informational = "unsound"
+
+[versions]
+patched = []
+```
+
+# Soundness issue with base::Error
+
+`base::Error` type contains public field named `ptr`.
+With this definition, it is possible to create a `base::Error` with an invalid pointer and trigger memory safety errors
+such as use-after-free or double-free with safe Rust.
+
+The users of `xcb` crate are advised not to manipulate the field.


Shnatsel:
Thanks!

psychon:
Sorry for highjacking this PR, but I do not think this warrants a separate issue/PR.
I am exaggerating, but almost all of xcb is unsafe. Here are some examples that are easier to trigger accidentally than what this PR is about:
- interpret some bytes coming from the X11 server as any type you like: https://github.com/rtbo/rust-xcb/issues/95 
- Out-of-bounds read for sending bytes to the server: https://github.com/rtbo/rust-xcb/issues/94
- a "safe" version of `mem::transmute` (with some limits on the types, but not enough): https://github.com/rtbo/rust-xcb/issues/78
- I'm not sure if this even counts, but here is an old issue about a `NULL` pointer dereference in safe rust: https://github.com/rtbo/rust-xcb/issues/64

psychon:
I'm not looking for these on purpose. I am just reading some code and wonder "why doesn't this return a `Result`?". This one is is a call to `std::str::from_utf8_unchecked` with a comment next to it saying "should we check what comes from X?". Yes, you should, and you should not simply assume that the data is fine.
https://github.com/rtbo/rust-xcb/issues/96

(Side note: X11 atoms are specified to use latin1)

Shnatsel:
@psychon thank your for investigating the `xcb` crate and reporting the issues upstream! Could you add a bit more detail to the upstream report - that atoms are specified to use latin1, what exactly is the problem with the current approach, and what should be done instead? I am not sure the issue is actionable for the maintainers as-is.

psychon:
Done, I hope. I also improved my other report there slightly, explaining that `bool` can only have values 0 and 1.

> I am not sure the issue is actionable for the maintainers as-is.

Well, according to its README, I would think that the xcb crate is unmaintained. At its very top:

> Maintainance request
>
> I've been very happy to work on this project, but I don't have the possibility anymore to maintain these bindings to the level the Rust community deserves. I can't spend as much time on it as I used to, and I'm not using neither Rust nor XCB anymore, so I clearly can't improve the bindings with the latest Rust features. Person with motivation and good knowledge of Rust and XCB may contact me per email.

PR #580, opened 2021-01-19
Add advisory for soundness issue in rusb

ammaraskar:
Upstream issue: https://github.com/a1ien/rusb/issues/44

Marked `informational = unsound` because it's unlikely that users of the library would be implementing `UsbContext` themselves.


tarcieri:
> Marked `informational = unsound` because it's unlikely that users of the library would be implementing `UsbContext` themselves.

As an `rusb` user, that sounds reasonable to me.

PR #583, opened 2021-01-19
Add advisory for data race in may_queue

setuid0x0:
Upstream issue: https://github.com/Xudong-Huang/may/issues/88


Shnatsel:
Thanks for the report!

This seems to be easy to trigger and may result in memory corruption, so I'll promote it from `informational="unsound"` to a full-on vulnerabilty.

PR #586, opened 2021-01-19
bra: Read on uninitialized buffer

JOE1994:
# bra: reading on uninitialized buffer can cause UB (`impl<R> BufRead for GreedyAccessReader<R>`)

Original issue report: https://github.com/Enet4/bra-rs/issues/1

Thank you for reviewing!


Shnatsel:
Thanks!

PR #588, opened 2021-01-19
unconditional Sync impl of `AtomicOption<T>` allows UB

JOE1994:
# atomic-option: AtomicOption should have Send + Sync bound on its type argument.

Original issue report: https://github.com/reem/rust-atomic-option/issues/4

The crate is popular on crates.io. (>=106,332 downloads).
The author hasn't responded to the issue report which was submitted more than 10 weeks ago.

Thank you for reviewing :crab:

### DIFF
--- a/crates/atomic-option/RUSTSEC-0000-0000.md
+++ b/crates/atomic-option/RUSTSEC-0000-0000.md
@@ -1,0 +1,19 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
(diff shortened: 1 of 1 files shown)


Shnatsel:
I want to write a nice message on the issue tracker letting the author know about the advisory, but my brain is not cooperating today, and this probably will not make much of a difference in this particular case. Latest commit was in 2016, so I'll go ahead and merge this without a patched version.

PR #641, opened 2021-01-21
ID assignment action: guard against race conditions 

Shnatsel:
Guard against race conditions resulting in duplicate ID assignment by writing a hash of the directory structure (not file contents) into a file. If two ID assignment jobs run concurrently, the file will create a merge conflict.

This is completely untested, since I don't know how to trigger a test run of a Github action.


alex (approved):

alex:
You might need to manually create this file first? Not sure if the action will add the new file before pushing.

alex (approved):

PR #642, opened 2021-01-21
va-ts: `Demuxer` can carry non-Send types across thread boundaries

JOE1994:
Original issue report: https://github.com/video-audio/va-ts/issues/4


Shnatsel:
Thanks!

PR #646, opened 2021-01-22
[patched] Add advisory for data race in ruspiro-singleton

ammaraskar:
Upstream issue: https://github.com/RusPiRo/ruspiro-singleton/issues/10


Shnatsel:
Thanks!

PR #614, opened 2021-01-20
raw-cpuid: Multiple soundness issues

niklasf:
Draft. To be updated soon with solutions.

PR #671, opened 2021-01-24
raw-cpuid: Optional `Deserialize` implementations unsound

niklasf:
Following up on #614: Thanks. Here's part 2, then.


khuey:
The merged advisory is incorrect (or rather, it was correct when the PR was filed but has been overtaken by events in the previous seven months). This was fixed in version 9.1.1 of raw-cpuid.

niklasf:
Sorry about not staying on top of this. Sent a pull request to update the advisory: #1002.

PR #675, opened 2021-01-24
[patched] Add advisory for data race in tiny_future

ammaraskar:
Upstream issue: https://github.com/KizzyCode/tiny_future/issues/1


Shnatsel:
Thank you!

PR #679, opened 2021-01-25
libsbc: Decoder<R> can carry `R: !Send` to other threads

JOE1994:
Original issue report: https://github.com/mvertescher/libsbc-rs/issues/4


Shnatsel:
Could you add a sentence or two on why this issue is minor? It would help downstream consumers evaluate the impact.

Shnatsel:
Thanks!

Issue #653, opened 2021-01-23
xcb is unsound and unmaintained


Shnatsel:
Following up from https://github.com/RustSec/advisory-db/pull/575#issuecomment-765878975, the following issues are currently outstanding in the `xcb` crate:

 * interpret some bytes coming from the X11 server as any type you like: [rtbo/rust-xcb#95](https://github.com/rtbo/rust-xcb/issues/95)
 * Out-of-bounds read for sending bytes to the server: [rtbo/rust-xcb#94](https://github.com/rtbo/rust-xcb/issues/94)
* a "safe" version of `mem::transmute` (with some limits on the types, but not enough): [rtbo/rust-xcb#78](https://github.com/rtbo/rust-xcb/issues/78)

The README states:

>    Maintainance request
>
> I've been very happy to work on this project, but I don't have the possibility anymore to maintain these bindings to the level the Rust community deserves. I can't spend as much time on it as I used to, and I'm not using neither Rust nor XCB anymore, so I clearly can't improve the bindings with the latest Rust features. Person with motivation and good knowledge of Rust and XCB may contact me per email.

Shnatsel:
I've emailed the author as requested in the readme, flagging these issues and offering to help with finding a new maintainer.

Shnatsel:
I have not received a reply in 12 days, so let's proceed with advisories for the known soundness bugs as well as unmaintained status.

@psychon would you do the honors and open the advisory PR, like #575?

Shnatsel:
`x11rb` crate looks like a maintained alternative. We don't carry an "unmaintained" advisory yet, just reports of soundness issues.

PR #750, opened 2021-02-04
Report various rust-xcb issues to RustSec

psychon:
Closes: https://github.com/RustSec/advisory-db/issues/653

---

As requested by @Shnatsel, here is my attempt at an advisory. I tried to follow what #575 does, but elaborate all the various issues in sub-headings. I did not include `informational = "unsound"` (#575 has this), because this is also `informational = "unmaintained"` and this `informational` entry seems to be undocumented (I only found its use in various advisories), so I am not sure what to do. My best idea: If wanted/necessary, I can split this into two PRs, one for the maintenance status and one for the other issues.

---

I'm also nervous about "there is a code generator that turns XML into lots and lots of `unsafe` code, most of that in functions that are supposed to be safe", but that's likely not worth an advisory. At least this makes it a lot harder to review the code in this crate since I first have to find the actual code in `target/`. Just saying.

(I suppose `cargo geiger` would warn users about this, at least)

---

I take this chance to say "Thank you!" to @Shnatsel. I saw your nick name already a couple of times and my first association to "Shnatsel" is "lots of work for Rust security", like fuzzing, advisory, code review infrastructure etc. If someone else had pinged me and asked for a PR, I am not sure what I would have done, but this is my chance to help you helping Rust. Thanks for all your work.

### DIFF
--- a/crates/xcb/RUSTSEC-0000-0000.md
+++ b/crates/xcb/RUSTSEC-0000-0000.md
@@ -1,0 +1,70 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "xcb"
+date = "2021-02-04"
+categories = ["memory-corruption", "memory-exposure"]
+
+[versions]
+patched = []
+```
+
+# Multiple soundness issues and unmaintained
+
+## Calls `std::str::from_utf8_unchecked()` without any checks
+
+The function `xcb::xproto::GetAtomNameReply::name()` calls
+`std::str::from_utf8_unchecked()` on the raw bytes that were received from the
+X11 server without any validity checks. The X11 server only prevents interior
+null bytes, but otherwise allows any X11 client to create an atom for arbitrary
+bytes.
+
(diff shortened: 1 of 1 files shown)


Shnatsel:
**TL;DR:** I'm going to drop the "unmaintained" part from this advisory and merge it.

I believe it would be best to make a separate advisory for the maintenance status. That's where `informational = "unmaintained"` would be appropriate. See https://rustsec.org/advisories/RUSTSEC-2016-0005.html or https://rustsec.org/advisories/RUSTSEC-2017-0007.html for examples of how those are usually handled.

Thanks for pointing out the lack of docs, I've opened #751 about this!

And thank you for the kind words! In case you're wondering, here's a sneak peek at what I'm currently working on:

 * https://github.com/rust-secure-code/cargo-supply-chain
 * https://github.com/algesten/ureq/issues/312

Shnatsel:
Done. Thanks!

PR #759, opened 2021-02-08
Add link to site

djc:
When I was l looking for the rustsec.org recently with keywords like "rust advisory database", I found the repository but had some trouble finding the site (perhaps to my defaulting to use DuckDuckGo). A link here makes sense, I think?


tarcieri:
Sure

PR #764, opened 2021-02-14
rand_core: incorrect check on buffer length when seeding RNGs

dhardy:
**Summary:** `rand_core::le::read_u32_into` and `read_u64_into` have incorrect checks on the source buffer length, allowing the destination buffer to be under-filled.

**Implications:** some downstream RNGs, including `Hc128Rng` (but not the more widely used `ChaCha*Rng`), allow seeding using the `SeedableRng::from_seed` trait-function with too short keys.

**Link:** https://github.com/rust-random/rand/pull/1096

---

I suspect I am being over-cautious by reporting this and it can be closed, but as you say, if in doubt...
Affected versions of `rand_core` have not (yet) been yanked.


Shnatsel:
I agree not giving enough entropy to an RNG that's supposed to be cryptographically secure deserves an advisory.

The advisory format has transitioned to markdown. See e.g. https://github.com/RustSec/advisory-db/blob/master/crates/rgb/RUSTSEC-2020-0029.md for the new template. The file should also have the `.md` extension.

I've looked at our README again and that's not made sufficiently clear, so I'll file an issue about that.

dhardy:
> I agree not giving enough entropy to an RNG that's supposed to be cryptographically secure deserves an advisory.

That isn't the question. This issue doesn't make correct use of the RNGs in libraries unsafe (nor does the fix make incorrect usage such as using a seed from a weak source impossible), it's just one extra check. (Note the [`CryptoRng`](https://docs.rs/rand_core/0.6.2/rand_core/trait.CryptoRng.html) docs.)

This advisory *is* Markdown format... without an extension.

crates/rand_core/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,16 @@
+[advisory]
```
tarcieri, line 1: Add a ```toml front matter delimiter here:

```suggestion
```toml
[advisory]
```

crates/rand_core/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,16 @@
...
+keywords = []
+
+# Optional: metadata which narrows the scope of what this advisory affects
+[affected]
+functions = { "rand_core::le::read_u32_into" = ["< 0.6.2, >= 0.6.0"] }
+functions = { "rand_core::le::read_u64_into" = ["< 0.6.2, >= 0.6.0"] }
+
+[versions]
+patched = [">= 0.6.2"]
+unaffected = ["< 0.6.0"]
```
tarcieri, line 16: Close the TOML front matter delimiter here and add a title and description:

```suggestion
unaffected = ["< 0.6.0"]
```

# Incorrect check on buffer length when seeding RNGs

Summary: rand_core::le::read_u32_into and read_u64_into have incorrect checks on the source buffer length, allowing the destination buffer to be under-filled.

Implications: some downstream RNGs, including Hc128Rng (but not the more widely used ChaCha*Rng), allow seeding using the SeedableRng::from_seed trait-function with too short keys.
```

File: crates/rand_core/RUSTSEC-0000-0000.md [hunk 3]
```diff
@@ -0,0 +1,24 @@
...
+package = "rand_core"
+date = "2021-02-12"
+url = "https://github.com/rust-random/rand/pull/1096"
+categories = ["crypto-failure"]
+keywords = []
+
+# Optional: metadata which narrows the scope of what this advisory affects
+[affected]
+functions = { "rand_core::le::read_u32_into" = ["< 0.6.2, >= 0.6.0"] }
+functions = { "rand_core::le::read_u64_into" = ["< 0.6.2, >= 0.6.0"] }
```
tarcieri, line 13: Try this:

```
[affected.functions]
"rand_core::le::read_u32_into" = ["< 0.6.2, >= 0.6.0"]
"rand_core::le::read_u64_into" = ["< 0.6.2, >= 0.6.0"]
```

PR #765, opened 2021-02-14
Add advisory on nb-connect SocketAddr casting

faern:
The equivalent of #503, #505, #509 and #507. But for the `nb-connect` crate. Before the 1.0.3 release `nb-connect` invalidly assumed the memory layout of a standard library type. Something that is not part of the stable API of `std`.

https://github.com/smol-rs/nb-connect/issues/1


Shnatsel:
Thanks!

Issue #751, opened 2021-02-04
`informational` field is undocumented


Shnatsel:
We do not have the `informational` field in the advisory template, which leads to confusion for reporters (see e.g. #750).

tarcieri:
It's documented in `rustdoc` (which, in general, probably provides the most detailed documentation for the entire advisory format):

https://docs.rs/rustsec/0.23.0/rustsec/advisory/informational/enum.Informational.html

However it'd be good to link this in the comments on the advisory template, along with a short synopsis.

pinkforest:
https://github.com/rustsec/advisory-db/pull/1341 clarified on this in the example

This was also discussed in zulip that led to that PR - closing this as Resolved

PR #769, opened 2021-02-15
Add use-after-free advisory for `yottadb 1.1.0` and earlier

jyn514:
I've also yanked all affected versions.

### DIFF
--- a/crates/yottadb/RUSTSEC-0000-0000.md
+++ b/crates/yottadb/RUSTSEC-0000-0000.md
@@ -1,0 +1,41 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "yottadb"
+date = "2021-02-09"
+url = "https://gitlab.com/YottaDB/Lang/YDBRust/-/issues/40"
+categories = ["memory-corruption"]
+keywords = ["use-after-free"]
+
+[versions]
+patched = [">= 1.2.0"]
+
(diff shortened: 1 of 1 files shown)


Shnatsel:
The advisory looks very detailed! Thanks for the report!

crates/yottadb/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,42 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "yottadb"
+date = "2021-02-09"
+url = "https://gitlab.com/YottaDB/Lang/YDBRust/-/issues/40"
+categories = ["memory-corruption"]
+keywords = ["use-after-free"]
+informational = "unsound"
```
Shnatsel, line 9: `informational = "unsound"` is used when the issue cannot occur in typical usage of the library. (We still need proper docs for it, which is #751). The provided code example looks fairly trivial to me, so if that causes memory corruption, I'd prefer to drop this field and file it as a regular advisory (i.e. vulnerability).

What do you think?

jyn514: Hmm, ok, seems reasonable. I agree this is pretty trivial, I'll drop the field.

Shnatsel:
Thanks again!

PR #774, opened 2021-02-21
Add advisory on comrak XSS

mrtc0:
Add an advisory for comrak XSS. https://github.com/kivikakk/comrak/releases/tag/0.9.1


Shnatsel:
Thanks!

PR #780, opened 2021-03-01
Add an advisory on lifetime extension in generic-array

xfix:
This is an advisory for an old version of `generic-array`, however some crates (notably `headers`) still use those old versions of `generic-array`.


Shnatsel:
Thank you!

PR #784, opened 2021-03-01
[patched] Add advisory for memory safety issue in toodee's insert_row

ammaraskar:
Upstream issue: https://github.com/antonmarsden/toodee/issues/13


Shnatsel:
Thanks!

PR #788, opened 2021-03-02
Mark RUSTSEC-2020-0146 as unsound

xfix:
I didn't realize this was an option because it wasn't in example advisory.

### DIFF
--- a/crates/generic-array/RUSTSEC-2020-0146.md
+++ b/crates/generic-array/RUSTSEC-2020-0146.md
@@ -5,6 +5,7 @@ package = "generic-array"
 date = "2020-04-09"
 informational = "unsound"
 url = "https://github.com/fizyk20/generic-array/issues/98"
+categories = ["memory-corruption"]
 
 [versions]
 patched = [">= 0.14.0"]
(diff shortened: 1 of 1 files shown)


crates/generic-array/RUSTSEC-2020-0146.md
```diff
@@ -3,9 +3,8 @@
 id = "RUSTSEC-2020-0146"
 package = "generic-array"
 date = "2020-04-09"
+informational = "unsound"
 url = "https://github.com/fizyk20/generic-array/issues/98"
-categories = ["memory-corruption"]
```
tarcieri, line 7: I'd suggest leaving this one

Shnatsel:
`informational = "unsound"` downgrades this from an error to a warning. Is that intentional? Since this allows use-after-free, I'd suggest leaving it as a hard error.

tarcieri:
@Shnatsel hmm, maybe you're right. Seems this is a bigger problem than I thought.

Shnatsel:
@xfix could you elaborate on the reasoning for marking it informational? (We do need to document that field; that's #751)

Shnatsel:
I meant to open a pull request reverting this for discussion and instead accidentally committed to master :facepalm:

xfix:
I had no idea about policies, if it shouldn't be informational, that's fine with me.

Shnatsel:
Yeah, we really should document that better. Thanks!

PR #792, opened 2021-03-02
Revert "Mark RUSTSEC-2020-0146 as unsound (#788)"

Shnatsel:
Allows use-after-free, and looks easy enough to trigger.


tarcieri (approved):

Shnatsel:
Got a reply from the author of the original PR, merging.

PR #793, opened 2021-03-02
[patched] Add advisory for double-free in scratchpad

ammaraskar:
Upstream issue: https://github.com/okready/scratchpad/issues/1


Shnatsel:
Thank you!

PR #795, opened 2021-03-02
[patched] Add advisory for aliasing violation in nano_arena

ammaraskar:
Upstream issue: https://github.com/bennetthardwick/nano-arena/issues/1

### DIFF
--- a/crates/nano_arena/RUSTSEC-0000-0000.md
+++ b/crates/nano_arena/RUSTSEC-0000-0000.md
@@ -1,0 +1,30 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "nano_arena"
+date = "2021-01-31"
+url = "https://github.com/bennetthardwick/nano-arena/issues/1"
+categories = ["memory-corruption"]
+keywords = ["memory-safety", "aliasing", "unsound"]
+
+[versions]
+patched = [">= 0.5.2"]
+
+[affected.functions]
+"nano_arena::Arena::split_at" = ["< 0.5.2"]
+"nano_arena::ArenaSplit::split_at" = ["< 0.5.2"]
+```
+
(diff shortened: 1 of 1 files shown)


crates/nano_arena/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,30 @@
...
+
+[versions]
+patched = [">= 0.5.2"]
+
+[affected.functions]
+"nano_arena::Arena::split_at" = ["< 0.5.2"]
+"nano_arena::ArenaSplit::split_at" = ["< 0.5.2"]
+```
+
+# split_at allows violation of Rust's reference aliasing rules
```
Shnatsel, line 18: "Aliasing rules" is a bit ambiguous - it may refer both to obtaining a `&mut` reference while another reference exists and to the currently-experimental Stacked Borrows model. Something like "allows multiple mutable references to the same data" would be more descriptive.

crates/nano_arena/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,30 @@
...
+
+Affected versions of this crate assumed that `Borrow<Idx>` was garunteed to
```
Shnatsel, line 20:
```suggestion
Affected versions of this crate assumed that `Borrow<Idx>` was guaranteed to
```

Shnatsel (changes requested):
I've flagged a couple of nits, but looks good otherwise. Thanks!

crates/nano_arena/RUSTSEC-0000-0000.md
ammaraskar: Good point, thanks for the suggestion. I've changed this to be more explicit about multiple mutable references.

Shnatsel (approved):

PR #799, opened 2021-03-03
[patched] Add advisory for double-free in stack_dst

ammaraskar:
Upstream issue: https://github.com/thepowersgang/stack_dst-rs/issues/5


Shnatsel:
Thanks!

PR #809, opened 2021-03-04
Report use-after-free issue in diesels sqlite backend

weiznich:
We've discovered a use-after-free issue in diesel's sqlite backend, which can be triggered by real world code. See the corresponding PR (https://github.com/diesel-rs/diesel/pull/2663) for details.

### DIFF
--- a/crates/diesel/RUSTSEC-0000-0000.md
+++ b/crates/diesel/RUSTSEC-0000-0000.md
@@ -15,7 +15,7 @@ patched = [">= 1.4.6"]
 
 # Fix a use-after-free bug in diesels Sqlite backend
 
-We've missused `sqlite3_column_name`. The
+We've misused `sqlite3_column_name`. The
 [SQLite](https://www.sqlite.org/c3ref/column_name.html) documentation
 states that the following:
 
--- a/crates/diesel/RUSTSEC-0000-0000.md
+++ b/crates/diesel/RUSTSEC-0000-0000.md
@@ -1,0 +1,31 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "diesel"
+date = "2021-03-05"
+url = "https://github.com/diesel-rs/diesel/pull/2663"
+categories = ["memory-exposure"]
+keywords = ["use after free"]
+
+[affected]
+functions = { "diesel::SqliteConnection::query_by_name" = ["< 1.4.6"] }
+[versions]
+patched = [">= 1.4.6"]
+```
+
+# Fix a use-after-free bug in diesels Sqlite backend
+
+We've missused `sqlite3_column_name`. The
+[SQLite](https://www.sqlite.org/c3ref/column_name.html) documentation
+states that the following:
+
+> The returned string pointer is valid until either the prepared statement
+> is destroyed by sqlite3_finalize() or until the statement is automatically
+> reprepared by the first call to sqlite3_step() for a particular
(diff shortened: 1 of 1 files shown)


crates/diesel/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "diesel"
+date = "2021-03-05"
+url = "https://github.com/diesel-rs/diesel/pull/2663"
+categories = ["memory-exposure"]
```
Shnatsel, line 7:
```suggestion
categories = ["memory-corruption"]
```
Memory exposure is meant for disclosing memory contents, e.g. Heartbleed.

crates/diesel/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
...
+
+[affected]
+functions = { "diesel::SqliteConnection::query_by_name" = ["< 1.4.6"] }
+[versions]
+patched = [">= 1.4.6"]
+```
+
+# Fix a use-after-free bug in diesels Sqlite backend
+
+We've missused `sqlite3_column_name`. The
```
Shnatsel, line 18:
```suggestion
We've misused `sqlite3_column_name`. The
```

Shnatsel (commented):
Thanks for the report!

I understand the fixed version is not yet live? We'd prefer to issue an advisory once a patched version is available on crates.io.

crates/diesel/RUSTSEC-0000-0000.md
Ten0: This particular use-after-free builds an `&str` of longer lifetime than it should. Can this really lead to more than memory exposure?

Shnatsel: This can violate the invariant of `&str` being UTF-8 and cause arbitrary malfunctions in any code relying on that invariant, including memory safety issues.

weiznich: This particular case cannot violate the invariant that a `&str` is UTF-8, because we don't use the any unsafe function to convert the `CStr` to a `&str`. Instead we call [`CStr::to_str`](https://github.com/diesel-rs/diesel/blob/4ddaeb072569c5a86532dc62a7690b994e88cb4b/diesel/src/sqlite/connection/statement_iterator.rs#L54-L58) which should be only successful if it's valid UTF8.
Additionally the so constructed string is not directly exposed for any user, but "only" used internally to check against a user provided field label [here](https://github.com/diesel-rs/diesel/blob/4ddaeb072569c5a86532dc62a7690b994e88cb4b/diesel/src/sqlite/connection/sqlite_value.rs#L117), which is a read only access to the underlying memory. This means as far as I can tell nobody outside of diesel itself can access the exposed memory.

weiznich:
The fixed version is not live yet. I've opened this PR to just have something in place to go as soon at the code is ready. I will add another comment here as soon as this is the case.

weiznich:
I've published the 1.4.6 release some minutes back.

crates/diesel/RUSTSEC-0000-0000.md
weiznich: After thinking a bit more about this: @Shnatsel You are right `memory-corruption` is fitting as this corrupts the memory diesel reads from.

Shnatsel:
Thanks!

PR #812, opened 2021-03-06
add known issues with fltk

MoAlyousef:
Hello
This is a PR with known issues with fltk. Please feel free to improve.

### DIFF
--- a/crates/fltk/RUSTSEC-0000-0000.md
+++ b/crates/fltk/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "fltk"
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thanks for reporting!

Are there upstream issues filed? If so, it would be nice to reference them or a release announcement in the URL field.

CI fails because parsing the function specification fails; specifying multiple functions is a bit tricky, see this for an example:
https://github.com/RustSec/advisory-db/blob/master/crates/once_cell/RUSTSEC-2019-0017.md

MoAlyousef:
Only 1 of these issues was filed. The others were found during testing. I've reformatted the function spec.

Shnatsel:
Thanks!

PR #821, opened 2021-03-07
Fix `main` -> `master` mirroring

tarcieri:


Shnatsel:
We can even make it check for 0000-0000 advisories and only mirror if none of those are present. So that anyone on `main` will not see any advisories without IDs assigned.

PR #827, opened 2021-03-18
parse_duration: `parse` DoS through payloads with big exponent

disconnect3d:


crates/parse_duration/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,22 @@
...
+url = "https://github.com/zeta12ti/parse_duration/issues/21"
+categories = ["denial-of-service"]
+keywords = ["parsing", "untrusted data"]
+
+[versions]
+patched = []
+unaffected = []
+
+[affected]
+functions = { "parse_duration::parse" = ["*"] }
```
disconnect3d, line 15: Not sure if this is correct since the vulnerability hasn't been fixed and it's unknown whether it will be.

tarcieri:
It looks like the directory the advisory is inside of doesn't match the package name, i.e.:

`crates/parse_duration` should be `crates/parse-duration`

tarcieri:
Oh whoops, my bad, it looks like the name on crates.io has an underscore: https://crates.io/crates/parse_duration

So: `crates/parse-duration` should be `crates/parse_duration`

Sorry about that

disconnect3d:
Haha; fixed! :)

disconnect3d:
> Oh whoops, my bad, it looks like the name on crates.io has an underscore: https://crates.io/crates/parse_duration
> 
> So: `crates/parse-duration` should be `crates/parse_duration`
> 
> Sorry about that

Hmm the linter says it should be `parse-duration` 🤐

crates/parse_duration/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,22 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "parse-duration"
```
tarcieri, line 4:
```suggestion
package = "parse_duration"
```

crates/parse_duration/RUSTSEC-0000-0000.md
tarcieri: I think you can leave it empty...

tarcieri:
Advisory looks well-formatted now.

I suppose the question remains of whether this fits our DoS policy or not.

If I understand correctly this is an algorithmic DoS as opposed to a simple panic, in a crate which appears designed to act on untrusted data, so I'd vote yes.

@alex @Shnatsel any thoughts?

crates/parse_duration/RUSTSEC-0000-0000.md
disconnect3d: so that should be `functions = {}`?

Shnatsel:
If the crate developers want to report this, I don't see why not include this advisory.

disconnect3d:
> If the crate developers want to report this, I don't see why not include this advisory.

Just to be clear: I am not the `parse_duration` crate developer. I found this issue independently while someone else also reported it a year ago. There was no response from the maintainer or fix since then. Anyway, I think making people aware of this issue by adding it to RustSec (and so e.g. `cargo audit`) may benefit projects who use or would use this crate.

crates/parse_duration/RUSTSEC-0000-0000.md
tarcieri: Whoops sorry, maybe not giving the clearest advice.

On second thought, what you had was fine. You can format it like this which is a little easier to read:

```toml
[affected.functions]
"parse_duration::parse" = ["*"]
```

PR #834, opened 2021-03-26
[patched] Add advisory for use-after-free in rocket

ammaraskar:
Upstream issue: https://github.com/SergioBenitez/Rocket/issues/1534

Marked as `informational = unsound` because as per the original bug thread, it is rare for users to actually implement a custom `UriDisplay` and it's unlikely for such a method to panic since it just deals with string formatting.

### DIFF
--- a/crates/rocket/RUSTSEC-0000-0000.md
+++ b/crates/rocket/RUSTSEC-0000-0000.md
@@ -1,0 +1,31 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rocket"
+date = "2021-02-09"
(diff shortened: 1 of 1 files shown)


alex:
To actually get a UAF in this case, you'd need to catch the panic and then do something else with the `Formatter` that interacted with `self.prefixes`, is that right? Merely dropping the `SmallVec<&'static str>` does not trigger a UAF since there's no `Drop` impl for an `&str`.

ammaraskar:
That's correct yeah, you'd have to access the `Formatter` again on the unwind path e.g in one of your type's `Drop` or through a `catch_unwind`.

alex (approved):

PR #847, opened 2021-03-29
Report out-of-bounds write in stackvec

ammaraskar:
Upstream issue: https://github.com/Alexhuszagh/rust-stackvector/issues/2

Been open for more than a month now.

### DIFF
--- a/crates/stackvector/RUSTSEC-0000-0000.md
+++ b/crates/stackvector/RUSTSEC-0000-0000.md
@@ -1,0 +1,22 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "stackvector"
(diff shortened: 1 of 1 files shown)


Alexhuszagh:
This should be closed as of this [commit](https://github.com/Alexhuszagh/rust-stackvector/commit/f45657d5a823a67bb3f5cffee65efbb401a44192), all versions prior to this fix (up till 1.0.8) have been yanked from [crates.io](https://crates.io/crates/stackvector/versions), and I've gone further and checked the `insert_many` is unlikely to be unsound based on `size_hint`, since it merely uses the lower bound as an optimization, and does bounds checking whenever the number of elements in the in the stackvector exceeds the lower bound hint.

PR #866, opened 2021-04-05
Added patch to "fix" vulnerability.

tiby312:
Marked function as unsafe

### DIFF
--- a/crates/reorder/RUSTSEC-2021-0050.md
+++ b/crates/reorder/RUSTSEC-2021-0050.md
@@ -26,3 +26,11 @@ As noted by the Rust documentation, [`len()`](https://doc.rust-lang.org/std/iter
 and `size_hint()` are primarily meant for optimization and incorrect values
 from their implementations should not lead to memory safety violations.
 
+
+# Patch
+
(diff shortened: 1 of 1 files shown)


crates/reorder/RUSTSEC-2021-0050.md
```diff
@@ -26,3 +26,11 @@ As noted by the Rust documentation, [`len()`](https://doc.rust-lang.org/std/iter
 and `size_hint()` are primarily meant for optimization and incorrect values
 from their implementations should not lead to memory safety violations.
 
+
+# Patch
+
+A new version crate was pushed that makes this function as unsafe.
```
tarcieri, line 32: I assume you meant this?

```suggestion
A new version crate was pushed that marks this function as unsafe.
```

tarcieri:
It'd be good to update this as well:

```toml
[versions]
patched = [">= 1.1.0"]
```

crates/reorder/RUSTSEC-2021-0050.md
tiby312: thanks fixed

PR #868, opened 2021-04-13
Update advisory to indicate patched versions of stackvector.

Alexhuszagh:
A memory corruption issue in the `extend` method of https://github.com/Alexhuszagh/rust-stackvector/issues/2 has been patched in this [commit](https://github.com/Alexhuszagh/rust-stackvector/commit/f45657d5a823a67bb3f5cffee65efbb401a44192). All versions prior to this commit have been yanked from [crates.io](https://crates.io/crates/stackvector/versions) out of precaution, so any available versions should be unaffected.


Shnatsel:
Thanks!

PR #872, opened 2021-04-15
algorithmica: 'merge_sort::merge()' crashes with double-free for `T: Drop`

JOE1994:
Original issue report: https://github.com/AbrarNitk/algorithmica/issues/1

Thank you for reviewing this PR :)

### DIFF
--- a/crates/algorithmica/RUSTSEC-0000-0000.md
+++ b/crates/algorithmica/RUSTSEC-0000-0000.md
@@ -1,0 +1,16 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "algorithmica"
+date = "2021-03-07"
+url = "https://github.com/AbrarNitk/algorithmica/issues/1"
+categories = ["memory-corruption"]
+
+[versions]
+patched = []
+```
+
+# 'merge_sort::merge()' crashes with double-free for `T: Drop`
+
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thanks!

Shnatsel:
@tarcieri looks like the ID assignment is not running on main, and the force push is not triggered either. How is the automation is set up with `master` vs `main`?

tarcieri:
`master` is set up as a (hopefully temporary) mirror of `main`:

https://github.com/RustSec/advisory-db/blob/master/.github/workflows/mirror-main-to-master.yml

PRs shouldn't be merged to `master`. They'll get clobbered.

I used the rename branch feature of GitHub which should've reparented all open PRs to `main`. I guess keep an eye out for PRs opened against `master`.

Shnatsel:
Oh, I was hoping to use `master` as a staging branch and only mirror it to `main` once the IDs are assigned, so that RUSTSEC-0000-0000 IDs are not visible to clients at all. And that way we'd mirror from `master` to `main`, conditionally.

tarcieri:
That'd be a pretty big change from how things work now. I'd prefer to get rid of the `master` branch entirely.

Also the client library already ignores `RUSTSEC-0000-0000`.

If you'd like to do something like that, I'd suggest opening a discussion issue and we can plan it out.

PR #875, opened 2021-04-16
Mark patched tiny-http version for 2020-0031

bradfier:
A backport of the fix for 2020-0031 has been applied to the 0.6.x branch, starting at 0.6.3, subsequent 0.6 versions are no longer vulnerable.

I wasn't quite sure how to express "0.6.x versions are no longer vulnerable where x >= 3", I hope this is correct!


Shnatsel:
Thanks! I've adjusted the version specification to express what you meant. I think the current version specification is not properly documented.

Shnatsel:
I've double-checked that the version specification I wrote works as intended.

PR #883, opened 2021-05-01
Add CVE-2021-3450 for openssl-src

amousset:


Shnatsel:
Thanks again for reporting these! Advisories for source packages like this one is a big missing piece in the database!

PR #891, opened 2021-05-02
im and im-rc are unmaintained

xu-cheng:


xu-cheng:
Any comments?

tarcieri:
We can potentially merge this, but it might be good to give a little more time for a (non-)response for an unmaintained crate than two weeks.

xu-cheng:
May I suggest to merge this PR now? Noted that the crate authors have actually been [unresponsive for quite a long time](https://github.com/bodil/im-rs/pull/154#issuecomment-774583027).

Shnatsel:
This is an interesting situation with, on one hand, the author being clearly unresponsive on the repository, and on the other hand, contributing elsewhere on github and being active on twitter.

Could anyone with a twitter account contact the author and ask about the status of the repo and potentially handing it off to another maintainer?

tarcieri:
@Shnatsel I'll ask

tarcieri:
I received a response that these crates are maintained

decathorpe:
Could this decision be reconsidered? It's been two more months without any activity on the GitHub project, and it's now over a year since an issue was closed or a PR was merged by the upstream developer(s).

tarcieri:
We only mark crates as unmaintained with permission of the author or if there is no response from the author for a prolonged period of time.

In this case we received a response from the author.

decathorpe:
I'd say actions (or rather, inactions, in this case) speak louder than words, but alright :(

arthurprs:
This is concerning. The crate still popular but has critical bugs as described here https://github.com/bodil/im-rs/issues/185. It also hasn't seen updates for over an year even if fixes for the issues were already proposed.

tarcieri:
Have RUSTSEC advisories been filed for all of those bugs?

PR #914, opened 2021-05-19
Add advisory for iced-x86 soundness bug

landaire:
More information here: https://github.com/icedland/iced/issues/168

[Rendered](https://github.com/landaire/advisory-db/blob/iced-x86-advisory/crates/iced-x86/RUSTSEC-0000-0000.md).

### DIFF
--- a/crates/iced-x86/RUSTSEC-0000-0000.md
+++ b/crates/iced-x86/RUSTSEC-0000-0000.md
@@ -1,3 +1,4 @@
+```toml
 [advisory]
 id = "RUSTSEC-0000-0000"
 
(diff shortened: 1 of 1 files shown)


landaire:
From the CI run:

```
error: error loading advisory DB repo from .: RustSec error: parse error: error parsing ./crates/iced-x86/RUSTSEC-0000-0000.md: parse error: unexpected start of advisory: "[advisory]
id = "RUS"
Error: Process completed with exit code 1.
```

Is a newline expected at the top of the file? Otherwise it looks similar to the sample.

Shnatsel:
Hmm, for some reason re-running the workflow installed `rustsec-admin` 0.4.2, which is gonna fail with the same error related to crates.io, so CI is gonna fail here. Weird.

tarcieri:
@Shnatsel the CI config was changed upstream. It would need a rebase to pick up those changes.

Shnatsel:
Cool, this broke CI in main and probably broke `cargo audit` users too :see_no_evil:

tarcieri:
Fixing it

PR #918, opened 2021-05-21
Add RUSTSEC notice for CVE-2021-32629, a Cranelift miscompilation bug.

cfallin:
This is a miscompilation that can result in a Wasm heap sandbox escape.

More details are in our GitHub security advisory at:

https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-hpqh-2wqx-7qp5


Shnatsel:
Thanks for the report!

PR #934, opened 2021-06-07
Update RUSTSEC-2020-0043.md

U007D:
Version of `parity-ws` containing fix changed to read `>=0.10.0', not '>0.10.0' (0.10.0 is the latest as of this writing and contains the fix).


Shnatsel:
Thanks!

PR #939, opened 2021-06-14
crates/grep-cli: add advisory for arbitrary binary execution on Windows

BurntSushi:
Ref https://github.com/BurntSushi/ripgrep/issues/1773

### DIFF
--- a/crates/grep-cli/RUSTSEC-0000-0000.md
+++ b/crates/grep-cli/RUSTSEC-0000-0000.md
@@ -13,7 +13,6 @@ patched = [">= 0.1.6"]
 unaffected = []
 
 [affected]
-#arch = ["x86"]
 os = ["windows"]
 functions = { "grep_cli::DecompressionReader::new" = ["< 0.1.6"] }
 ```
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thanks for the report!

> If you use `grep-cli`'s `CommandReader` with a `std::process::Command` value on Windows, then it is recommended to either construct the `Command` with an absolute binary name, or use `grep-cli`'s new [`resolve_binary`](https://docs.rs/grep-cli/0.1.6/grep_cli/fn.resolve_binary.html) helper function.

To clarify, is the old function with the problematic behavior still accessible through the public API?

CodesInChaos:
You forgot to fill in the title

````
> # RustSec Advisory Template - Advisory Title Goes Here
````

BurntSushi:
@CodesInChaos Thanks, fixed!

@Shnatsel I pushed an update. Does it answer your question?

Shnatsel:
It does. Thanks!

CodesInChaos:
btw is there a github issue for the underlying `std::process::command` pitfall?

BurntSushi:
@CodesInChaos No, but there should be.

PR #942, opened 2021-06-20
Add `withdrawn` field

Shnatsel:
`cargo audit` works normally and does not require any changes., but this makes the current linter barf because of unknown field `withdrawn`.

I have linter with validation for these fields in the `osv` branch. We could publish that as a pre-release, that would also require a pre-release of `rustsec` crate.

Another approach is to extract just the linter changes (they _seem_ to be self-contained) and publish a new release. That should work in theory, but I haven't tested that.

Only @tarcieri currently has the permissions to publish `rustsec-admin`; adding as a reviewer.

### DIFF
--- a/crates/directories/RUSTSEC-2020-0054.md
+++ b/crates/directories/RUSTSEC-2020-0054.md
@@ -6,6 +6,7 @@ date = "2020-10-16"
 informational = "unmaintained"
 url = "https://github.com/dirs-dev/directories-rs"
 yanked = true
+withdrawn = "2021-04-19"
 
 [versions]
 patched = []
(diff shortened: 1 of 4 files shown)


tarcieri:
I can try to take a look at the `osv` branch soon, or if you'd like to extract the `withdrawn`-related changes into a separate branch, that works too

Shnatsel:
I'll get OSV into a mergeable state on Friday. I need to drop some code that's not needed (mostly Git repo parsing) and make OSV export an optional feature. OSV is going to be unstable for now but it's probably worth merging anyway for that pre-release handling fix.

Shnatsel:
Once OSV branch is ready for review, I'll unmark the PR as draft and @mention you.

PR #946, opened 2021-07-02
Update five std CVEs

Qwaz:
Update five std CVEs

### DIFF
--- a/rust/std/CVE-2017-20004.md
+++ b/rust/std/CVE-2017-20004.md
@@ -1,0 +1,16 @@
+```toml
+[advisory]
+id = "CVE-2017-20004"
+package = "std"
+categories = ["thread-safety"]
+date = "2017-04-29"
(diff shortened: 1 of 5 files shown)


Qwaz:
The lint complains that the reported date `2018-08-21` does not match the year of CVE ID `CVE-2019-1010299`, but unfortunately the issue is in the CVE ID itself and both are the correct values. Is there a way to override the lint?

tarcieri:
We could change the linter to only apply that restriction to advisories with `RUSTSEC-*` IDs

Qwaz:
Opened a PR that softens the lint for CVE IDs in rustsec repository :)

tarcieri:
@Qwaz it's deployed now: https://github.com/RustSec/advisory-db/pull/949

Mind rebasing?

Qwaz:
Rebased!

PR #956, opened 2021-07-08
Add rust-ammonia/ammonia#142

notriddle:
XSS vuln in Ammonia.

I don't know how to exploit it without allowing `<iframe>`, which it doesn't allow by default, but that's not an unreasonable thing for someone to do.

rust-ammonia/ammonia#142


Shnatsel:
> I don't know how to exploit it without allowing <iframe>, which it doesn't allow by default, but that's not an unreasonable thing for someone to do.

Would it make sense to add this to the text of the advisory, to help users assess its impact?

notriddle:
Okay, I added a better version of that text to the doc.

Shnatsel (approved):

Shnatsel:
Merged. Thanks for reporting!

PR #961, opened 2021-07-09
`ark_r1cs_std::mul_by_inverse` generated unsound constraints in versions below `0.3.1`

Pratyush:
(Note that there is no violation of Rust UB guarantees; everything is entirely safe)

### DIFF
--- a/crates/ark-r1cs-std/RUSTSEC-0000-0000.md
+++ b/crates/ark-r1cs-std/RUSTSEC-0000-0000.md
@@ -5,6 +5,7 @@ package = "ark-r1cs-std"
 date = "2021-07-08"
 categories = ["crypto-failure"]
 keywords = ["r1cs", "zksnark", "arkworks"]
+url = "https://github.com/arkworks-rs/r1cs-std/pull/70"
 
 [versions]
 patched = [">= 0.3.1"]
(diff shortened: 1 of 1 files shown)


crates/ark-r1cs-std/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "ark-r1cs-std"
+date = "2021-07-08"
+categories = ["unsound", "r1cs"]
```
Shnatsel, line 6: The linter is complaining about nonexistent categories. Please move these to keywords.

Shnatsel (changes requested):
Thanks for reporting! I've pointed out what makes CI fail.

Is there an upstream issue report or pull request we could refer people to for more info?

Pratyush:
> Is there an upstream issue report or pull request we could refer people to for more info?

The corresponding PR is https://github.com/arkworks-rs/r1cs-std/pull/70; should I include the link in the report?

crates/ark-r1cs-std/RUSTSEC-0000-0000.md
Pratyush: Should be fixed now, I think?

tarcieri (approved):

Shnatsel:
Yes, a link to the PR in the `url` field would be nice. It allows people who want more in-depth information to get it easily.

Shnatsel (approved):

Shnatsel:
Thanks!

PR #963, opened 2021-07-13
Add advisory for libsecp256k1

s3krit:


Shnatsel:
Advisories are now in markdown and should include a description of the issue. That's why CI is failing. Example: https://raw.githubusercontent.com/RustSec/advisory-db/main/crates/untrusted/RUSTSEC-2018-0001.md

s3krit:
Oops thanks for that - I was working from old knowledge :sweat_smile:

Shnatsel:
Thanks for the report!

disconnect3d:
Btw can someone add 'functions' list to this advisory so it is known which functions are potentially dangerous with a single click? :)

Also, is there an overflowing PoC somewhere?

disconnect3d:
It seems that the problematic functions were: `Signature::parse` and `Signature::parse_slice`.

PR #965, opened 2021-07-19
Add directory traversal for tar

kpcyrd:
Original report at https://github.com/alexcrichton/tar-rs/issues/238, discovered and reported by @mgjm.

Even though the report is from 2020 it still has 0day status, there's currently no patch. The stale issue was shared with me by @stoeckmann.

### DIFF
--- a/crates/tar/RUSTSEC-0000-0000.md
+++ b/crates/tar/RUSTSEC-0000-0000.md
@@ -4,7 +4,6 @@ id = "RUSTSEC-0000-0000"
 package = "tar"
 date = "2021-07-19"
 url = "https://github.com/alexcrichton/tar-rs/issues/238"
-categories = ["directory-traversal"]
 
 [versions]
 # none, 0day
--- a/crates/tar/RUSTSEC-0000-0000.md
+++ b/crates/tar/RUSTSEC-0000-0000.md
@@ -1,0 +1,58 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tar"
+date = "2021-07-19"
+url = "https://github.com/alexcrichton/tar-rs/issues/238"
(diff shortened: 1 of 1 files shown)


crates/tar/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,58 @@
...
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tar"
+date = "2021-07-19"
+url = "https://github.com/alexcrichton/tar-rs/issues/238"
+categories = ["directory-traversal"]
+
+[versions]
+# none, 0day
+#patched = ["> 0.4.35"]
```
tarcieri, line 11:
```suggestion
patched = []
```

tarcieri:
I asked about the advisory on the linked issue.

Would suggest giving a bit of time to hear back before merging this.

crates/tar/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,58 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tar"
+date = "2021-07-19"
+url = "https://github.com/alexcrichton/tar-rs/issues/238"
+categories = ["directory-traversal"]
```
tarcieri, line 7: There isn't a "directory traversal" category, only a `file-disclosure` one for LFD, which wouldn't seem to apply here.

We should really move to CWE but for now I'd suggest just removing this.

```suggestion
```

kpcyrd:
Oops, I just noticed there's a placeholder here:
```
[affected]
functions = { "tar::Archive::unpack" = ["< 1.2.3"] }
```
It's the correct function but there's no patched version yet.

Issue #826, opened 2021-03-17
totally-safe-transmute is totally unsafe


sanxiyn:
Just in case someone uses https://crates.io/crates/totally-safe-transmute, let's have a permanent advisory against it so that cargo-audit can flag it.

tarcieri:
There's been some active discussion in the past about these sorts of crates and whether they deserve advisories or not. See the discussion of `plutonium` on #275.

I think it might be okay to file an `informational = "notice"` advisory for it.

Lokathor:
The crate is ultimately no worse than running a program in a debugger, then using the debugger to change the program's memory while it's running. All of rust collapses under such circumstances. The worst that you can really say about it is that it's giving you a footgun to speed up the process. You can do the same thing if you run the program in a debugger that changes the memory while the program is paused. Or if someone uses LD_PRELOAD. Or any other number of things.

That said, you should be carefully vetting your dependencies, and the point of the advisory system is to make that vetting easier, so we should naturally have an advisory about this crate.

SoniEx2:
Please don't, we use this in production!

tarcieri:
Coming from the perspective of perhaps a bit of a prankster myself....

I think there are good reasons to differentiate `totally-safe-transmute` from `plutonium` which I previously cited as "case precedent" for these sorts of crates.

Specifically `plutonium` is a procedural macro. I'd personally consider it "expected" that a procedural macro can inject `unsafe`. Perhaps that's a problem that should be tamed, but that's orthogonal to how to consider advisories filed against these kind of joke/art/provocative crates today.

`totally-safe-transmute` implements an unsound API. It does it in a circuitous way, to comedic effect, but from a vulnerability classification perspective I don't think we should treat it any differently from any other crate which contains code resembling the following:

```rust
pub fn quote_unquote_safe_fn_with_unsound_impl(...) {
   unsafe {
      [...]
   }
}
```

SoniEx2:
`totally-safe-transmute` should just optimize to a panic once llvm fixes the noalias stuff and rust stops emitting noalias-breaking async stuff.

slightlyoutofphase:
I'd tend to argue that both `totally-safe-transmute` and `plutonium` are very obvious "joke" crates that aren't actually intended to be used by anyone. Plutonium specifically seems to be trying to make the overall point of "you can't always rely on automated tools to audit your dependencies, and nothing beats actually reading through their source".

RalfJung:
[...]

PR #966, opened 2021-07-22
better-macro has deliberate RCE in proc-macro

jsgf:
It's "Proving A Point" in https://github.com/raycar5/better-macro/blob/master/doc/hi.md but there's
no guarantee that this will remain benign (or is actually benign right now). The crate also has no useful functionality.

### DIFF
--- a/crates/better-macro/RUSTSEC-0000-0000.md
+++ b/crates/better-macro/RUSTSEC-0000-0000.md
@@ -1,0 +1,27 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "better-macro"
+date = "2021-07-22"
+url = "https://github.com/raycar5/better-macro/blob/24ff1702397b9c19bbfa4c660e2316cd77d3b900/src/lib.rs#L36-L38"
+categories = ["code-execution"]
+keywords = ["rce", "proc-macro"]
+
+[affected]
+functions = { "better_macro::println" = ["> 1.0.0"] }
+
+[versions]
+patched = []
+```
+
+# `better-macro` has deliberate RCE to prove a point
+
+[better-macro](https://crates.io/crates/better-macro) is a fake crate which is
+"Proving A Point" that proc-macros can run arbitrary code. This a particularly
+novel or interesting observation.
+
+It currently opens `https://github.com/raycar5/better-macro/blob/master/doc/hi.md`
+which doesn't appear to have any malicious content, but there's no guarantee that
+will remain the case.
+
+This crate has no useful functionality, and should not be used.


tarcieri:
There's been some conflicting opinions on how to handle crates like this. See #826 as some precedent.

Personally I think this one in particular is OK to include, but I'm curious what other people think.

jsgf:
Well, with all these examples I think I'd want them flagged if they turn up in my dep graphs - certainly at a higher pri than the "X is unmaintained" advisories, which are pretty low value from from my POV.

In this case, because its safety depends on whatever the embedded URL is currently pointing to, which can be changed at any moment, it can't be statically audited. At least with the other examples you can fully describe what any specific version would do (perhaps in combination with a specific compiler given the UB games they play).

Shnatsel:
We do carry `informational = "notice"` for some deliberately subversive crates. I can't see why not do so here.

They're not particularly dangerous in their current form, but it's something you don't want in your dependency graph in a production binary.

tarcieri:
That's not really what it was made for, although we have used it for that once.

jsgf:
OK, I refreshed my memory of the other "interesting" cases, specifically plutonium and totally-safe-transmute.

I agree that those two are marginal, but the way I'd frame it is that the RustSec DB is a set of information about crates, particularly various non-standard behaviours. What one does with that information is a local policy choice; just because something is listed doesn't mean one must or should act on it. It really depends on how the advisory interacts with the local threat model.

So in the case of plutonium, I see it as a legitimate part of the conversation about what Rust's safety rules actually mean in practice. But also it has very unusual behaviour for a Rust package, so I'd want to know if it ever turns up in my dep graph, regardless of whether it's functioning as intended. Obviously I could build my own tooling to track packages of note, so it's really a question of whether this is in RustSec's scope, or if it should be. We can leave that for another conversation.

But I think this case is distinct. It's similar to plutonium in that it's trying to make a point, but:
1. The point it's making is that "build systems run arbitrary code from unknown places", which isn't a Rust-specific. But OK.
2. The name `better-macro`, while vague, gives no clue at all to the intent (unlike plutonium and totally-safe-transmute). Someone might read the doc and think it does something useful for them.
3. The mechanism is fundamentally insecure, as it (effectively) does a live http fetch at build time, so even if the code is well known and the URL is currently safe, it could become actually malicious at any point. It's intrinsically unauditable.

So it's possible that someone *could* use this package without real due diligence (perhaps without noticing if none of the mechanisms the proc macro uses to draw attention to itself work in their environment), and in the process introduce an unbounded build-time RCE vulnerability for their downstream dependencies.

I would therefore argue that this crate *does not* work as intended.

tarcieri:
I think it's worth including. We'll see if people disagree, but if so, we can talk about it then.

PR #971, opened 2021-08-07
added CVE-2021-29922

TheAlgorythm:
Added CVE-2021-29922 like described [here](https://sick.codes/sick-2021-015).
@sickcodes Are you okay with reusing your explanation?

### DIFF
--- a/rust/std/CVE-2021-29922.md
+++ b/rust/std/CVE-2021-29922.md
@@ -4,7 +4,7 @@ id = "CVE-2021-29922"
 package = "std"
 date = "2020-01-31"
 url = "https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-29922"
-categories = ["SSRF", "RFI", "LFI"]
+keywords = ["IpAddr", "SSRF", "RFI", "LFI"]
 
 [versions]
 patched = [">= 1.53.0"]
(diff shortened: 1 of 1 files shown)


rust/std/CVE-2021-29922.md
```diff
@@ -0,0 +1,26 @@
...
+package = "std"
+date = "2020-01-31"
+url = "https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-29922"
+categories = ["SSRF", "RFI", "LFI"]
+
+[versions]
+patched = [">= 1.53.0"]
+
+[affected]
+modules = { "std::net" = ["< 1.52.1"] }
```
TheAlgorythm, line 13: The `modules`-field is not specified but as additional information shouldn't hurt and it could be added later down the road, I added it unless it should be removed.

sickcodes:
Yes @TheAlgorythm :)

rust/std/CVE-2021-29922.md
tarcieri: It won’t pass lint with this. Also `paths` can be used to describe a module as well.

sickcodes: Feel free to let me know if my version number is incorrect also

TheAlgorythm: So the `functions`-field doesn't have to be function, it could also be a module, or not?

tarcieri: Aah sorry, was thinking of the older schema.

We moved from `paths` to `functions` because the main intended consumer of this data is call graph analyzers.

I'd suggest either naming specific functions or omitting it, otherwise it won't pass lint.

tarcieri:
I'll go ahead and merge to get this in. Feel free to post follow-ups to make the affected functions more specific.

PR #980, opened 2021-08-10
Update RUSTSEC-2021-0080 [affected] version

kpcyrd:
This replaces a place holder with the actual fixed version.


Shnatsel:
Thanks!

PR #981, opened 2021-08-14
Report vec-const as unsound

saethlin:
It looks like most of the reports are for higher-profile crates, but someone on Reddit already mistook `vec-const` for something they should consider using.

I've added a URL to a simple example that demonstrates one aspect of the unsoundness of this crate. Is anything else needed in that area?

### DIFF
--- a/crates/vec-const/RUSTSEC-0000-0000.md
+++ b/crates/vec-const/RUSTSEC-0000-0000.md
@@ -1,0 +1,16 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "vec-const"
+date = "2021-08-14"
+url = "https://github.com/Eolu/vec-const/issues/1#issuecomment-898908241"
+categories = ["memory-corruption"]
+keywords = ["memory-safety"]
+
+[versions]
+patched = []
+```
+
+# vec-const attempts to construct a Vec from a pointer to a const slice
+
+This crate claims to construct a const `Vec` with nonzero length and capacity, but that cannot be done because such a `Vec` requires a pointer from an allocator.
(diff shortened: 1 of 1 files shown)


crates/vec-const/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,13 @@
+[advisory]
```
tarcieri, line 1: You need to surround the TOML in a ```toml "front matter" block, as well was add a title/description authored in markdown.

```suggestion
```toml
[advisory]
```

tarcieri: You still don't have a ```toml delimiter on the front matter.

Here is another advisory you can look off of:

https://raw.githubusercontent.com/rustsec/advisory-db/c71cfec8c3fe313c9445a9ab0ae9b7faedda850a/crates/smallvec/RUSTSEC-2018-0018.md

saethlin: 🤦 I appreciate your patience. I tried to do this over the sketchiest internet connection and chaos ensued.

tarcieri: It's passing lint now 🎉

crates/vec-const/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,16 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "vec-const"
+date = "2021-08-14"
+url = "https://github.com/Eolu/vec-const/issues/1#issuecomment-898908241"
+categories = ["memory-corruption"]
+keywords = ["memory-safety"]
+
```
tarcieri, line 9: Since this is a soundness advisory and not a security vulnerability, it needs to be marked as such

```suggestion
informational = "unsound"

```

PR #984, opened 2021-08-21
Patched release of reffers

diwic:
I'm not sure anyone uses this old crate but if they do, at least now they have a fixed version.


Shnatsel:
Thanks!

PR #1003, opened 2021-08-23
Fix CVE alias CVE-2020-35920

Shnatsel:
It was linked to the wrong advisory. Thanks to @OS-WS for pointing it out [here](https://github.com/rustsec/advisory-db/commit/7fb26418880c22999ee69be313d713b8b579e5ff#commitcomment-55284597).


OS-WS:
Thank you for the quick response! 
@Shnatsel

PR #1006, opened 2021-08-24
add cve-2021-3711 for openssl-src

amousset:


crates/openssl-src/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,35 @@
...
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "openssl-src"
+aliases = ["CVE-2021-3711"]
+categories = ["crypto-failure"]
+date = "2021-08-24"
+url = "https://www.openssl.org/news/secadv/20210824.txt"
+
+[versions]
+patched = [">= 111.16"]
```
amousset, line 11: not released yet

Shnatsel: Thanks!

We can merge the advisory before a fix is actually published, if it's known that the fix will ship in that specific release.

amousset: :+1: I mainly wanted to avoid the "non-yet-fixable security alert" window as openssl-src releases are usually very quick

tarcieri: Looks like it was just released: https://crates.io/crates/openssl-src/111.16.0+1.1.1l

PR #1015, opened 2021-09-09
Add rustsec advisory for GHSA-v666-6w97-pcwm

darakian:


crates/ckb/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,33 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "ckb"
+date = "2021-07-25"
+url = "https://github.com/nervosnetwork/ckb/security/advisories/GHSA-v666-6w97-pcwm"
+[versions]
+patched = [">= 0.40.0"]
```
Shnatsel, line 8: I see the [Github advisory page](https://github.com/nervosnetwork/ckb/security/advisories/GHSA-v666-6w97-pcwm) lists 0.39.2 as the first patched version, but here you list 0.40.0; is that intentional?

darakian: It is. The 0.39.2 package doesn't exist on crates  
https://crates.io/crates/ckb/versions
the closest available version was picked so that the advisory was more actionable.

PR #1019, opened 2021-09-09
Add rustsec advisory for GHSA-hw4v-5x4h-c3xm

darakian:


crates/frontier/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,15 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "frontier"
+date = "2021-09-01"
+url = "https://github.com/paritytech/frontier/security/advisories/GHSA-hw4v-5x4h-c3xm"
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:L/A:N"
+aliases = ["CVE-2021-39193"]
+[versions]
+patched = []
```
Shnatsel, line 10: The original advisory seems to specify that a given range of git commits is affected. However, `patched = []` would mean that all releases on crates.io are affected. Is that intentional?

crates/frontier/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,15 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "frontier"
+date = "2021-09-01"
+url = "https://github.com/paritytech/frontier/security/advisories/GHSA-hw4v-5x4h-c3xm"
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:L/A:N"
+aliases = ["CVE-2021-39193"]
```
Shnatsel, line 8:
```suggestion
aliases = ["CVE-2021-39193", "GHSA-hw4v-5x4h-c3xm"]
```

crates/frontier/RUSTSEC-0000-0000.md
darakian: Correct. Frontier has exactly one release on crates. I suspect to prevent squatting  
https://crates.io/crates/frontier/versions

Shnatsel: I don't think it makes sense to carry an advisory for the package on crates.io. This issue only affects git versions, it's not clear if the crates.io version is affected, and the crates.io package does not appear to be used at all at any rate.

Are you OK with me closing this?

darakian: Totally fine with closing this if you prefer 👍

PR #1024, opened 2021-09-09
Add rustsec advisory for GHSA-48vq-8jqv-gm6f

darakian:


crates/ckb/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "ckb"
+date = "2021-07-25"
+url = "https://github.com/nervosnetwork/ckb/security/advisories/GHSA-48vq-8jqv-gm6f"
+[versions]
+patched = [">= 0.40.0"]
```
Shnatsel, line 8: Same as in #1015: GHSA specifies 0.39.2 as patched but here it's 0.40.0. Is that intentional?

darakian: Same answer as in #1015; 0.40.0 was chosen to be more actionable.

PR #936, opened 2021-06-08
metrics-util: AtomicBucket<T> unconditionally implements Send/Sync

JOE1994:
Original issue report: https://github.com/metrics-rs/metrics/issues/190

PR #1049, opened 2021-09-18
Add disappeared advisories

Qwaz:
#935 and #936 are merged but do not appear to exist on the current main branch.


Shnatsel:
They probably got lost in the `master` -> `main` migration. Thanks for catching this!

PR #1054, opened 2021-09-24
`#[zeroize(drop)]` not working for `enum`s

daxpedda:
See https://github.com/iqlusioninc/crates/issues/876.
I would appreciate anybody confirming this.

### DIFF
--- a/crates/zeroize_derive/RUSTSEC-0000-0000.md
+++ b/crates/zeroize_derive/RUSTSEC-0000-0000.md
@@ -3,8 +3,6 @@ id = "RUSTSEC-0000-0000"
 package = "zeroize_derive"
 date = "2021-09-24"
 url = "https://github.com/iqlusioninc/crates/issues/876"
-categories = ["crypto-failure", "memory-exposure"]
-keywords = ["zeroize"]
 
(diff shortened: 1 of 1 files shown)


crates/zeroize_derive/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,10 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "zeroize_derive"
+date = "2021-09-24"
+url = "https://github.com/iqlusioninc/crates/issues/876"
+categories = ["crypto-failure", "memory-exposure"]
```
daxpedda, line 6: I'm not too sure about these categories.

Shnatsel:
Thanks for the report! I'll leave the review up to @tarcieri since he should be somewhat familiar with the codebase.

If the bug is confirmed, could you add a description in Markdown? The info from the original issue seems fine. See [this file](https://raw.githubusercontent.com/rustsec/advisory-db/main/EXAMPLE_ADVISORY.md) for an example of formatting.

daxpedda:
Done.

PR #1060, opened 2021-09-27
nix::unistd::getgrouplist buffer overflow

geofft:
Reported in nix-rust/nix#1541, fix currently under review.

### DIFF
--- a/crates/nix/RUSTSEC-0000-0000.md
+++ b/crates/nix/RUSTSEC-0000-0000.md
@@ -9,6 +9,7 @@ keywords = ["nss"]
 
 [versions]
 patched = []
+unaffected = ["< 0.16.0"]
 
 [affected]
 
--- a/crates/nix/RUSTSEC-0000-0000.md
+++ b/crates/nix/RUSTSEC-0000-0000.md
@@ -1,0 +1,34 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "nix"
+date = "2021-09-27"
+url = "https://github.com/nix-rust/nix/issues/1541"
+categories = ["memory-corruption"]
+keywords = ["nss"]
+
+[versions]
+patched = []
+
+[affected]
+
+functions = { "nix::unistd::getgrouplist" = [">= 0.16.0"] }
+```
+
+# Out-of-bounds write in nix::unistd::getgrouplist
+
+On certain platforms, if a user has more than 16 groups, the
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you! Looks good to me. Is there a chance that a fix is going to be released within the next 24 hours or so? In that case I'd prefer to include the fixed version in the advisory and only then merge it in order to make the advisory more actionable.

geofft:
I think that depends on the nix maintainers (we're just contributors), and my guess would be no.

asomers:
Yes, I'll merge that PR and publish soon.  But while this is certainly a bug, is it really a security issue?  It doesn't seem exploitable to me, except by somebody who already has the ability to modify /etc/groups.

asomers:
Fixed in Nix 0.23.0, 0.20.2, 0.21.2, and 0.22.2.

Shnatsel:
Since this appears to require root to exploit, I'd like to drop the severity down to a warning. Thoughts?

asomers:
> Since this appears to require root to exploit, I'd like to drop the severity down to a warning. Thoughts?

I agree with decreasing the severity.  Also, you should set the `os` tag to linux, freebsd, android, netbsd, dragonfly, openbsd, and fuchsia.

Shnatsel:
Merged. Apologies for the delay. Thanks @geofft for the timely and detailed advisory!

geofft:
Thank you, and thanks @asomers for the backports to stable branches (we're unfortunately still running Debian oldoldstable for a few more months).

I agree this is not high-priority, but there are contexts where it's theoretically exploitable by an unprivileged user, including
* At sites with very large numbers of UNIX groups (including ours), there's generally some self-service interface to create groups and add yourself to them. Even if it doesn't explicitly let you pick a specific GID (which it might - ours does, which we didn't consider security-sensitive), you might be able to create one at the right time to get the one you want, and it's definitely easy to get yourself into a large number of groups and cause a DoS.
* In a site with networked groups (e.g., LDAP), while the source of groups is somewhat trusted, it's not necessarily completely trusted. SSSD, for instance, has the [`min_id` parameter](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/configuring_domains), which by default prevents LDAP from returning data for UID/GID 0 and is usually set to prevent it from returning UIDs/GIDs < 1000. So this bug allows lateral movement from the LDAP server to the rest of the environment, where lateral movement may have been hard (though admittedly not impossible, getting control of UIDs >= 1000 is certainly bad too).

PR #1061, opened 2021-09-28
Clarify meaning of RUSTSEC-2021-0077.md

teor2345:
There appear to be some missing words, which unintentionally give a sentence the opposite meaning.


alex (approved):

jsgf:
Thanks for fixing this - I'd noticed it, and thought I'd put up a PR for it, but apparently not.

PR #1076, opened 2021-10-14
Add CVE information to RUSTSEC-2020-0142

alexanderkjall:
See: https://cve.circl.lu/cve/CVE-2020-36462


alexanderkjall:
Funny that you say that, I was just sitting and playing with some code for detecting these automatically :)

I'm all in favor of you becoming a CNA.

tarcieri:
We were partway through the CNA application and held off pending a larger discussion of whether the Rust Foundation should be the CNA instead.

Given the new GHSA integration, I'm wondering if we could leverage them for CVE assignments now?

PR #1077, opened 2021-10-14
add missing cve info to advisories

alexanderkjall:
looks like RUSTSEC-2020-0036 might be a special case, someone got a cve for that the crate is unmaintained.

This was based on a bit of automation effort, this data is based on all cve's that refeferences the rustsec database, and where there isn't a link back. But I have looked at all of the cve pages manually and it looks correct to me.

### DIFF
--- a/crates/abox/RUSTSEC-2020-0121.md
+++ b/crates/abox/RUSTSEC-2020-0121.md
@@ -5,6 +5,7 @@ package = "abox"
 date = "2020-11-10"
 url = "https://github.com/SonicFrog/abox/issues/1"
 categories = ["memory-corruption", "thread-safety"]
+aliases = ["CVE-2020-36441"]
 
 [versions]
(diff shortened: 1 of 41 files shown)


Shnatsel:
Speaking of, we still have no idea who is filing those CVEs, and would very much like to get in touch with them.

Qwaz:
Side note: Some bugs are reported through GHSA and directly receive CVE numbers without corresponding RUSTSEC ID and RustSec reference. I think it would be great if we can automatically detect them as well. e.g., [CVE-2021-32619](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-32619).

Shnatsel:
You don't actually need the `Cargo.toml` files, just the crates.io index, which is conveniently a Git repo: https://github.com/rust-lang/crates.io-index

PR #1085, opened 2021-10-18
Add cvss information from nvd

amousset:
Adds the base cvss information, fetched from nvd using the CVE ids in the aliases.
This was done using the following quick-and-dirty shell script:

```shell
#!/bin/bash

set -e

cd crates

for adv in $(ls -1 */*.md | sort); do
  echo "${adv}"
  # skip if cvss already defined
  if grep -q "^cvss =" "${adv}"; then
    echo "   already has a cvss filed, skipping"
    continue;
  fi
  if ! grep -q "^aliases = .*CVE.*" "${adv}"; then
    echo "   does not have aliases, skipping"
    continue;
  fi
  r_aliases=$(grep aliases ${adv} | cut -d "=" -f2 | sed "s/[,\"\[\]]*//g")
  # define an array
  aliases=(${r_aliases})

  for alias in "${aliases[@]}"; do
    if echo ${alias} | grep -q CVE
    then
      echo "   querying ${alias}"
      vector=$(curl -sS https://services.nvd.nist.gov/rest/json/cve/1.0/${alias} | jq '.result.CVE_Items[0].impact.baseMetricV3.cvssV3.vectorString')
      if [ "$vector" = "null" ]
      then
        echo "   can't find information"
        continue
      fi
      # insert as many cvss line as CVEs is vectors are different
      if ! grep -q "cvss = ${vector}" "${adv}"
      then
        sed -i "/^aliases.*=.*/a cvss = ${vector}" "${adv}"
      fi
      # rate limiting for nvd api
      sleep 0.5
    fi
  done
  # remove cvss for advisories with several different vectors
  num=$(grep -c "cvss =" "${adv}")
  if [ "$num" -gt "1" ]
  then
    sed -i '/cvss =/d' "${adv}"
    echo "   ${adv} has aliases with ${num} different vectors, skipping"
  fi
done
```

* When the advisory has several CVE aliases with different cvss, the advisory is skipped
* When the advisory already has a cvss field, it is skipped

In the future, this could be properly added to an automated curation tool that would keep information in sync with external sources.

We could also add CWE data to the advisories (but I'm not sure it would be relevant for rustsec).

### DIFF
--- a/crates/abi_stable/RUSTSEC-2020-0105.md
+++ b/crates/abi_stable/RUSTSEC-2020-0105.md
@@ -3,6 +3,7 @@
 id = "RUSTSEC-2020-0105"
 package = "abi_stable"
 aliases = ["CVE-2020-36212", "CVE-2020-36213"]
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
 date = "2020-12-21"
 url = "https://github.com/rodrimati1992/abi_stable_crates/issues/44"
 categories = ["memory-corruption"]
--- a/crates/abox/RUSTSEC-2020-0121.md
+++ b/crates/abox/RUSTSEC-2020-0121.md
@@ -6,6 +6,7 @@ date = "2020-11-10"
 url = "https://github.com/SonicFrog/abox/issues/1"
 categories = ["memory-corruption", "thread-safety"]
 aliases = ["CVE-2020-36441"]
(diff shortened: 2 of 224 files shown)


Shnatsel:
Thanks! Before we merge this, we need to check if the [CVE terms of use](https://cve.mitre.org/about/termsofuse.html) are compatible with our CC0 license, since we're importing data from them (not just link to their identifiers).

PR #1093, opened 2021-10-31
Add `flatbuffers` advisory for flatbuffers#6627

jorgecarleitao:
See https://github.com/google/flatbuffers/issues/6627


jorgecarleitao:
Note: this is my first advisory, so extra feedback / nits are more than welcome to get the gist.

crates/flatbuffers/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "flatbuffers"
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H"
+date = "2021-10-31"
+url = "https://github.com/google/flatbuffers/issues/6627"
+
+[affected]
+All Rust code generated by flatbuffers compiler <=2.0
```
tarcieri, line 10: This needs to be TOML, e.g.:

```suggestion
[affected]
patched = [">=2.0"]
```

jorgecarleitao: Thanks. The issue has not been patched yet (maybe I should have removed the `<=2.0`?)

Also, I do not know how to phrase this because it is not just the flatbuffer crate that has an issue; Rust code generated by flatbuffer's compiler has issues also. So, not only the dependency needs to be bumped, but any code generated by flatbuffers' compiler needs to be re-generated as well (once there is a compiler that generates safe code).

tarcieri: If there are no patched versions, you can do `patched = []`

Re: generated code, just make a note of it in the advisory. It's the best we can do since we can't detect it otherwise.

tarcieri:
Thank you!

PR #1099, opened 2021-11-06
add cve info to advisories

alexanderkjall:


Shnatsel:
Thanks a lot!

CI failed because one of the files was legitimately malformed - the `aliases` field should not go under `[versions]`. I've fixed it manually.

PR #1102, opened 2021-11-14
Add fruity advisory for nvzqz/fruity#14

dylni:
Issue: nvzqz/fruity#14

A CVE request has been submitted as well.

Questions:
- Can I add trait implementations to the `functions` list?
- Should I create an unmaintained crate advisory too?
    - nvzqz/fruity#14
    - nvzqz/fruity#13

### DIFF
--- a/crates/fruity/RUSTSEC-0000-0000.md
+++ b/crates/fruity/RUSTSEC-0000-0000.md
@@ -1,0 +1,46 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "fruity"
+date = "2021-11-14"
+url = "https://github.com/nvzqz/fruity/issues/14"
+
+[affected]
+functions = {
+    "fruity::foundation::NSString::to_str" = ["> 0.0.0"],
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you for the report!

Could you clarify why this presents a security issue, ideally in the advisory text? 
It does appear to be a logic bug, but I don't see any evidence of memory corruption or the like.

dylni:
@Shnatsel An impact section has now been added. Does it seem sufficient?

This is easily one of the less severe security issues in many cases, since there are usually other mitigations that would prevent it. However, it was one of the main criticisms of PHP, for instance, so it has caused issues for filename validation in the past:
https://www.php.net/manual/en/security.filesystem.nullbytes.php

Shnatsel:
Yes, the Impact section looks very well-written to me. Merging.

And thank you very much for discovering and reporting the issue!

dylni:
Thanks @Shnatsel!

Should I create an issue to ask if the crate is no longer maintained, or is nvzqz/fruity#13 sufficient?

Also, the `affected.functions` section does not include the affected traits (e.g., `Display`). Is there a way to include those as well?

PR #1103, opened 2021-11-14
Security advisory on simple_asn1 version 0.6.0

nmathewson:
The maintainer has acknowledged and fixed this issue; see
https://github.com/acw/simple_asn1/pull/28 .

This is my first RustSec advisory; please check carefully to see if I have done anything incorrectly.  Thanks!

### DIFF
--- a/crates/simple_asn1/RUSTSEC-0000-0000.md
+++ b/crates/simple_asn1/RUSTSEC-0000-0000.md
@@ -1,0 +1,46 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "simple_asn1"
+date = "2021-11-14"
+url = "https://github.com/acw/simple_asn1/issues/27"
+categories = ["denial-of-service"]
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you very much for discovering and reporting this bug!

We usually don't treat panics as security issues, since they cause denial-of-service at worst, and also too numerous to track accurately. I'll double-check with the crate maintainer if they wish a vulnerability to be filed.

P.S. You might want to also show it off in the [Rust Fuzzing Trophy Case](https://github.com/rust-fuzz/trophy-case).

tarcieri:
I would consider format decoders that act on untrusted data an exception to the rule

Shnatsel:
I have found ~50 panics in those just by myself. Shall I file those 50 as RustSec vulns?

tarcieri:
@Shnatsel if those exist in programs that process untrusted data from the network, I would consider that worrisome, yes

PR #1109, opened 2021-11-18
Adds maintained alternative to slice_deque

LiquidityC:
I decided to fork slice_deque due to the outstanding security
advisory and not getting a response from the creator.

This adds a mention to the original advisory to help people 'patch' the issue.


tarcieri:
Thanks!

PR #1112, opened 2021-11-29
Add advisory for rust-embed path traversal

5225225:
Reported this in https://github.com/pyros2097/rust-embed/issues/159. It is *somewhat* unlikely that people will hit it in a production environment since it does require you to be running in a debug mode, but https://github.com/agora-org/agora/issues/270 is an example of where it was vulnerable.


Shnatsel:
Thanks a lot for discovering and reporting the vulnerability! I'll merge the advisory as soon as CI passes.

PR #1117, opened 2021-12-08
Report `rusqlite` closure lifetime issue

thomcc:
Advisory for issue reported as https://github.com/rusqlite/rusqlite/issues/1048.

This is a soundness hole in several `rusqlite` APIs where you can pass a closure with a too-short lifetime.

Doing so can lead to UAF. As a small mercy, it's somewhat unlikely to have code that uses these API incorrectly but silently works — most usage should either be fine, unless the code has not been tested at all.

That said, this is obviously not guaranteed — I've backported it to all affected major versions (0.25.x before 0.25.4 and 0.26.x before 0.26.2), and will yank the affected releases after filing this issue, so there should be no reason not to update.

### DIFF
--- a/crates/rusqlite/RUSTSEC-0000-0000.md
+++ b/crates/rusqlite/RUSTSEC-0000-0000.md
@@ -1,0 +1,42 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rusqlite"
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thanks for such a detailed advisory! Merging.

PR #1124, opened 2021-12-20
brotli2-sys CVE-2020-8927

insanitybit:
Hey, haven't submitted anything before. Brotli reported CVE-2020-8927. The brotli-sys crate is appears to link to a vulnerable version. I'm unsure if the brotli2 crate itself is vulnerable, but I think so based on the following:

https://docs.rs/brotli2/latest/src/brotli2/raw.rs.html#180-195

I also didn't see any limits imposed on the streaming reader/writer that would prevent those interfaces from being vulnerable.

### DIFF
--- a/crates/brotli2-sys/RUSTSEC-0000-0000.md
+++ b/crates/brotli2-sys/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+
+package = "brotli2-sys"
+
+date = "2021-12-20"
+
+categories = ["memory-corruption"]
+
+keywords = ["integer-overflow"]
+
+related = ["CVE-2020-8927"]
+
+[affected]
+
+# There isn't a patch for brotli2-sys, but version 1.0.9 of google/brotli is patched
+[versions]
+patched = []
--- a/crates/brotli2-sys/RUSTSEC-0000-0000.md
+++ b/crates/brotli2-sys/RUSTSEC-0000-0000.md
@@ -1,3 +1,4 @@
+```toml
 [advisory]
 id = "RUSTSEC-0000-0000"
 
@@ -5,14 +6,24 @@ package = "brotli2-sys"
 
 date = "2021-12-20"
 
+url = "https://github.com/bitemyapp/brotli2-rs/issues/45"
+
 categories = ["memory-corruption"]
 
 keywords = ["integer-overflow"]
 
-related = ["CVE-2020-8927"]
+aliases = ["CVE-2020-8927"]
 
 [affected]
 
 # There isn't a patch for brotli2-sys, but version 1.0.9 of google/brotli is patched
 [versions]
 patched = []
+```
+# Integer overflow in brotli2
+
+A buffer overflow exists in the Brotli library versions prior to 1.0.8 where an attacker controlling the input length of a "one-shot" decompression request to a script can trigger a crash, which happens when copying over chunks of data larger than 2 GiB.
+
+It is recommended to update your Brotli library to 1.0.8 or later. 
+
+If one cannot update, we recommend to use the "streaming" API as opposed to the "one-shot" API, and impose chunk size limits.


crates/brotli2-sys/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,18 @@
...
+
+package = "brotli2-sys"
+
+date = "2021-12-20"
+
+categories = ["memory-corruption"]
+
+keywords = ["integer-overflow"]
+
+related = ["CVE-2020-8927"]
```
Shnatsel, line 12:
```suggestion
aliases = ["CVE-2020-8927"]
```

Shnatsel (changes requested):
Thank you for the report!

CI failed because the file format is not just TOML, it's Markdown with embedded TOML content - you can use [this](https://raw.githubusercontent.com/rustsec/advisory-db/main/crates/openssl-src/RUSTSEC-2021-0129.md) as a template.

It would also be nice to include a description of the vulnerability. Copy-pasting from the CVE would be sufficient.

tarcieri:
Have a look at an example advisory for the format. It's Markdown with TOML "front matter":

https://raw.githubusercontent.com/rustsec/advisory-db/main/crates/openssl-src/RUSTSEC-2021-0057.md

Shnatsel:
Also, please report the issue to the brotli-rs issue tracker if you haven't already.

Shnatsel:
Since no fix is available, perhaps we should mention that the pure-Rust [`brotli`](https://crates.io/crates/brotli) crate is not affected, and is a valid although higher-effort mitigation path? @tarcieri what do you think?

tarcieri:
Sounds good to me

insanitybit:
Thanks for the quick response and feedback. I've pushed up a revision that should match your requests, and I've also created in issue in the brotli2-rs repository https://github.com/bitemyapp/brotli2-rs/issues/45.

colin-grapl:
From what I can tell it may be fairly straightforward to migrate to the `brotli` crate for code currently using `brotli2`. 

For example,

`brotli2::read::BrotliDecoder` and `brotli::reader::Decompressor` have very similar APIs.

* Decompressor implements all non-trait api compatible methods (as well as a dictionary method)
* Decompressor implements Read

It even appears that the `brotli` crate exports a C API (`brotli::ffi`) that's compatible, so one could potentially just swap out the packaged `brotli` for the crate.

jorgecarleitao:
I would refrain from making recommendations about using other crates here - imo that is a discussion for reddit, forums, etc. Imo the scope here is to disclose the vulnerability, which code branches hits it, and how to mitigate it within the dependency taken (e.g. avoid using function X with parameters Y).

crates/brotli2-sys/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,29 @@
...
+
+categories = ["memory-corruption"]
+
+keywords = ["integer-overflow"]
+
+aliases = ["CVE-2020-8927"]
+
+[affected]
+
+# There isn't a patch for brotli2-sys, but version 1.0.9 of google/brotli is patched
```
Shnatsel, line 19: I understand version 1.0.8 is also patched?

crates/brotli2-sys/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,29 @@
...
+[versions]
+patched = []
+```
+# Integer overflow in brotli2
+
+A buffer overflow exists in the Brotli library versions prior to 1.0.8 where an attacker controlling the input length of a "one-shot" decompression request to a script can trigger a crash, which happens when copying over chunks of data larger than 2 GiB.
+
+It is recommended to update your Brotli library to 1.0.8 or later. 
+
+If one cannot update, we recommend to use the "streaming" API as opposed to the "one-shot" API, and impose chunk size limits.
```
Shnatsel, line 29: I understand the issue with the brotli-rs package specifically is that it bundles an old version of the C implementation of Brotli? Perhaps this should be explicitly called out?

crates/brotli2-sys/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,29 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+
+package = "brotli2-sys"
```
Shnatsel, line 5: CI complains that there is no package `brotli2-sys` on crates.io; I assume you meant `brotli-sys`?

Shnatsel (changes requested):

tarcieri:
> I would refrain from making recommendations about using other crates here

@jorgecarleitao we frequently provide curated advice about alternatives, although typically for abandoned crates as opposed to the context of a security vulnerability

> imo that is a discussion for reddit, forums, etc.

My personal opinion: that's a terrible antipattern. Far too often people reach for the first recommendation they see in the context of a reddit/forum thread, but those media can't provide curated advice. This results in outdated, unreviewed advice becoming standard practice just because it happens to be the top hit on Google et al for a particular search term.

With an advisory we can provide up-to-date advice which evolves along with the crate ecosystem, rather than one person's dated opinion.

Shnatsel:
I wanted to get this merged quickly, but this PR does not allow edits by maintainers, so I've created a new PR: #1127.

Shnatsel:
I went ahead and merged #1127 not to delay the notification to users any longer. I'm open to amendments in case I missed something.

Thanks again for reporting the issue!

insanitybit:
Awesome, thanks for taking care of that.

PR #1125, opened 2021-12-21
Add use after free advisory for lru crate

oherrala:
# Use after free in lru crate

Lru crate has use after free vulnerability.

Lru crate has two functions for getting an iterator. Both iterators give
references to key and value. Calling specific functions, like pop(), will remove
and free the value, and but it's still possible to access the reference of value
which is already dropped causing use after free.

More information in upstream repository issue: https://github.com/jeromefroe/lru-rs/issues/120 and the already released fix can be found from https://github.com/jeromefroe/lru-rs/pull/121


Shnatsel:
Thanks a lot!

PR #1131, opened 2021-12-22
Mark arrow advisories as fixed in 6.4.0

saethlin:
It looks to me like all the outstanding arrow-rs advisories were fixed, as mentioned in https://github.com/apache/arrow-rs/issues/817, and those fixes were released in version 6.4.0. @alamb can you confirm?

### DIFF
--- a/crates/arrow/RUSTSEC-2021-0116.md
+++ b/crates/arrow/RUSTSEC-2021-0116.md
@@ -8,7 +8,7 @@ categories = ["memory-exposure"]
 keywords = ["buffer-overflow"]
 
 [versions]
-patched = []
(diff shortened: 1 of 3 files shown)


alamb:
Thank you @saethlin  -- yes the issues covered by the outstanding advisories I think has been fixed in 6.4.0.

As an aside, there are two other tickets that may deserve (different) advistories:
* https://github.com/apache/arrow-rs/issues/786
* https://github.com/apache/arrow-rs/issues/197

But I think we should have a reproducer / demonstration for them, which I have not had a chance to do yet. I am hoping (!) to find time over the holidays to do so

PR #1132, opened 2021-12-25
Mark cargo-download unmaintained

pinkforest:
Use of this crate will result into corrupted crate downloads due to lacking flush into filesystem.

No activity in over a year either in the repo or from the publisher(s)

Issue where maintenance takeover has been proposed around 2 Oct
https://github.com/Xion/cargo-download/issues/10

Critical PR that would resolve the write flush
https://github.com/Xion/cargo-download/pull/11


Shnatsel:
Thanks!

PR #1134, opened 2021-12-27
Turn the issue about shamir into an advisory

saethlin:
Closes https://github.com/rustsec/advisory-db/issues/235


Shnatsel:
Thanks! Not sure how we missed that one, since it has a TOML and everything.

PR #1138, opened 2021-12-28
cookie-rs is unmaintained

spikecodes:
It appears @SergioBenitez hasn't made any GitHub contributions since September so it would be safe to mark `cookie-rs` as unmaintained.

image


SergioBenitez:
As with #1139.

PR #1139, opened 2021-12-28
rocket is unmaintained

spikecodes:
It appears @SergioBenitez hasn't made any GitHub contributions since September so it would be safe to mark Rocket as unmaintained.

image


SergioBenitez:
Rocket isn't abandoned. I think this advisory will cause more harm than good.

alex:
Good enough for me, thanks @SergioBenitez.

PR #1140, opened 2022-01-01
Add advisory for rustc_serialize

5225225:
rustc_serialize is unmaintained (I'm assuming, there's a note in the readme that bugfixes will be merged, but the [repo itself](https://github.com/rust-lang-deprecated/rustc-serialize) is archived, so I suppose that's not happening), is it worth filing just an unmaintained advisory / filing two different advisories?


Shnatsel:
Thanks for the report! The advisory looks good to me, merging.

A separate advisory for `rustc-serialize` being unmaintained would also be welcome.

Issue #1092, opened 2021-10-28
Process for advisories without a fix?


saethlin:
@Shnatsel wanted to discuss this https://github.com/rustsec/advisory-db/pull/1079#issuecomment-945198609 and I feel like now is a pretty good time because `arrow-rs` has had this issue open about a soundness bug for a month now without a fix or an advisory: https://github.com/apache/arrow-rs/issues/777 and I think I just found one or two more soundness defects.

So what should be done? Notify the crate author(s) and give them some number of days to patch before issuing an advisory? Do we track time from when an issue is opened identifying the problem as a soundness issue, or from when we tell them that this has prompted a RustSec advisory?

jorgecarleitao:
I agree with incentivizing maintainers to fill their own advisories on every publicly known vulnerability. 

I think Tokyo offered us a good example:

* Unsoundness disclosed on the 2021-7-6 11:45 CET: https://github.com/tokio-rs/tokio/issues/3929
* Patch released on the 2021-7-6 20:06 CET: https://github.com/tokio-rs/tokio/pull/3934
* Advisory PRed on the 2021-7-7 23:47 CET: https://github.com/rustsec/advisory-db/pull/951

(All actions were executed by the maintainer)

It is unfortunate that both developers and maintainers act differently if/when the vulnerability is presented in advisory-db because it incentivizes to _not_ being transparent. IMO we should work towards changing this: imo advisories should a "chore" that maintainers perform as part of a disclosure process; it should not have a negative connotation associated to it - bugs are a fact of life, we are just increasing transparency about them given their implications to infosec.

IMO the decision tree goes a bit like this:

1. maintainers do have the bandwidth to patch:
    * they release a patch and document it in an advisory
    * users update
2. maintainers do not have the bandwidth to patch:
    * they document the issue in an advisory
    * users audit their own code, fix uses of unsound APIs, and add exception to `.cargo/audit.toml` (e.g, like [here](https://github.com/influxdata/influxdb_iox/blob/main/.cargo/audit.toml#L23) or [here](https://github.com/jorgecarleitao/arrow2/blob/main/.cargo/audit.toml#L3))
3. maintainers do not have the bandwidth to document the vulnerability:
    * the community steps in and documents it instead

In this direction, I would consider re-wording the first notification to something like:

> This appears to be an instance of use-after-free, which is usually an exploitable security vulnerability.
>
[...]

PR #1141, opened 2022-01-04
Add rs-async-zip Zip Traversal issue to DB

snoopysecurity:
A Zip Traversal issue exists in rs-async-zip. Makes sense to add it. Examples of other libraries that have fixed it can be seen here: https://github.com/snyk/zip-slip-vulnerability#affected-libraries. 

To reproduce this issue, you can follow the steps here: https://gist.github.com/snoopysecurity/007503097536b557bc22a7ef24f4d11d

This issue was privately disclosed to the maintainer. He has stated that he has made conscious decision not to mitigate any sort of traversal attacks within the library itself. He has however added a notice to to the example code like here https://github.com/Majored/rs-async-zip/blob/main/examples/fs_parallel_extract.rs#L10 but the library itself does not do any sanitization.

### DIFF
--- a/crates/async_zip/RUSTSEC-0000-0000.md
+++ b/crates/async_zip/RUSTSEC-0000-0000.md
@@ -1,0 +1,57 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "async_zip"
+date = "2022-01-04"
+url = "https://gist.github.com/snoopysecurity/007503097536b557bc22a7ef24f4d11d"
+keywords = ["zip traversal"]
+
+[versions]
+patched = []
+
+[affected]
+functions = { "async_zip::read::fs::ZipFileReader::new" = ["*"] }
+```
+
+# Zip Traversal in async_zip when extracting files from a Zip
+
+ZIP Path traversal is possible during extraction due to no validation and sanitization of filenames.
+
+Example code that triggers the vulnerability:
+
+```rust
+use std::path::Path;
+use std::sync::Arc;
+
+use async_zip::read::fs::ZipFileReader;
+use tokio::fs::File;
+
+#[tokio::main]
+
+async fn main() {
+
+    let zip = Arc::new(ZipFileReader::new("/maliciouszip.zip").await.unwrap());
+
+    println!("Extracting Archive");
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you for reporting this!

This is a tricky case with upstream being unwilling to fix the issue. We have carried similar advisories in the past, but that caused some issues, see #1092. We wanted to get the moderation team to take a look, but the moderation team has stepped down at an inopportune moment...

So we don't really have a good story around issues that upstream is unwilling to fix. However, I don't think carrying the advisory alone is going to solve anything; we need to talk to the maintainer and either get more prominent warnings put up on the repo (in README and on docs.rs), or point to other projects that have mitigated this vulnerability (e.g. `tar`, see [RUSTSEC-2018-0002](https://rustsec.org/advisories/RUSTSEC-2018-0002.html) and [RUSTSEC-2021-0080](https://rustsec.org/advisories/RUSTSEC-2021-0080.html)) and ask to perform similar checks, using those crates as precedent.

Majored:
Heyo. Maintainer here.

I'm failing to see why an advisory notice would be needed here. Directory traversal is a very clear and present issue, I don't disagree, but my library doesn't provide any functions/helpers/types that directly target extracting a ZIP to the disk (and I don't plan on ever adding such either).

Both of the examples provided by @Shnatsel target specific functions in "tar-rs" that extract an entry to the disk.

My crate is intended to be a raw reading/writing crate. You could entirely work with ZIPs in-memory and directory traversal isn't ever an issue. Sure, if you incorrectly use the library and specifically target extraction to the disk, you could be vulnerable to directory traversal - but that's up to the library implementer.

No element of the crate is vulnerable to directory traversal - only the two examples provided in the source GitHub repository, which as already noted, has had clear notices added to them.

Majored:
To further my point, Java's ZipFile/ZipInputStream classes don't contain any mitigation either - do we now consider these two classes, which have been around since 2001, vulnerable to directory traversal?

Clearly not, because neither class targets extraction to disk - they're raw readers, just like my crate.

I appreciate @snoopysecurity's disclosure regarding the repository examples, but to label/push the entire crate as vulnerable is objectively incorrect.

PR #1146, opened 2022-01-10
Undefined behavior in `dashmap`

baptiste0928:
Undefined behavior may occur when using the `Ref::value` of `dashmap`. See https://github.com/xacrimon/dashmap/issues/167 for more information.

Example vulnerable code:
```rust
use dashmap::DashMap;

fn main() {
    let x = DashMap::new();

    x.insert(1, 1);

    let borrow = x.get(&1).unwrap();
    let val = borrow.value();
    std::mem::drop(borrow);
    let mut borrow_mut = x.get_mut(&1).unwrap();
    let val_mut = borrow_mut.value_mut(); // non-exclusive mutable reference!!!
    println!("{}", val);
}
```

I'm unsure about in which categories this issue falls and what to write in the description.

### DIFF
--- a/crates/dashmap/RUSTSEC-0000-0000.md
+++ b/crates/dashmap/RUSTSEC-0000-0000.md
@@ -5,10 +5,11 @@ package = "dashmap"
 date = "2022-01-10"
 url = "https://github.com/xacrimon/dashmap/issues/167"
 categories = ["memory-corruption"]
-keywords = ["segfault"]
+keywords = ["segfault", "use-after-free"]
 
 [versions]
 patched = []
+unaffected = ["< 5.0.0"]
 ```
 
 # Unsoundness in `Ref::value`
--- a/crates/dashmap/RUSTSEC-0000-0000.md
+++ b/crates/dashmap/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "dashmap"
+date = "2022-01-10"
+url = "https://github.com/xacrimon/dashmap/issues/167"
+categories = ["memory-corruption"]
+keywords = ["segfault"]
+
+[versions]
+patched = []
+```
+
+# Unsoundness in `Ref::value`
+
+Reference returned by `Ref::value` may outlive the `Ref` and escape the lock. This cause undefined behavior and may result in a segfault.
+
+More information in [`dashmap#167`](https://github.com/xacrimon/dashmap/issues/167) issue.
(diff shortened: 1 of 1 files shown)


crates/dashmap/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,18 @@
...
+id = "RUSTSEC-0000-0000"
+package = "dashmap"
+date = "2022-01-10"
+url = "https://github.com/xacrimon/dashmap/issues/167"
+categories = ["memory-corruption"]
+keywords = ["segfault"]
+
+[versions]
+patched = []
+```
```
Shnatsel, line 12:
```suggestion
```unaffected = ["< 5.0.0"]
```

I've talked to the `dashmap` maintainer and they said that 4.x should be unaffected.

Shnatsel (changes requested):
Thanks for the report! I've left a comment, looks good otherwise.

This seems to be a use-after-free, in case you want to categorize it further.

crates/dashmap/RUSTSEC-0000-0000.md
baptiste0928: I also tested on my side, only version `5.0.0` is affected.

crates/dashmap/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,19 @@
...
+keywords = ["segfault", "use-after-free"]
+
+[versions]
+patched = []
+unaffected = ["< 5.0.0"]
+```
+
+# Unsoundness in `Ref::value`
+
+Reference returned by `Ref::value` may outlive the `Ref` and escape the lock. This cause undefined behavior and may result in a segfault.
```
LeSeulArtichaut, line 17: (Typo)
```suggestion
Reference returned by `Ref::value` may outlive the `Ref` and escape the lock. This causes undefined behavior and may result in a segfault.
```

crates/dashmap/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,19 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "dashmap"
+date = "2022-01-10"
+url = "https://github.com/xacrimon/dashmap/issues/167"
+categories = ["memory-corruption"]
```
LeSeulArtichaut, line 7: Is there potential for memory exposure as well?

crates/dashmap/RUSTSEC-0000-0000.md
ray-kast: I believe in addition to `Ref::value` every method with a changed lifetime in [this PR](https://github.com/xacrimon/dashmap/pull/168/files) is affected by the same issue.

ray-kast:
(bump from a reply I accidentally added to a resolved conversation)
I believe in addition to `Ref::value` every method with a changed lifetime in [this PR](https://github.com/xacrimon/dashmap/pull/168/files) is affected by the same issue.

Shnatsel:
Merging. Thanks for the report and for the detailed advisory!

PR #1159, opened 2022-01-21
Add advisory for tower-http#204

davidpdrsn:


Shnatsel:
Looks good to me, merging. Thanks for the report!

Roguelazer:
this appears to have been made in the wrong year; would it be worth fixing the date to be 2022 instead of 2021?

Shnatsel:
Ah, thanks for catching that! I've opened #1165 to fix that

davidpdrsn:
> this appears to have been made in the wrong year; would it be worth fixing the date to be 2022 instead of 2021?

Ah ups. Sorry about that 😞

PR #1162, opened 2022-01-21
Add soundness advisory for array-macro

xfix:
This is somewhat ancient unsoundness issue (like, one that was fixed 20 months ago), but may as well report it, even if really nobody will care. It may have some educational value however - somebody may see this issue and realize there is a similar issue in macros they wrote. See https://github.com/rust-lang/unsafe-code-guidelines/issues/278 for more context.


Shnatsel:
Thanks! Yeah, having this tracked sounds like a good idea for analysis purposes. I know some folks are looking through the advisories to get a feel for what kind of vulnerabilities even exist.

PR #1165, opened 2022-01-21
fix year in tower-rs advisory

Shnatsel:


Shnatsel:
@tarcieri any tips on how to fix the year without causing a future advisory to be assigned the same ID?

tarcieri:
@Shnatsel you'd need to yank the advisory and republish it

pinkforest:
Closing this one as doing the yank/re-create in 2022 move

https://github.com/rustsec/advisory-db/pull/1319 - yank from 2021
https://github.com/rustsec/advisory-db/pull/1320 - re-create into 2022

PR #1166, opened 2022-01-22
Update which tower-http versions are affected by RUSTSEC-2021-0135

davidpdrsn:
I have backported the fix for RUSTSEC-2021-0135 to tower-http 0.1.3 so that version is no longer affected. 0.2.0 does have the bug so that is still insecure. I _think_ I got the versions right 🤞


Shnatsel:
Yep, looks good. Thanks a lot!

PR #1167, opened 2022-01-22
Add unmaintained crate advisory for `ftd2xx-embedded-hal`

newAM:
Has been renamed to `ftdi-embedded-hal`.


Shnatsel:
Thanks!

You might also want to push an update to the `ftd2xx-embedded-hal` crate so that the notice about the new name also shows up on crates.io.

PR #1169, opened 2022-01-24
Add advisory for Amanieu/thread_local-rs#33

ibraheemdev:
https://github.com/Amanieu/thread_local-rs/issues/33.


Shnatsel:
Thanks a lot for reporting the issue!

The directory should be called `thread-local`, not `thread_local` - that's what is causing CI to fail. Otherwise looks good!

PR #1176, opened 2022-02-03
Add advisory for unicase

xfix:
I think this is a valid vulnerability, see CVE-2000-0979, CVE-2002-1374 and CVE-2014-6394 for similar vulnerabilities. In short, unicase used to consider `"abc"` and `"a"` to be equal.

### DIFF
--- a/crates/unicase/RUSTSEC-0000-0000.md
+++ b/crates/unicase/RUSTSEC-0000-0000.md
@@ -1,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "unicase"
+date = "2019-11-02"
+url = "https://github.com/seanmonstar/unicase/issues/38"
+keywords = ["partial string comparison"]
+
+[versions]
+patched = [">= 2.6.0"]
+unaffected = ["< 2.0.0"]
+```
+
+# Incorrect check for string equality
+
+Affected versions of this crate did not properly check for string equality allowing substrings of a string to be considered equal.
+
+This can be used to bypass protection mechanisms.
+
+The flaw was corrected in commit [c14856b](https://github.com/seanmonstar/unicase/commit/c14856b0ef9b2e6d8dce802d1a9d5b83a7ce2118) by modifying the code to use a manual implementation of `Iterator::eq` rather than `Iterator::all`.


pinkforest:
Thanks for noticing it - It seems like the fixed one is latest 2.x release track

Affected versions have not been yanked - including non-semver for 1.x release track that has no fix if anyone is stuck on that.

Whilst I would personally agree that it is a valid vulnerability -

Controversial aspect here is that whilst non-controversially this was reported as a test failure - with the fact that the maintainer acted very soon after to fix it - did not acknowledge and announce / agree either implicitly or explicitly on any security related issue making it controversial to write advisory without maintainer providing response around the security implication.

Could we get a simple informed response from the maintainer on this and outlay the CVE vector to @seanmonstar and perhaps allow the maintainer to yank the affected versions ?

@amousset would like to get your opinion as well on this

amousset:
About the relevance of an advisory, I'm not sure it is clear with current policies but I think it makes sense. We could say that is not a vulnerability in itself, and that a lot of other bugs could be a sources of vulnerabilities is dependent crates, without justifying an advisory for the upstream lib. But given the bug and the way this library could be used for ACLs/Headers matching (like in [jsonrpc-http-server](https://github.com/paritytech/jsonrpc/blob/68511bc10ae1e0abb2d6a3f8978cc2e9d962a57c/server-utils/src/cors.rs#L241)/etc. which seems to use it to filter allowed HTTP headers), it feels a bit like the [netmask vulnerability](https://nvd.nist.gov/vuln/detail/CVE-2021-28918) that affected the npm ecosystem, and was a CVE on its own.

I agree that we should try to get the opinion of the maintainers about opening an advisory anyway.

amousset:
After a closer look, I don't see cases in most downloaded dependent crates where this leads to actual vulnerabilities (especially as the `Ascii` comparison is unaffected), except maybe for [interledger-rs](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-service/src/username.rs#L23) (I opened https://github.com/interledger-rs/interledger-rs/pull/744 to update the dependency).
Therefore it's not that severe, but an advisory could still be relevant.

pinkforest:
I propose the below actions:

If the maintainer is not going to address this in two weeks we merge as *informational unsound - 

*This will require a change from @xfix to change the advisory to informational / unsound

If the maintainer responds, yanks or otherwise addresses it we don't need an advisory for the older versions

amousset:
"informational = unsound" has a [specific meaning](https://docs.rs/rustsec/latest/rustsec/advisory/enum.Informational.html#variant.Unsound) that does not match this case. It could maybe be a "notice" but it does not seem to have been used yet, so I'm not sure we want to use it (as it could break some tooling made around the db).

pinkforest:
@tarcieri What is the guidance around this ? as informational necessary doesn't require concrete security issue but we have concerns but we are unsure about the terminology effect on this one.

Our unsound in the guideline is defined differently and broadly:
`informational = "unsound" is used for issues that can only be triggered by a programmer (as opposed to e.g. a malicious input), and/or require very contrived code to trigger.`

unsafe-code-guidelines WG:
```
Unsound

Crate is not [sound](https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#soundness-of-code--of-a-library), i.e., unsound.

A crate is unsound if, using its public API from safe code, it is possible to cause [Undefined Behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html).
```

I think the undefined behaviour defined by the WG Unsafe code guidelines per above is different as unsafe WG only considers it in the context of unsafe and nothing else.

seanmonstar:
I certainly could be wrong in this, so take with a grain of salt. But this seems like a slippery slope of marking all bugs as security vulnerabilities. I could see yanking perhaps being a better case here. Even then, though, I think simply fixing the bug in this case is sufficient. :shrug:

Though, with regards to "unsound", this bug is not what Rust usually calls "unsound", which is behavior that can trigger undefined behavior in the compiler. The bug type is closer to just "incorrect".

pinkforest:
@seanmonstar thank you for taking time and offering to yank the old versions - that would certainly alleviate my concerns :purple_heart:

pinkforest:
Since there are no identified downstream usage which could drive any indication that there are security issues related this bug, I am going to close this one since additionally we do not have any recognised category together with the maintainer indicating that there are no security issues to action around this bug.

We'll have to have wider discussion around binary / library relationships in the Rust ecosystem and how to deal with them.

I've recommended if the old versions could be yanked but that will be up to the maintainer if @seanmonstar wishes to do so.

Re-open if there is identified impacted downstream usage and we can have another discussion.

PR #1177, opened 2022-02-04
Add advisory for windows

xfix:


Shnatsel:
Looks good to me. Thanks for reporting!

I'll try to get to your `unicase` advisory today as well, but I want to get maintainer buy-in before we publish an advisory.

PR #1184, opened 2022-02-07
Mark tokio-proto as deprecated

saethlin:


Shnatsel:
Looks good. Merging, thanks!

PR #1186, opened 2022-02-07
Update RUSTSEC-2022-0009.md

thomaseizinger:
We published a semver compatible upgrade that includes the security fix.


thomaseizinger:
So `0.31.0` does _not_ contain this fix. Do we need to say that somehow in the advisory? `>= 0.30.2` is not strictly correct now. Should we yank that one?

Shnatsel:
We support all the Cargo operators; I've used `^` to specify that 0.30.2 and later versions in 0.30.x series are patched.

mxinden:
Thanks @Shnatsel for your help! :heart:

PR #1189, opened 2022-02-09
Mark mio-uds as deprecated

saethlin:
`mio-uds` has been deprecated for almost 2 years now, but it's still seeing >10,000 daily downloads. But it looks like most of those downloads are coming through `tokio-uds` and `tokio-signal`, both of which also seem to be deprecated. I've tried to trace back the common users of these crates too, but it seems like all the git repos I end up on for the crates don't depend on them anymore on the latest commit. The biggest user I was able to track down was `websocat`.

So I really have no idea how disruptive publishing this would be. It seems fair to expect people to migrate after almost 2 years, but this crate may also be deep in people's dependency trees, so I wouldn't be surprised if people have CI configs that start to break because they're denying informational advisories.

### DIFF
--- a/crates/mio-uds/RUSTSEC-0000-0000.md
+++ b/crates/mio-uds/RUSTSEC-0000-0000.md
@@ -1,0 +1,16 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "mio-uds"
+date = "2020-02-09"
+informational = "unmaintained"
+url = "https://github.com/deprecrated/mio-uds/commit/32d27e91be614936194d8555bdda8d8a776a9b8e"
+
+[versions]
+patched = []
+unaffected = []
+```
+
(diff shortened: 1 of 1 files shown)


Shnatsel:
We generally reserve the unmaintained advisories for crates whose maintainers are completely unresponsive.

I see that the repo is under an organization that's dedicated to maintaining it? I would expect to merge and publish a critical security patch, if one appears.

Combined with the apparent lack of high-profile users, I'm not sure I see value in this advisory.

Shnatsel:
Also I will write a more clear policy for unmaintained crates when I get some time off work, we really do need more clarity around this.

saethlin:
> Combined with the apparent lack of high-profile users

This crate is still getting ~10,000 weekday downloads so I think it generally qualifies as impactful.

That said, this doesn't seem to definitively fall into the unmaintained category per the documentation you wrote up, so I'm closing this.

PR #1191, opened 2022-02-12
List `markdown` as unmaintained

dbanty:
The `markdown` crate is, naturally, the first one that comes up if you're searching crates.io for "markdown". Unfortunately, that particular crate has not received any updates since November of 2020 despite several known issues with open PRs. I opened https://github.com/johannhof/markdown.rs/issues/48 to request an update on maintenance status nearing a month ago and have not heard anything back.

I believe the best course is to list the crate as unmaintained so folks will (hopefully) correct course more quickly than I did 😅.

### DIFF
--- a/crates/markdown/RUSTSEC-0000-0000.md
+++ b/crates/markdown/RUSTSEC-0000-0000.md
@@ -1,0 +1,17 @@
+```toml
+ [advisory]
+ id = "RUSTSEC-0000-0000"
+ package = "markdown"
+ date = "2020-11-09"
+ informational = "unmaintained"
+ url = "https://github.com/johannhof/markdown.rs/issues/48"
+
+ [versions]
+ patched = []
+ unaffected = []
+ ```
+
+ # `markdown` is unmaintained
+
+ The [`markdown`](https://crates.io/crates/markdown) crate is no longer actively maintained. For Markdown parsing, you can use the [pulldown-cmark](https://crates.io/crates/pulldown-cmark) crate.
+ 
--- a/crates/markdown/RUSTSEC-0000-0000.md
+++ b/crates/markdown/RUSTSEC-0000-0000.md
@@ -1,17 +1,17 @@
 ```toml
- [advisory]
- id = "RUSTSEC-0000-0000"
- package = "markdown"
- date = "2020-11-09"
- informational = "unmaintained"
- url = "https://github.com/johannhof/markdown.rs/issues/48"
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "markdown"
+date = "2020-11-09"
+informational = "unmaintained"
+url = "https://github.com/johannhof/markdown.rs/issues/48"
 
- [versions]
- patched = []
- unaffected = []
- ```
+[versions]
+patched = []
+unaffected = []
+```
 
- # `markdown` is unmaintained
+# `markdown` is unmaintained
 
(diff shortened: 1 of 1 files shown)


pinkforest:
This gets ~ 150 downloads a day with some weird peak at Jul 8 - 12 with total 86k downloads for 0.3.0

I don't think there is anything inherently insecure about that this crate is unmaintained and it doesn't belong to any category that would have necessarily security impact - thus I would be hesitant to merge despite policy not being explicit about this

If you can come up with a reason why unmaintained makes this insecure somehow then that could be it.

The maintainer has not acknowledged the issue - have we tried e-mailing the author or @ mentioning as the author seems active ?

@amousset interested on second opinion re: category relevance with no security implication on unmaintained

dbanty:
> I don't think there is anything inherently insecure about that this crate is unmaintained and it doesn't belong to any category that would have necessarily security impact

For my own information, is there any documentation about what types of categories are considered at-risk? From my perspective—this crate was being used to parse user input. If there were a vulnerability in the crate, and the crate was being used in a web service (hypothetically, a Markdown -> HTML converter), then the vulnerability could cause real harm.

That being said I'm no security expert, and I have not audited the code, I just noticed it was not maintained and had open issues 😅.

pinkforest:
That was just a side-note as there was no concrete / suspect security issues in the Issues list -

But since our new policy is stricter due to supply chain concerns we will have easier time to merge these - 

In the past these tended to get stuck

Thanks for bringing it up!

pinkforest:
HI @dbanty I just realised that you did notify the maintainer on 17 January (I read June) and I apologise for that -

Could we please adjust the date to when it was originally reported e.g. 2022-01-17 and I think we can merge this ? Thanks

dbanty:
@pinkforest date fixed!

pinkforest:
Thanks :)

qtfkwk:
Is this still current? The repository is <https://github.com/wooorm/markdown-rs> now according to <https://crates.io/crates/markdown>. They haven't published a new stable version, but latest is 1.0.0-alpha.5.

pinkforest:
Looks like it has a new maintainer where another new crate took over the old name:

> Note: this is a new crate that reuses an old name. The old crate (0.3.0 and lower) has a bunch of problems. Make sure to use the new crate, currently in alpha at 1.0.0-alpha.5.

Even though crates would give you crate that isn't pre-release I think we could mark it as "withdrawn" on this per our format.

To me it looks like @wooorm is working hard to stabilise it at 1.0.0 and we could make a note of this too ?

tarcieri:
Yeah that's all good information to include in the advisory, even if it's withdrawn

wooorm:
Very maintained, very well tested!
There are some small things to do before I'd actually want to release v1. But I also would like some more people to try it out. A bit chicken and egg.
Is there a way to get crates and friends to show the alpha (especially maybe interesting because this is pre 1.0.0, when semver is all different from after 1.0.0)

pinkforest:
crates.io doesn't show pre-releases unless you click releases tab and Cargo requires exact version match to use it.

Also @wooorm would you mind sending us a PR to update this advisory in here:
https://github.com/rustsec/advisory-db/blob/main/crates/markdown/RUSTSEC-2022-0044.md

with the below added line in the metadata section:
`withdrawn = "2022-01-21"`

If you could please also add or copy note into the bottom markdown section there (into the advisory) as well regarding the status that I believe you already had the note about in your readme what happened - thanks.

Cheers

wooorm:
Done! No clue how you want me to word withdrawn deprecations, feel free to edit

PR #1192, opened 2022-02-12
HOWTO_UNMAINTAINED.md: guide for unmaintained crate advisories

tarcieri:
Initial guide on policy around what RustSec considers to be an unmaintained crate as well as the policy for filing an advisory.

This has come up / been requested often enough I thought I'd finally take a go at writing something up.

[Rendered](https://github.com/rustsec/advisory-db/blob/024202565812a765d3b02f32f74c08475689ac14/HOWTO_UNMAINTAINED.md)

### DIFF
--- a/HOWTO_UNMAINTAINED.md
+++ b/HOWTO_UNMAINTAINED.md
@@ -30,7 +30,7 @@ project's source code repository. If an author/maintainer of a particular crate
 thinks filing an unmaintained crate advisory is a good idea, then great! Go ahead.
 
 If the author is responsive and declares that the project *is* maintained, then
-the from the perspective of RustSec is maintained. Again, when in doubt, defer
+the from the perspective of RustSec the crate is maintained. Again, when in doubt, defer
 to the author's discretion. So long as the author is responsive and avows that
 a crate is maintained, we take them at their word. Repository metrics like
 recent commits, open issues, latest release, etc are not reasons to go against
--- a/HOWTO_UNMAINTAINED.md
+++ b/HOWTO_UNMAINTAINED.md
@@ -30,7 +30,7 @@ project's source code repository. If an author/maintainer of a particular crate
 thinks filing an unmaintained crate advisory is a good idea, then great! Go ahead.
 
 If the author is responsive and declares that the project *is* maintained, then
-the from the perspective of RustSec the crate is maintained. Again, when in doubt, defer
+the RustSec organization considers it maintained and won't accept advisories about its maintenance status. Again, when in doubt, defer
 to the author's discretion. So long as the author is responsive and avows that
 a crate is maintained, we take them at their word. Repository metrics like
 recent commits, open issues, latest release, etc are not reasons to go against
--- a/HOWTO_UNMAINTAINED.md
+++ b/HOWTO_UNMAINTAINED.md
@@ -27,7 +27,7 @@ we are able to make contact.
 
 First and foremost: *ASK THE AUTHOR(S)*, preferably in a public issue on the
 project's source code repository. If an author/maintainer of a particular crate
-thinks filing an unmaintained crate advisory is a good idea, then great! Go ahead
+thinks filing an unmaintained crate advisory is a good idea, then great! Go ahead.
 
 If the author is responsive and declares that the project *is* maintained, then
 the from the perspective of RustSec is maintained. Again, when in doubt, defer
--- a/HOWTO_UNMAINTAINED.md
+++ b/HOWTO_UNMAINTAINED.md
@@ -1,0 +1,80 @@
+# HOWTO Guide: Unmaintained Crate Advisories
+
+This document describes the policy for adding advisories for unmaintained
+crates to the [RustSec Advisory Database].
+
+These advisories serve to inform the Rust community about both the existence
+of unmaintained crates within a particular project, and also serve to guide
+switching to maintained alternatives.
+
+## Definition of an "unmaintained" crate
+
+We consider crates unmaintained when they fall into either of the following
+categories:
+
+- Explicitly unmaintained: a crate's author has declared that they are no
+  longer maintaining a particular crate.
+- Implicitly unmaintained: the author is incommunicado for a prolonged period
+  of time and cannot advise as to a crate's status.
+
+## Creating an unmaintained crate advisory
+
+### Policy
+
+When in doubt, we always defer to the author of a crate's discretion as to
+whether they would prefer an unmaintained crate advisory be filed, provided
+we are able to make contact.
+
+First and foremost: *ASK THE AUTHOR(S)*, preferably in a public issue on the
+project's source code repository. If an author/maintainer of a particular crate
+thinks filing an unmaintained crate advisory is a good idea, then great! Go ahead
+
+If the author is responsive and declares that the project *is* maintained, then
+the from the perspective of RustSec is maintained. Again, when in doubt, defer
+to the author's discretion. So long as the author is responsive and avows that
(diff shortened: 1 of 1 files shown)


HOWTO_UNMAINTAINED.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Policy
+
+When in doubt, we always defer to the author of a crate's discretion as to
+whether they would prefer an unmaintained crate advisory be filed, provided
+we are able to make contact.
+
+First and foremost: *ASK THE AUTHOR(S)*, preferably in a public issue on the
+project's source code repository. If an author/maintainer of a particular crate
+thinks filing an unmaintained crate advisory is a good idea, then great! Go ahead
```
BlackHoleFox, line 30:
```suggestion
thinks filing an unmaintained crate advisory is a good idea, then great! Go ahead.
```

HOWTO_UNMAINTAINED.md
```diff
@@ -0,0 +1,80 @@
...
+
+If the author is responsive and declares that the project *is* maintained, then
+the from the perspective of RustSec is maintained. Again, when in doubt, defer
```
BlackHoleFox, line 33: Chopped sentence?

```suggestion
the from the perspective of RustSec the crate is maintained. Again, when in doubt, defer
```

BlackHoleFox, line 33: As an alternate wording that reads over a bit smoother:
```suggestion
If the author is responsive and declares that the project *is* maintained, then
the RustSec organization considers it maintained and won't accept advisories about its maintenance status. Again, when in doubt, defer
```

HOWTO_UNMAINTAINED.md
```diff
@@ -0,0 +1,80 @@
+# HOWTO Guide: Unmaintained Crate Advisories
+
+This document describes the policy for adding advisories for unmaintained
+crates to the [RustSec Advisory Database].
+
+These advisories serve to inform the Rust community about both the existence
+of unmaintained crates within a particular project, and also serve to guide
+switching to maintained alternatives.
```
BlackHoleFox, line 8: I've seen this as a pretty big influence when going through the registry for an alternative, so its nice to see called out here.

Should this guide have a small section with a few tips about setting up the `alternatives` field in the advisory and any advice picking one (ie don't put things with almost no downloads, outstanding soundness issues, etc)?

tarcieri: Yes! I strongly agree there should be more information about trying to collect recommended alternatives to an unmaintained crate in order to keep the advisories actionable.

HOWTO_UNMAINTAINED.md
tarcieri: I like your suggested wording but the suggestion has become unmergeable from the GitHub UI.

HOWTO_UNMAINTAINED.md
```diff
@@ -0,0 +1,80 @@
...
+When in doubt, we always defer to the author of a crate's discretion as to
+whether they would prefer an unmaintained crate advisory be filed, provided
+we are able to make contact.
+
+First and foremost: *ASK THE AUTHOR(S)*, preferably in a public issue on the
+project's source code repository. If an author/maintainer of a particular crate
+thinks filing an unmaintained crate advisory is a good idea, then great! Go ahead.
+
+If the author is responsive and declares that the project *is* maintained, then
+the from the perspective of RustSec the crate is maintained. Again, when in doubt, defer
```
BlackHoleFox, line 33: Duplicated from [here](https://github.com/rustsec/advisory-db/pull/1192#discussion_r805103804) since GitHub's UI died.

```suggestion
If the author is responsive and declares that the project *is* maintained, then
the RustSec organization considers it maintained and won't accept advisories about its maintenance status. Again, when in doubt, defer
```

HOWTO_UNMAINTAINED.md
BlackHoleFox: I was wondering how it'd behave with two changes on the same spot, guess that's answered. I copy/pasted the suggestion into https://github.com/rustsec/advisory-db/pull/1192#discussion_r805106812

HOWTO_UNMAINTAINED.md
BlackHoleFox: Here's my rough barrel of them and why:
- \~relatively\~ widely-used. The crate should have an all-time download count >= 1,000. The goal isn't to eliminate unpopular options, but to recommend something that more than a few people have put trust into.
- No outstanding soundness or security bugs: This would just lead from fixing one advisory to a different, worse one.
- Active maintenance: The inverse of the de-facto unmaintained guidelines here should apply to the repository/crate/etc
- Relevant: The proposed alternate crate should offer APIs that do the same thing and have (ideally...) minimal changes required to migrate over. Someone shouldn't need to re-build their project around an API difference. I admit that this one is pretty shaky, but maybe it could be worded in a nudging style instead of a requirement?

HOWTO_UNMAINTAINED.md
tarcieri: Cool, thanks

HOWTO_UNMAINTAINED.md
tarcieri: Those sound like fairly reasonable guidelines

HOWTO_UNMAINTAINED.md
```diff
@@ -0,0 +1,80 @@
...
+These advisories serve to inform the Rust community about both the existence
+of unmaintained crates within a particular project, and also serve to guide
+switching to maintained alternatives.
+
+## Definition of an "unmaintained" crate
+
+We consider crates unmaintained when they fall into either of the following
+categories:
+
+- Explicitly unmaintained: a crate's author has declared that they are no
```
5225225, line 15: Is a repo being marked as archived explicitly or implicitly unmaintaining it (assuming there *is* a linked repo to the crate, and there's no indication that it's not the primary one)?

tarcieri: An archived repo is certainly a signal that a crate is unmaintained, however we'd still ask the author for clarity in that case unless they have left an explicit commit message to that effect

HOWTO_UNMAINTAINED.md
```diff
@@ -0,0 +1,80 @@
...
+
+- Stale repository: no recent maintenance activity, including any of the
+  following: recent commits, responses from the author on open issues,
+  crate releases, or other publically visible activity by the author.
+  Inactivity over a period of 1 year or more is the preferred threshold.
+- Contact attempts with the author made with no response. Ideally these
+  attempts are made via a public GitHub issue, so that issue can be
+  cited in an unmaintained crate advisory if need be. Unresponsiveness
+  by the author over a period of 90 days is suggested before filing an
+  advisory.
```
djc, line 56: I'd suggest at least also trying to send an email outside of GitHub directly to any email addresses from maintainers that can be harvested (for example, from GitHub commits). I've found several times that maintainers who were hard to reach via GitHub were much more responsive via email.

chris-morgan:
This is procedure for when a crate becomes unmaintained; but what about when a crate becomes maintained again? Is there procedure for that, some way of rescinding or countermanding an advisory? Feels like it’d at least warrant a mention here, as this should be a type of advisory particularly prone to reversion.

tarcieri:
Went ahead and merged this so we at least have something to work with.

Please submit PRs with specific changes you'd like to see in the wording.

tarcieri:
@chris-morgan to answer your question, we can yank/unyank advisories by adding the `yanked = true` attribute to the advisory metadata

PR #1197, opened 2022-02-17
Add unsoundness advisory for enum-map

xfix:
This was caused by a mistake during refactoring: https://gitlab.com/KonradBorowski/enum-map/-/commit/ec9eecd63114a40bfc2505cbb2a883dd0ef89d3f. Versions 2.0.0-2, 2.0.0 and 2.0.1 are affected.

### DIFF
--- a/crates/enum-map/RUSTSEC-0000-0000.md
+++ b/crates/enum-map/RUSTSEC-0000-0000.md
@@ -1,0 +1,56 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "enum-map"
+date = "2022-02-17"
+url = "https://gitlab.com/KonradBorowski/enum-map/-/blob/master/CHANGELOG.md#version-202"
+categories = ["code-execution", "memory-corruption", "memory-exposure"]
+informational = "unsound"
+
+[versions]
+patched = [">= 2.0.2"]
+unaffected = ["< 2.0.0-2"]
+```
+
+# enum_map macro can cause UB when `Enum` trait is incorrectly implemented
+
+Affected versions of this crate did not properly check the length of an enum when using `enum_map!` macro, trusting user-provided length.
+
+When the `LENGTH` in the `Enum` trait does not match the array length in the `EnumArray` trait, this can result in the initialization of the enum map with uninitialized types, which in turn can allow an attacker to execute arbitrary code.
+
+This problem can only occur with a manual implementation of the Enum trait, it will never occur for enums that use `#[derive(Enum)]`.
+
+Example code that triggers this vulnerability looks like this:
+
+```rust
+enum E {
+    A,
+    B,
+    C,
+}
+
+impl Enum for E {
+    const LENGTH: usize = 2;
+
+    fn from_usize(value: usize) -> E {
+        match value {
+            0 => E::A,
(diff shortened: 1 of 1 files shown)


crates/enum-map/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
...
+id = "RUSTSEC-0000-0000"
+package = "enum-map"
+date = "2022-02-17"
+url = "https://gitlab.com/KonradBorowski/enum-map/-/blob/master/CHANGELOG.md#version-202"
+categories = ["memory-corruption", "memory-exposure"]
+informational = "unsound"
+
+[versions]
+patched = [">= 2.0.2"]
+unaffected = ["< 2.0.0"]
```
Shnatsel, line 12:
```suggestion
unaffected = ["< 2.0.0-2"]
```

`2.0.0-2` is a pre-release under the semver rules, so it goes before 2.0.0

Shnatsel (commented):
Thanks for the report! I've proposed an edit for the version range, does that look good to you?

crates/enum-map/RUSTSEC-0000-0000.md
xfix: Fair, I had no idea whether this syntax would work. Applied the change.

xfix:
Removed `memory-corruption` from categories, because this doesn't actually allow to perform out of bounds writes due to bound checks while writing to an enum map with a macro.

crates/enum-map/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
...
+informational = "unsound"
+
+[versions]
+patched = [">= 2.0.2"]
+unaffected = ["< 2.0.0-2"]
+```
+
+# enum_map macro can cause UB when `Enum` trait is incorrectly implemented
+
+Affected versions of this crate did not properly length of an enum when using `enum_map!` macro trusting user-provided length.
```
Shnatsel, line 17:
```suggestion
Affected versions of this crate did not properly check the length of an enum when using `enum_map!` macro, trusting user-provided length.
```

crates/enum-map/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
...
+
+This can result in a memory corruption when `LENGTH` in `Enum` trait doesn't match the array in `EnumArray` trait.
```
Shnatsel, line 19: I understand that it may result in reading the contents of uninitialized memory if the actual array length is larger than the one specified? Or is it the other way around?

This is important because this defines what kind of exploits can be pulled off here. For example, if the enum contains a pointer, and can be constructed from uninit memory, then that is exploitable for arbitrary code execution.

xfix: Older versions of `enum-map` could be tricked into not initializing array elements. It's as if you created `[MaybeUninit::uninit(), ...]` of whatever type you want.

xfix: Added more details to an advisory and updated categories.

Shnatsel:
Looks very good now. Merging. Thanks a lot!

PR #1217, opened 2022-03-31
Add CVE-2022-24791 for Wasmtime

fitzgen:


crates/wasmtime/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
...
+id = "RUSTSEC-0000-0000"
+package = "wasmtime"
+date = "2022-03-31"
+url = "https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-gwc9-348x-qwv2"
+categories = ["memory-corruption", "memory-exposure"]
+keywords = ["use-after-free", "Wasm", "garbage collection"]
+aliases = ["CVE-2022-24791"]
+
+[versions]
+patched = [">= 0.35.2", ">= 0.34.2"]
```
alex, line 12: This should be `>= 0.34.2, < 0.35` right?

fitzgen: I guess? Not totally sure what the semantics of semver relational operators are and whether they are a total ordering or a partial ordering

alex: Presumabely 0.35.1 is impacted, but the `>= 0.34.2` would indicate its patched.

fitzgen: Should be fixed now

PR #1219, opened 2022-04-04
Report lifetime bug in get_or_insert method in lru crate

jeromefroe:


tarcieri:
Why did you close this?

PR #1220, opened 2022-04-05
RUSTSEC-2022-0012: note that v0.10.0+ is patched

benesch:
@jorgecarleitao can you confirm this is true?

----

As far as I can tell, v0.10.0+ was not affected by this bug [0]. The commit which
fixes the unsoundness landed in main before v0.10.0 was released.

[0]: https://github.com/jorgecarleitao/arrow2/commit/9d4342c5ff2ff1593232373a9998c3da18c7854d


jorgecarleitao:
Thanks! Yes, I confirm; I filled the issue incorrectly :/

Shnatsel:
Thanks!

PR #1224, opened 2022-04-27
Add advisory for using impure constants in array-macro

xfix:


Shnatsel:
Do I understand correctly that the value has to be passed at compile time, and is not controllable by the attacker?

I agree with the `informational = unsound` in that case, but I think it could be called out in the advisory explicitly.

Otherwise looks good, thanks! :+1:

xfix:
> Do I understand correctly that the value has to be passed at compile time, and is not controllable by the attacker?

That's correct, yes.

PR #1228, opened 2022-05-03
sass-rs is deprecated

CAD97:
Issue: https://github.com/compass-rs/sass-rs/issues/83

The deprecation/unmaintained status is also now noted in the README: https://github.com/compass-rs/sass-rs

> **This crate is not maintained anymore as libsass is deprecated. Consider using <https://github.com/connorskees/grass> or <https://github.com/kaj/rsass> instead.**


pinkforest (approved):
LGTM from me :) thanks!

PR #1231, opened 2022-05-10
add crossbeam advisories for incorrect (unsound) zeroed memory

RalfJung:
This issue is ancient and has been fixed long ago, but it showed up in a [crater run](https://crater-reports.s3.amazonaws.com/pr-87041-1/try%23ddc023471205c186b40f4bf8ffb598008dbf4885/reg/advantage-0.1.0/log.txt) not too long ago so it might still be worth adding this to the database (considering that there are no other advisories against crossbeam-queue that might urge people to update).

Cc https://github.com/crossbeam-rs/crossbeam/pull/458

### DIFF
--- a/crates/crossbeam-channel/RUSTSEC-0000-0000.md
+++ b/crates/crossbeam-channel/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "crossbeam-channel"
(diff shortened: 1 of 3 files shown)


RalfJung:
Looks like up until crossbeam 0.6.0, the queue was [in the main crossbeam package](https://docs.rs/crossbeam/0.6.0/src/crossbeam/seg_queue.rs.html), so I also added an advisory for that.

PR #1232, opened 2022-05-10
add hyper uninitialized memory advisory

RalfJung:
Adds an advisory for the fix in https://github.com/hyperium/hyper/pull/2545.

### DIFF
--- a/crates/hyper/RUSTSEC-0000-0000.md
+++ b/crates/hyper/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "hyper"
+date = "2022-05-10"
+informational = "unsound"
+url = "https://github.com/hyperium/hyper/pull/2545"
+
+[versions]
+patched = [">= 0.4.12"]
+```
+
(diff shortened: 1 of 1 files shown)


RalfJung:
I am pretty sure the crate never dereferences that uninit pointer. It was unused. But it had a type which indicated it should be non-null. (In theory the compiler might have introduced spurious loads from that pointer, but we have no evidence of that actually happening.)

Shnatsel:
Alright. Since it's an old issue and is only `informational = unsound`, i.e. a warning, I'll go ahead and merge it without maintainer sign-off.

Shnatsel:
Thanks a lot for reporting!

seanmonstar:
I wouldn't call it a vulnerability myself. The compiler *could* have done a thing, but it never did. I don't see any reason to warn people, warnings can fatigue.

RalfJung:
> The compiler could have done a thing, but it never did.

That's exactly what the "unsound" category is for, I think.
Quite a few crates came up still depending on old hyper in the crater run https://github.com/rust-lang/rust/pull/87041 so I feel nudging the ecosystem a bit more might be a good idea...

PR #1234, opened 2022-05-10
Security advisory: malicious crate rustdecimal

Shnatsel:


Shnatsel:
Ah, CI fails because this package has been permanently removed, so there's no match on crates.io for `rustdecimal`

Shnatsel:
I've tried simply squatting it, but the name has been marked as reserved, so I cannot do that either.

tarcieri:
@Shnatsel we'd need to add a lint skipping option to `rustsec-admin`, which wouldn't be too hard

pinkforest:
I just added a PR for the tooling:
https://github.com/rustsec/rustsec/pull/620

Design:
https://github.com/rustsec/rustsec/issues/619

Shnatsel:
Superseded by https://github.com/rustsec/advisory-db/pull/1312

PR #1236, opened 2022-05-11
Add unmaintained crate advisory for `static_type_map`

malobre:
It has been renamed to `erased_set`.

See latest [readme on crates.io](https://crates.io/crates/static_type_map) and https://github.com/malobre/erased_set/issues/6.


Shnatsel:
Thanks!

Issue #1190, opened 2022-02-11
Clarify advisories about crates that read from (not write to) the environment


briansmith:
There are some advisories like https://rustsec.org/advisories/RUSTSEC-2020-0159 and https://rustsec.org/advisories/RUSTSEC-2020-0071 that have spawned good discussion and are (probably) leading to good changes to the standard library. However, these advisories also seem to be at odds with the current thinking, which is that it is OK to read from the environment but not write to the environment. 

I think these advisories are now distracting from real soundness (memory safety and other race condition) issues like https://github.com/chronotope/chrono/issues/648 that are the result of writing to the environment.

We're in a weird state where we now have issues like #926 that stay open here because we can't agree on whether reading from the environment constitutes a vulnerability. I don't think we can really consider it to be a vulnerability at this point, so I think those advisories should be closed with such an explanation. Are we waiting for the Rust library team to agree and to do the deprecation of `env::{set_var,remove_var}`?

briansmith:
Oh, sorry, I forgot the main request I had: We should update the text of the RUSTSEC advisories and the related CVE with more explanation of the issue, as the current advisory text, though not totally wrong, not really helpful in helping people understand the issue.

Otherwise, presumably there are many, many RUSTSEC advisories to file against many, many crates, for the same reason.

tarcieri:
Yeah, this was something we've waffled on a few times as the individual advisories were part of a much bigger systemic issue that goes all the way down to the POSIX API. We had previously rejected some advisories for this and ended up accepting the ones for `time` and `chrono`, perhaps in haste.

In the time since I think there's been a lot of discussion, perhaps in part motivated by the amount of noise this created, and perhaps a clearer path to a solution.

I think at the very least these advisories aren't security critical and can thus be demoted to `informational = "unsound"` which in turn demotes them to warnings in `cargo audit`.

We could also consider yanking them, especially if `env::set_var` is deprecated.

pinkforest:
Sum summarum: 

There are upstream rust-lang/rust issue(s) https://github.com/rust-lang/rust/issues/90308 and libs are well aware of this by now with extensive discussions.

time was only re-filed because the maintainer wanted it back - https://github.com/rustsec/advisory-db/pull/1242#issuecomment-1144903688

chrono was only re-filed to advertise fix because there was RiR localtime_r in 0.4.20 - https://github.com/rustsec/advisory-db/pull/1310

[...]

PR #1241, opened 2022-05-12
Withdraw RUSTSEC-2020-0159: unsound `localtime_r` call in `chrono`

tarcieri:
Per rustsec/advisory-db#1190, it would be good to move to a policy where we don't file advisories against crates which perform unsynchronized reads from the process environment, and instead focus only on crates which modify the process environment in an unsynchronized manner.

### DIFF
--- a/crates/chrono/RUSTSEC-2020-0159.md
+++ b/crates/chrono/RUSTSEC-2020-0159.md
@@ -7,6 +7,8 @@ url = "https://github.com/chronotope/chrono/issues/499"
 categories = ["code-execution", "memory-corruption"]
 keywords = ["segfault"]
 related = ["CVE-2020-26235", "RUSTSEC-2020-0071"]
+withdrawn = "2022-05-12" # see rustsec/advisory-db#1190
+yanked = true
 
 [versions]
 patched = []


crates/chrono/RUSTSEC-2020-0159.md
```diff
@@ -7,6 +7,8 @@ url = "https://github.com/chronotope/chrono/issues/499"
 categories = ["code-execution", "memory-corruption"]
 keywords = ["segfault"]
 related = ["CVE-2020-26235", "RUSTSEC-2020-0071"]
+withdrawn = "2022-05-12" # see rustsec/advisory-db#1190
```
Shnatsel, line 10: Maybe we should move the rationale for yanking to the Markdown part, so that it shows up in exported data as well (OSV, GHSA etc)?

tarcieri: That'd be a good idea, but it seems this feature is still halfway migrated from a boolean `yanked = true` flag to `withdrawn = ` with the date.

So we'd need to get support into the next `rustsec` crate release, wait for everyone to upgrade, then start using it after the next release.

Maybe open a tracking ticket for this on https://github.com/rustsec/rustsec ?

Shnatsel: I meant the comment: `# see rustsec/advisory-db#1190`

Also, I thought I've migrated the `rustsec` crate to dates, and left the `yanked = true` in the advisories for backwards compatibility... if that's not the case, I should add that to my TODO.

tarcieri: Yeah, I understand what you're saying. I'm just saying this field is already in flux, and we'd need to add another forward-compatible change to the code first before we could modify the database.

PR #1256, opened 2022-05-23
Use after free in Neon externally allocated JavaScript buffers

kjvalencik:
In versions prior to Neon `0.10.1`, users could create a JavaScript buffer backed by bytes owned by Rust *without* the `'static` lifetime. The following example demonstrates how this could trivially result in a use-after-free bug.

```rust
pub fn soundness_hole(mut cx: FunctionContext) -> JsResult<JsArrayBuffer> {
    let mut data = vec![0u8, 1, 2, 3];
    
    // Creating an external from `&mut [u8]` instead of `Vec<u8>` since there is a blanket impl
    // of `AsMut<T> for &mut T`
    let buf = JsArrayBuffer::external(&mut cx, data.as_mut_slice());
    // `buf` is still holding a reference to `data`!
    drop(data);
    Ok(buf)
}
```

### DIFF
--- a/crates/neon/RUSTSEC-0000-0000.md
+++ b/crates/neon/RUSTSEC-0000-0000.md
@@ -1,0 +1,38 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "neon"
+date = "2022-05-22"
+url = "https://github.com/neon-bindings/neon/issues/896"
+categories = ["memory-corruption", "memory-exposure"]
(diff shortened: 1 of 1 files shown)


alex (approved):

Shnatsel:
Thanks for the detailed advisory!

kjvalencik:
👍  For any Neon users who happen to come across this issue. If your code continues to compile after updating the Neon version, it means your code was *not* impacted by the soundness issue.

Thanks @Shnatsel!

PR #1263, opened 2022-06-17
Fix RUSTSEC-2022-0025,26,27 openssl-src for the 111 stream

### LINKED ISSUE #1262: openssl-src misses 1.1.1 users on RUSTSEC-2022-00[25,26,27]
Dependabot is now pinging patched 1.1.1 users about outdated openssl-src and asking to switch to 300 stream

Previous PR #1243 #1244 #1245 #1246 #1247 #1248 #1249 

I wonder if dependabot follows the unaffected field how rustadvisory has specified it

This is a bit annoying since vendored openssl pulls 1.1.1o - via 111.20.0 - which is patched

**However .. to address the three advisories for 1.1.1 stream users -**

I've sent PR to to mark 111.20.0 patched following earlier syntax
 #1263 

https://www.openssl.org/news/secadv/20220503.txt
OpenSSL 1.0.2 users should upgrade to 1.0.2ze (premium support customers only)
OpenSSL 1.1.1 users should upgrade to 1.1.1o
OpenSSL 3.0 users should upgrade to 3.0.3

This will currently show up as an issue in dependabot as e.g. vendored openssl pulls 111.20.0+1.1.1o

pinkforest:
Closes #1262 

On 2022-0025, 26 and 27 111 is affected and is reflected on openssl-src 111.20.0 that brings 1.1.1o (patched)

I folllowed RUSTSEC-2022-0014.md way to give effect to this

### DIFF
--- a/crates/openssl-src/RUSTSEC-2022-0025.md
+++ b/crates/openssl-src/RUSTSEC-2022-0025.md
@@ -9,8 +9,7 @@ date = "2022-05-03"
 url = "https://www.openssl.org/news/secadv/20220503.txt"
 
 [versions]
-patched = [">= 300.0.6"]
-unaffected = ["< 300.0"]
+patched = [">= 300.0.6", ">= 111.20.0"]
 ```
 
 # Resource leakage when decoding certificates and keys
--- a/crates/openssl-src/RUSTSEC-2022-0025.md
+++ b/crates/openssl-src/RUSTSEC-2022-0025.md
@@ -9,7 +9,8 @@ date = "2022-05-03"
 url = "https://www.openssl.org/news/secadv/20220503.txt"
 
 [versions]
-patched = [">= 300.0.6", ">= 111.20.0"]
+patched = [">= 300.0.6"]
+unaffected = ["< 300.0.0"]
 ```
 
 # Resource leakage when decoding certificates and keys
--- a/crates/openssl-src/RUSTSEC-2022-0025.md
+++ b/crates/openssl-src/RUSTSEC-2022-0025.md
@@ -9,8 +9,7 @@ date = "2022-05-03"
 url = "https://www.openssl.org/news/secadv/20220503.txt"
 
 [versions]
-patched = [">= 300.0.6"]
-unaffected = ["< 300.0.0"]
+patched = [">= 111.20.0, < 300.0.0", ">= 300.0.6"]
 ```
 
 # Resource leakage when decoding certificates and keys
--- a/crates/openssl-src/RUSTSEC-2022-0026.md
+++ b/crates/openssl-src/RUSTSEC-2022-0026.md
@@ -9,7 +9,8 @@ date = "2022-05-03"
 url = "https://www.openssl.org/news/secadv/20220503.txt"
 
 [versions]
-patched = [">= 300.0.6", ">= 111.20.0"]
+patched = [">= 300.0.6"]
+unaffected = ["< 300.0.0"]
 ```
 
 # Incorrect MAC key used in the RC4-MD5 ciphersuite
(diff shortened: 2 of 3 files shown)


pinkforest:
It looks like I cannot address both 300 and 111 streams as this does not work in semver basis

Shnatsel:
The issue affects only dependabot and not `cargo audit`, is that correct?

I'm afraid that applying this workaround will not actually fix the issue, at least not anytime soon, since GHSA doesn't sync RustSec advisories on a regular basis - it seems more like they've run a one-off import.

If the issue is specific to Dependabot and does not manifest in `cargo audit`, you should submit a PR against GHSA instead: https://github.com/github/advisory-database

pinkforest:
It's not just a dependabot. People get the wrong idea that switching to 300 solves the problem.

This isn't fixing dependabot but follows the previous pattern to indicate that 111 series has a fixed version - that should have been in the original advisory.

Shnatsel:
I see. Merging, then. Thanks a lot for the fix!

pinkforest:
Thx! - Apologies about the confusion I should have reframed the issue properly I should have treated dependabot issue separately -  :unicorn:

pinkforest:
I'll also chase the GHSA on this to see if I can get them to sync and fix

amousset:
I created these advisories, and I'm sorry if I missed affected versions. However I check the [openssl advisory](https://www.openssl.org/news/secadv/20220503.txt) again and it still appears to me that the 1.1.1 branch was not affected. See also the [3.0.3 changelog](https://github.com/openssl/openssl/blob/master/CHANGES.md#changes-between-302-and-303) vs. [1.1.1o changelog](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/CHANGES#L18) 

#### CVE-2022-1473 - RUSTSEC-2022-0025

> It was addressed in the 3.0.3 release on the 3rd May 2022. The fix can be
> found in git commit 64c85430f.
>
> OpenSSL 1.0.2 users are not affected.
> OpenSSL 1.1.1 users are not affected.
> OpenSSL 3.0 users should upgrade to 3.0.3.

#### CVE-2022-1434 - RUSTSEC-2022-0026

> The OpenSSL 3.0 implementation of the RC4-MD5 ciphersuite incorrectly uses the
> AAD data as the MAC key. This makes the MAC key trivially predictable.
> [...]
> This issue affects OpenSSL version 3.0.
>
> OpenSSL 3.0 users should upgrade to 3.0.3

#### CVE-2022-1343 - RUSTSEC-2022-0027

> This issue affects OpenSSL version 3.0.
>
> OpenSSL 3.0 users should upgrade to 3.0.3

Shnatsel:
@amousset thank you for double-checking. Upon further inspection it appears that you are correct. To double-check, reverting this PR restores the correct version ranges, is that right?

amousset:
I think so (but I did not fully understand the problem with dependabot).

I can also file an advisory for CVE-2022-1292 (_The c_rehash script allows command injection_) that affects 1.1.1o and 3.0.3, and was the cause of the confusion in https://github.com/rustsec/advisory-db/issues/1262. I initially skipped it as it only affected the `c_rehash` perl script and not the C library so it shouldn't affect Rust programs only linking against openssl, but as the script is still part of the source crate I guess it could also have an advisory.

pinkforest:
I can file the issue for 1292..  yes it was 1292 which indeed means we need to ugprade to it

This issue affects OpenSSL versions 1.0.2, 1.1.1 and 3.0.

OpenSSL 1.0.2 users should upgrade to 1.0.2ze (premium support customers only)
OpenSSL 1.1.1 users should upgrade to 1.1.1o
OpenSSL 3.0 users should upgrade to 3.0.3

amousset:
> Is the Perl script used during the build via openssl-src ? I don't simply remember - but even in that case probably uses it in very specific way controlled by the -src crate.

The script is not used during openssl-src build nor by openssl-sys

PR #1264, opened 2022-06-19
Revert "Fix RUSTSEC-2022-0025,26,27 openssl-src for the 111 stream"

Shnatsel:
Reverts rustsec/advisory-db#1263, see that thread for discussion

cc @pinkforest @amousset


pinkforest:
@Shnatsel good to go - thanks for fixing.

PR #1269, opened 2022-06-25
Copyedit RUSTSEC-2021-0122

8573:
Apply some copyedits to the text introduced by #1268.  Add a missing comma; inflect "be" better; and use normal quotation marks rather than backticks around the word "safe", referring to Safe Rust, as, unlike `unsafe`, "safe" is not Rust syntax.

I normally would use a more descriptive branch name than "patch-1", but I'm doing this on a phone and I didn't see an option to change the branch name.


Shnatsel:
Thanks!

alamb:
For what it is worth, https://github.com/google/flatbuffers/pull/7518 has just been merged. Not sure about the release schedule just yet but it is getting close

PR #1271, opened 2022-06-26
rulex advisory for stack overflow

evanrichter:


Shnatsel:
Thank you for the report!

I think we'll need to split this into two PRs, since the ID assignment script looks for the file named `RUSTSEC-0000-0000.md` specifically.

Also, we'll need one of the rulex maintainers (I assume @Aloso ) to confirm they are OK with releasing this information into the public domain, since Github information is under Creative Commons Attribution license.

Aloso:
@Shnatsel I am ok with this.

evanrichter:
ok, I'll split the PRs :)

evanrichter:
this one is now only for the stack overflow, and the other is in #1272

Shnatsel:
Merging. Thanks!

PR #1272, opened 2022-06-26
rulex advisory for string indexing panic

evanrichter:


Shnatsel:
The authors agreed to release the advisory information into the public domain in https://github.com/rustsec/advisory-db/pull/1271#issuecomment-1166607902

PR #1275, opened 2022-07-02
Add unmaintained advisory for term_size

LingMan:
Note that I've used the date when the unmaintained status was announced in the project's README for the date field.

### DIFF
--- a/crates/term_size/RUSTSEC-0000-0000.md
+++ b/crates/term_size/RUSTSEC-0000-0000.md
@@ -1,0 +1,16 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "term_size"
+date = "2020-11-03"
+informational = "unmaintained"
+url = "https://github.com/clap-rs/term_size-rs/pull/31"
+[versions]
+patched = []
+unaffected = []
+```
+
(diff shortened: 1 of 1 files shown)


crates/term_size/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,18 @@
...
+url = "https://github.com/clap-rs/term_size-rs/pull/31"
+[versions]
+patched = []
+unaffected = []
+```
+
+# `term_size` is unmaintained; use `terminal_size` instead
+
+The [`term_size`](https://crates.io/crates/term_size) crate is no longer maintained. It works fine,
+but the maintainers don't have time and willingness to work on it or review future PRs anymore.
```
sunfishcode, line 16: It turns out it doesn't work fine on all platorms; as reported [here](https://github.com/ogham/exa/issues/400), term_size has the wrong value for `TIOCGWINSZ` on at least PowerPC.

LingMan: Ah, good catch. Dropped this sentence entirely.

pinkforest (approved):
Thanks :)

PR #1276, opened 2022-07-05
Add advisory for openssl CVE-2022-2274

amousset:


crates/openssl-src/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,30 @@
...
+id = "RUSTSEC-0000-0000"
+package = "openssl-src"
+aliases = ["CVE-2022-2274"]
+categories = ["crypto-failure"]
+date = "2022-07-05"
+url = "https://www.openssl.org/news/secadv/20220705.txt"
+
+[versions]
+patched = [">= 300.0.9"]
+unaffected = ["< 300.0.8"]
```
Shnatsel, line 12: The text below states 3.0.4 but here we have 300.0.8 - just wanted to double-check that this is correct

amousset: Yes, 300.0.8 is actually 3.0.4 (https://crates.io/crates/openssl-src/300.0.8+3.0.4) and 300.0.7 was 3.0.3 (https://crates.io/crates/openssl-src/300.0.7+3.0.3)

Shnatsel: That's... a little confusing, but alright! Thanks!

PR #1277, opened 2022-07-05
Add advisory for openssl CVE-2022-2097

amousset:


pinkforest:
didn't take long for it to catch an advisory... good work @amousset

PR #1284, opened 2022-07-28
Add unmaintained advisory for audiotags

Serial-ATA:
The repo has sat dormant for nearly two years. I spoke with the author over email, and they no longer have time to work on the crate. They have since archived it.

### DIFF
--- a/crates/audiotags/RUSTSEC-0000-0000.md
+++ b/crates/audiotags/RUSTSEC-0000-0000.md
@@ -1,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "audiotags"
+date = "2022-07-28"
+informational = "unmaintained"
+url = "https://github.com/TianyiShi2001/audiotags"
+
+[versions]
+patched = []
+```
+
+# `audiotags` is unmaintained
+
+The author no longer has time to maintain the crate,
+and the repository has been archived. The author
+recommends switching to [`lofty`](https://crates.io/crates/lofty).


amousset:
Thanks. [audiotags2](https://crates.io/crates/audiotags2) seems to be a maintained fork of audiotags, which may be easier to switch to so it could be worth mentioning. However its actual maintenance state seems unsure (https://github.com/martpie/rust-audiotags2/issues/8).

Shnatsel:
We generally reserve unmaintained status for crates the maintainers of which are completely unresponsive. Since the author did reply, perhaps they could hand over the crates.io upload rights to a new maintainer, or at least add the https://github.com/rust-bus organization as an owner so they could transfer the crate later if someone wants to take over maintenance?

tarcieri:
A crate explicitly declared as unmaintained by its author falls under the ["explicitly unmaintained" category descxribed in HOWTO_UNMAINTAINED.md](https://github.com/rustsec/advisory-db/blob/main/HOWTO_UNMAINTAINED.md#definition-of-an-unmaintained-crate)

TianyiShi2001:
> We generally reserve unmaintained status for crates the maintainers of which are completely unresponsive. Since the author did reply, perhaps they could hand over the crates.io upload rights to a new maintainer, or at least add the https://github.com/rust-bus organization as an owner so they could transfer the crate later if someone wants to take over maintenance in the future?

I'm the author of audiotags. I've invited rust-bus-owner to own my crate. Let me know if someone wants to take over and I'm happy to transfer the ownership of the GitHub repository as well.

Serial-ATA:
Is a crate under rust-bus still considered unmaintained, or should this be closed?

pinkforest:
Would this have any security implications because it has not been maintained ?

We are planning on new Deprecated status to deal with these on a separate maintenance database that is not necessarily direct security related so we can use that database for this instead.

RustSec advisories are inherently security related and in my opinion we need to justify all advisories inherently as security related when read together with the explicit policies related to maintenance and unmaintained status - hope this  makes sense :)

Serial-ATA:
Ah, that makes sense.

I see that `CONTRIBUTING.md` mentions, "RustSec is a database of security vulnerabilities," I had just assumed unmaintained advisories were a special case since it wasn't mentioned in `HOWTO_UNMAINTAINED.md`.

I wouldn't say this is a security issue, so I'll close this. :)

pinkforest:
@Serial-ATA I interpreted the policy changes wrong and have to correct myself - 

If nobody will pick-up the maintenance on this for a week or two then this might get still merged due to supply chain concerns.

It all depends and is situational but in general our new explicit policy is to address informational advisories where there is no concrete threat.

Apologies for the confusion.

martpie:
poke @pinkforest, who is taking over audiotargs

pinkforest:
@Serial-ATA you may want to close this now :) rust-audio-visual org is / has taken over and has merged audiotags2 back into audiotags - thank you @martpie :purple_heart:

8573:
Now `audiotags2` is unmaintained, right?

martpie:
Yes, I archived it, please move back to `audiotags` :)

PR #1150, opened 2022-01-17
Add unmaintained advisory for r2d2-odbc

bbigras:
fix #1097

PR #1296, opened 2022-08-02
Fix r2d2_odbc name

pinkforest:
Un-breaks the main :blush: 

https://github.com/rustsec/advisory-db/runs/7636777341?check_suite_focus=true

https://github.com/rustsec/advisory-db/pull/1150

_TODO action: document the _ - strangeness in crates.io_


tarcieri:
Need to fix the directory name as well

pinkforest:
yup figured :) passes now :partying_face:

PR #1310, opened 2022-08-04
RUSTSEC-2020-0159: remove "withdrawn"

tarcieri:
Now that there's an actionable fix, we should encourage people to upgrade


pinkforest (approved):
Good idea imho. lgtm

PR #1319, opened 2022-08-04
Move tower-http out from year 2021

pinkforest:
Moving tower-http year forward
See: https://github.com/rustsec/advisory-db/pull/1165


tarcieri:
Has a replacement advisory been filed?

pinkforest:
Yes but not merged yet -
https://github.com/rustsec/advisory-db/pull/1320

PR #1323, opened 2022-08-05
Add Triage Labels Guide

pinkforest:
Adds guide to use labels for triaging PRs

### DIFF
--- a/MAINTAINERS_GUIDE.md
+++ b/MAINTAINERS_GUIDE.md
@@ -2,27 +2,6 @@
 
 Reviewing a pull request, step by step:
 
-## (Optional) Check Triage Labels
-
-It is helpful to Categorise by what the PR Author intent is - see: Propose-Close/Merge for proposed.
-
-- `Unsound` - The PR author suggests the advisory is informational/unsound
-- `Unmaintained` - The PR author suggests the advisory is informational/unmaintained
-
-It is sometimes important to ensure we know what is going on with each PR so we can efficiently check and follow-up.
-
-- `Waiting-Maintainer` - We are waiting for the Maintainer to provide any response if any.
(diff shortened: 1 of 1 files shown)


Shnatsel:
I am hesitant to add extra steps to an already lengthy process, so let's look at each and see if it solves an actual problem:

- `Unmaintained` - I think the distinction between unmaintained and a vulnerability is already apparent from the title.
- `Unsound` -  Is the distinction between an unsound crate and a vulnerability important enough to surface it in the pull request overview? I cannot think of any reason why I would need this info at a glance in the list, or why I should spend time tagging it.
- `Waiting-*` sounds like useful info to have at a glance. However, is there a mechanism to automatically remove them once the author or the maintainer responds? It sounds like they would get out of date easily, even if manually tagged at the start.
- `Propose-*` - is there an advantage to using labels over communicating this verbally? When I see complex labels of things like CNCF projects, I feel like I'm in the middle of some bysantine process I don't understand, and it makes the project feel less approachable to me, subjectively. Also, only maintainers can edit labels, which creates a perceived power asymmetry.

pinkforest:
It definitely helped myself to pick up things past couple of days when having to catch up where that PR was at.

It also signals everyone very clearly without having to read a lot of text what is going.

It also helps to nudge people - just writing about it doesn't necessarily make it as "official" sounding as a label does.

Comes down to psychology and awareness.

Shnatsel:
Now that they're optional, perhaps move it to the end of the guide?

pinkforest:
Done

Shnatsel (approved):
SGTM but I'd prefer to hear from another maintainer

alex (commented):
LGTM.

Perhaps we could even add a github action to apply the unsound/unmaintained labels automatically

PR #1290, opened 2022-08-01
Add advisories for async-graphql/juniper denial of service issues

djc:
See #1285.

Issue #1285, opened 2022-07-30
DoS vulnerabilities in juniper and async-graphql


djc:
https://github.com/async-graphql/async-graphql/security/advisories/GHSA-xq3c-8gqm-v648

https://github.com/graphql-rust/juniper/security/advisories/GHSA-4rx6-g5vg-5f3j

djc:
Unfortunately the fix release for async-graphql is not yet published:

https://github.com/async-graphql/async-graphql/discussions/1000

pinkforest:
Great - Thanks good catch - Apparently async-graphql patched published now

Feel free to send a PR or I can do later

pinkforest:
This was merged #1290

PR #1326, opened 2022-08-05
Fix async-graphql patched

pinkforest:
@djc looks like I didn't catch the patched on async-graphql

https://github.com/rustsec/advisory-db/issues/1285
https://github.com/rustsec/advisory-db/pull/1290
https://github.com/rustsec/advisory-db/pull/1298

Could you sanity ok this for me pls ? Thanks


djc (approved):
Oops, sorry about that!

PR #1332, opened 2022-08-08
Fix "ghost" ID assignment PRs

Shnatsel:
Addresses https://github.com/rustsec/rustsec/issues/630


alex (approved):

Shnatsel:
I'm not confident this will fix anything, actually, but it seems to be a good idea regardless.

PR #1337, opened 2022-08-09
oqs: PQC signature scheme Rainbow level I parameterset broken

thomwiggers:
The Rainbow I parameterset as packaged in liboqs has been proven to be vulnerable to cryptanalysis. It's been removed from the upcoming (but not yet out) 0.7.2 release.

### DIFF
--- a/crates/oqs/RUSTSEC-0000-0000.md
+++ b/crates/oqs/RUSTSEC-0000-0000.md
@@ -1,0 +1,29 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "oqs"
+date = "2022-02-25"
+url = "https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/KFgw5_qCXiI?pli=1"
+categories = ["crypto-failure"]
+
+
+# Optional: metadata which narrows the scope of what this advisory affects
+[affected.functions]
+"oqs::sig::Algorithm::RainbowIaClassic" = ["< 0.5.0, >= 0.2.0"]
+"oqs::sig::Algorithm::RainbowIaCyclic" = ["< 0.5.0, >= 0.2.0"]
+"oqs::sig::Algorithm::RainbowIaCyclicCompressed" = ["< 0.5.0, >= 0.2.0"]
+"oqs::sig::Algorithm::RainbowIClassic" = ["< 0.7.2, >= 0.5.0"]
+"oqs::sig::Algorithm::RainbowICircumzenithal" = ["< 0.7.2, >= 0.5.0"]
+"oqs::sig::Algorithm::RainbowICompressed" = ["< 0.7.2, >= 0.5.0"]
+
+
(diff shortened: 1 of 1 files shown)


thomwiggers:
By the way, how do I address the `oqs-sys` crate?

crates/oqs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,29 @@
...
+
+
+# Optional: metadata which narrows the scope of what this advisory affects
+[affected.functions]
+"oqs::sig::Algorithm::RainbowIaClassic" = ["< 0.5.0, >= 0.2.0"]
+"oqs::sig::Algorithm::RainbowIaCyclic" = ["< 0.5.0, >= 0.2.0"]
+"oqs::sig::Algorithm::RainbowIaCyclicCompressed" = ["< 0.5.0, >= 0.2.0"]
+"oqs::sig::Algorithm::RainbowIClassic" = ["< 0.7.2, >= 0.5.0"]
+"oqs::sig::Algorithm::RainbowICircumzenithal" = ["< 0.7.2, >= 0.5.0"]
+"oqs::sig::Algorithm::RainbowICompressed" = ["< 0.7.2, >= 0.5.0"]
```
thomwiggers, line 17: these are enum variants, not strictly functions, by the way.

crates/oqs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,29 @@
...
+
+
+# Versions which include fixes for this vulnerability (mandatory)
+[versions]
+patched = [">= 0.7.2"]
```
amousset, line 22: if versions < 0.2.0 are totally unaffected (which affected.functions seems to indicate) you can add:

```
unaffected = ["< 0.2.0"]
```

thomwiggers: They're probably still affected but the crate was using a completely different API back then, so the enum variants aren't present.

crates/oqs/RUSTSEC-0000-0000.md
tarcieri: They need to be functions (i.e. present in a call graph).

Unfortunately we don't have a proper lint for this yet.

thomwiggers: Okay, will it hurt to keep it like this (#1338 did the same), or is it fine to keep it like this?

tarcieri: It should be removed (from this PR and the advisory added in #1338).

The intended use of this information is call graph analysis, and enum variants are not present in the call graph.

crates/oqs/RUSTSEC-0000-0000.md
thomwiggers: See also discussion in #1340

amousset (approved):

amousset:
Thanks!

PR #1338, opened 2022-08-09
oqs: SIKE and SIDH are insecure

thomwiggers:
Wouter Castryck and Thomas Decru presented an efficient key recovery attack on the SIDH protocol.
As a result, the secret key of SIKEp751 can be recovered in a matter of hours.
The SIKE and SIDH schemes will be removed from the next oqs version 0.7.2.


Shnatsel:
Thanks!

PR #1341, opened 2022-08-09
Updated README with info on informational advisories

eminence:


README.md
```diff
@@ -56,6 +56,10 @@ date = "2021-01-31"
 # a change log entry, or a blogpost announcing the release (optional)
 url = "https://github.com/mystuff/mycrate/issues/123"
 
+# Optional: Indicates the type of informational advisory
+#informational = "unsound"
+#informational = "unmaintained"
```
eminence, line 61: Note:  I didn't mention `informational = "notice"` since there's only one advisory with that type, and I didn't know if that was a type that should be promoted

README.md
```diff
@@ -129,3 +133,5 @@ All content in this repository is placed in the public domain.
 [cargo-audit]: https://github.com/rustsec/cargo-audit
 [cargo-deny]: https://github.com/EmbarkStudios/cargo-deny
 [CONTRIBUTING.md]: https://github.com/RustSec/advisory-db/blob/main/CONTRIBUTING.md
+
+[^1]: The database also contains non-security advisories (known as informational advisories), such as advisories about unmaintained crates.
```
tarcieri, line 137: Maybe something like this...

```suggestion
[^1]: The database also contains non-security advisories (known as informational advisories), such as advisories about unmaintained crates,
which are optionally surfaced as warnings.
```

(it could probably be worded a little better)

eminence: That reads pretty good to me.  Maybe:

> The database also contains non-security advisories (known as informational advisories), such as advisories about unmaintained crates, which are optionally surfaced as warnings in `cargo audit`

(just to avoid implying that things like https://rustsec.org/advisories/ will hide informationals)

README.md
tarcieri: It has an intended use case which we've almost used a few times: allowing project maintainers to make security-related announcements about their project which don't map directly to a concrete vulnerability.

One of the main concrete examples of this is if the project provided example code or usage documentation which contains some sort of insecure pattern, which may have been copied-and-pasted into other projects and needs to be updated.

README.md
tarcieri: Sure, sounds good

README.md
eminence: Ok, so maybe we can include a terse description of each:

```toml
# Optional: Indicates the type of informational advisory
#  - "unsound" for soundness issues
#  - "unmaintained" for crates that are no longer maintained
#  - "notice" for security-related notices that don't have a associated concrete vulnerability
#informational = "unmaintained"
```

tarcieri: Sure

PR #1355, opened 2022-08-14
Document `withdrawn`

### LINKED ISSUE #928: document the meaning of yanked = true
A small number of advisories (for example [RUSTSEC-2020-0053](https://github.com/RustSec/advisory-db/blob/main/crates/dirs/RUSTSEC-2020-0053.md)), contain:

    yanked = true

The meaning of the key is not documented in the [Advisory Format](https://github.com/RustSec/advisory-db#advisory-format) section of the README. Best I can tell, `cargo-audit` ignores these advisories unless specifically asked to include them.

pinkforest:
Fixes #928

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,13 @@ package = "mycrate"
 # Disclosure date of the advisory as an RFC 3339 date (mandatory)
 date = "2021-01-31"
 
+# Whether the advisory is withdrawn (optional)
+#withdrawn = "YYYY-MM-DD"
+
+# Whether the advisory is yanked (deprecated from 0.26.1) (optional)
+# If withdrawn is used this must be also set true (until 0.26.1 rustsec)
+#yanked = false
+
 # URL to a long-form description of this issue, e.g. a GitHub issue/PR,
 # a change log entry, or a blogpost announcing the release (optional)
 url = "https://github.com/mystuff/mycrate/issues/123"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,6 +55,13 @@ package = "mycrate"
 # Disclosure date of the advisory as an RFC 3339 date (mandatory)
 date = "2021-01-31"
 
+# Whether the advisory is withdrawn (optional)
+#withdrawn = "YYYY-MM-DD"
+
+# Whether the advisory is yanked (deprecated from 0.26.1) (optional)
+# If withdrawn is used this must be also set true (until 0.26.1 rustsec)
+#yanked = false
```
tarcieri, line 63: We can just removed `yanked` at this point. No need to document it.

pinkforest: yanked it thx - rustsec-admin gets 0.26 but since it's cached at linter will it get 0.26.1 or do we need to flush the cache ?

tarcieri: If we encounter places the old version is cached we can just bump the cache key

pinkforest: k. I will send a PR to remove them and see if the linter still fails at it

pinkforest: oh we need rustsec 0.26.1 release btw so rustsec-admin picks it up

tarcieri: released

PR #1358, opened 2022-08-14
Add advisory `tempfile`

pinkforest:
Since this crate also advertises itself as being secure and portable it might be feasible to remove recommendations and / or flag an advisory either as informational or regular where the crate makes these guarantees and may not provide them.

https://github.com/Stebalien/tempfile/issues/178

We previously recommended to use `tempfile` at `temporary` here: https://github.com/rustsec/advisory-db/pull/1196

https://owasp.org/www-community/vulnerabilities/Insecure_Temporary_File

**Proposal**

Removing the recommendation to use `tempfile` from the earlier `temporary` advisory
Also removes advertised reference from `tempdir`

Should we also file an advisory in addition not recommending it ? 

There is no real PoC or anything but :woman_shrugging: 

Perhaps we should use the `informatonal = "notice"` category finally ?

@5225225 and @vks could you review this please ?

### DIFF
--- a/crates/tempdir/RUSTSEC-2018-0017.md
+++ b/crates/tempdir/RUSTSEC-2018-0017.md
@@ -13,5 +13,4 @@ unaffected = []
 
 # `tempdir` crate has been deprecated; use `tempfile` instead
 
-The [`tempdir`](https://crates.io/crates/tempdir) crate has been deprecated
-and the functionality is merged into [`tempfile`](https://crates.io/crates/tempfile).
+The [`tempdir`](https://crates.io/crates/tempdir) crate has been deprecated.
--- a/crates/tempfile/RUSTSEC-0000-0000.md
+++ b/crates/tempfile/RUSTSEC-0000-0000.md
@@ -1,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tempfile"
+date = "2022-05-01"
+url = "https://github.com/Stebalien/tempfile/issues/178"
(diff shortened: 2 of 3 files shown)


5225225:
From what I can tell, the only possible issue is a denial of service. As in, tempfile acts like `mkstemp`. The advisory is rather vague, if we do go ahead with this, i'd say "it's not secure against denial of service attacks if an attacker can create files in the directory that `std::env::temp_dir` returns, because an attacker can predict the names going to be created and create files in the temporary directory"

amousset:
I don't think we should file an advisory for this one, at least for now. It looks like the security expectations for temporary files creation methods are not clear-cut, and I don't think we should publish an non-actionable advisory for a hypothetical issue.

We could continue discussing possible improvements with the maintainer (as @5225225 just asked in https://github.com/Stebalien/tempfile/issues/178#issuecomment-1214352980).

5225225:
Yeah, "don't do it", or at the very least "do it, but only as a "hey we fixed it in this newer version you should all go update"" seems like the best plan.

tarcieri:
Yeah, it seems to me like this would be a very noisy advisory with relatively low impact. 

Agreed it's fine as encouragement to upgrade. `information = "notice"` seems OK although several people run with `--deny warnings` so it would still be high-impact there.

pinkforest:
Yeah agreed, I'll just close and we can re-hash later if needed. As positive note at least we got some upstream fix going.

Thanks all :purple_heart:

PR #1362, opened 2022-08-14
Remove `yanked`

pinkforest:
CI will fail on this most probably as the 0.26.1 rustsec is not published that deprecates the field: https://github.com/rustsec/rustsec/pull/631
After the rustsec 0.26.1 is published this probably requires cache bump in CI linter as discussed here: https://github.com/rustsec/advisory-db/pull/1355
PR to bump the CI cache key: https://github.com/rustsec/advisory-db/pull/1363

**Before**
```
 grep -R yanked *
crates/crossbeam-utils/RUSTSEC-2022-0041.md:Affected 0.8.x releases have been yanked.
crates/libpulse-binding/RUSTSEC-2020-0055.md:yanked = true
crates/directories/RUSTSEC-2020-0054.md:yanked = true
crates/tower-http/RUSTSEC-2021-0135.md:yanked = true
crates/reorder/RUSTSEC-2021-0050.md:Previous versions have also been yanked from crates.io.
crates/spin/RUSTSEC-2019-0031.md:yanked = true
crates/spin/RUSTSEC-2019-0031.md:unaffected = [">= 0"] # workaround for `yanked = true` not removing the advisory
crates/sha2/RUSTSEC-2021-0100.md:The crate has since been yanked, but any users who upgraded to v0.9.7 should
crates/rustsec-example-crate/RUSTSEC-2019-0024.md:(Technically there is a third release, v0.0.0, which is yanked, but otherwise
crates/dirs/RUSTSEC-2020-0053.md:yanked = true
```

**After**
```
$ grep -R yanked *
crates/crossbeam-utils/RUSTSEC-2022-0041.md:Affected 0.8.x releases have been yanked.
crates/reorder/RUSTSEC-2021-0050.md:Previous versions have also been yanked from crates.io.
crates/spin/RUSTSEC-2019-0031.md:unaffected = [">= 0"] # workaround for `yanked = true` not removing the advisory
crates/sha2/RUSTSEC-2021-0100.md:The crate has since been yanked, but any users who upgraded to v0.9.7 should
crates/rustsec-example-crate/RUSTSEC-2019-0024.md:(Technically there is a third release, v0.0.0, which is yanked, but otherwise
```

_Note: Need to check RUSTSEC-2019-0031 what was the case with it and if something needs to be done_

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -16,12 +16,12 @@ jobs:
       uses: actions/cache@v1
       with:
         path: ~/.cargo/bin
-        key: rustsec-admin-v0.8.0
+        key: rustsec-admin-v0.8.1
 
     - name: Install rustsec-admin
       run: |
         if [ ! -f $HOME/.cargo/bin/rustsec-admin ]; then
-            cargo install rustsec-admin --vers 0.8.0
+            cargo install rustsec-admin --vers 0.8.1
         fi
 
     - name: Lint advisories
--- a/README.md
+++ b/README.md
@@ -38,7 +38,8 @@ See [CONTRIBUTING.md] for more information.
 See [EXAMPLE_ADVISORY.md] for a template.
 
 Advisories are formatted in [Markdown] with [TOML] "front matter".
-Below is the schema of the "front matter" section of an advisory:
+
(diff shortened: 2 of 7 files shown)


tarcieri (approved):

pinkforest:
Re-doing the PR

PR #1366, opened 2022-08-16
iana-time-zone v0.1.{43,44}: Use after free in MacOS / iOS implementation

Kijewski:
In iana-time-zone v0.1.43 a use-after-free bug in the MacOS / iOS implementation was introduced.

The copied system time zone was released before its name was copied.
If the system time zone was changed between the call of `CFRelease` and `str::to_owned()`,
random memory would be copied.

### DIFF
--- a/crates/iana-time-zone/RUSTSEC-0000-0000.md
+++ b/crates/iana-time-zone/RUSTSEC-0000-0000.md
@@ -1,0 +1,26 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "iana-time-zone"
+date = "2022-08-15"
+url = "https://github.com/strawlab/iana-time-zone/pull/54"
+references = ["https://github.com/strawlab/iana-time-zone/pull/50#discussion_r945353515"]
+categories = ["memory-corruption"]
+informational = "unsound"
+
+[affected]
+os = ["ios", "macos"]
+functions = { "iana_time_zone::get_timezone" = ["> 0.1.42, < 0.1.45"] }
+
+[versions]
+patched = [">= 0.1.45"]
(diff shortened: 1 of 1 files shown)


pinkforest (approved):
Awesome thank you! :purple_heart:

pinkforest:
@Kijewski btw you might want to add `informational = "unsound"` here so it goes more as a warning instead of full error

But up to you - ought to give an option which one to use :)

crates/iana-time-zone/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "iana-time-zone"
+date = "2022-08-15"
+url = "https://github.com/strawlab/iana-time-zone/pull/54"
+references = ["https://github.com/strawlab/iana-time-zone/pull/50#discussion_r945353515"]
+categories = ["memory-corruption"]
+
```
pinkforest, line 9: @Kijewski btw you might want to add informational = "unsound" here so it goes more as a warning instead of full error

But up to you - ought to give an option which one to use :)

pinkforest: Great - Thank you for being responsible and proactive maintainer - welcome aboard contributing :ship: :rocket:

Kijewski: Yes, you are right. I opened #1368. :)

pinkforest: ta. merged :ship:

PR #1368, opened 2022-08-16
RUSTSEC-2022-0049: recategorize as memory-exposure

Kijewski:
Because of the bug random data was read, but still written into a sane buffer.


pinkforest (approved):
Awesome thanks :purple_heart:

PR #1371, opened 2022-08-17
Add a dependabot config

LingMan:
Dependabot will watch the workflow files in .github/workflows and
automatically file PRs if there are updates available for any of the used
Actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10


tarcieri:
Interesting. Does it pin them to hashes, or what?

LingMan:
No, from what I've seen it updates to the latest major tag (`@v1` -> `@v3`).

pinkforest (approved):
I would be in favor having Dependabot :robot: bumping PR on checkout, cache and all the rest of the actions it monitors around :octocat:

tarcieri:
It'd be really cool if it could actually pin to the latest tagged commit hash, but in the meantime this seems ok

alex:
If you pin by hash it will send bumps to the hash. And if that hash is a
tag, it appears to only bump tags. See pyca/cryptography#7522 for example

בתאריך יום ה׳, 18 באוג׳ 2022, 9:44, מאת Tony Arcieri ‏<
***@***.***>:

PR #1383, opened 2022-08-25
lz4-sys: Forward CVE-2021-3520

niklasf:
Forwarding CVE-2021-3520. lz4-sys 1.9.4 has not yet been released.

### DIFF
--- a/crates/lz4-sys/RUSTSEC-0000-0000.md
+++ b/crates/lz4-sys/RUSTSEC-0000-0000.md
@@ -1,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "lz4-sys"
+date = "2022-08-25"
+url = "https://github.com/lz4/lz4/pull/972"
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you!

Has the issue been reported to `lz4-sys` maintainers?

niklasf:
I just raised it in https://gitter.im/bozaro/lz4-rs.

cc @10XGenomics, @pmarks

pmarks:
Thanks, I published updated crate for lz4-sys:1.9.4 and lz4:1.24.  Should I do anything else? Does the old version need to be yanked or something?

Shnatsel:
Yes, yanking affected versions is a good idea. I see there have only ever been 1.x releases, so there is no risk of users on an older series finding all their versions yanked and no semver-compatible replacement available.

Shnatsel:
Merged. Thanks @niklasf for the report, and @pmarks for publishing the update!

PR #1389, opened 2022-08-27
Add unsound `mozjpeg`

### LINKED ISSUE #327: File advisories for vulnerabilities with upcoming fixes
I am aware of a number of memory safety issues that are slated to get fixes soon™ but do not have a release shipping yet. My life is a bit of a struggle as of late, so I don't have the capacity to follow up on them properly. I'm dumping them here so that someone else could follow up on them and issue advisories.

- [x] https://github.com/actix/actix-net/issues/160
- [x] https://github.com/actix/actix-net/issues/91
- [x] https://github.com/actix/actix-web/issues/1321
- [x] https://github.com/kornelski/rust-rgb/issues/35
- [ ] https://github.com/ImageOptim/mozjpeg-rust/issues/10

pinkforest:
Closes #327

https://github.com/ImageOptim/mozjpeg-rust/issues/10

### DIFF
--- a/crates/mozjpeg/RUSTSEC-0000-0000.md
+++ b/crates/mozjpeg/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "mozjpeg"
+date = "2020-07-04"
+url = "https://github.com/ImageOptim/mozjpeg-rust/issues/10"
+categories = ["memory-corruption"]
+references = ["https://github.com/kornelski/rust-rgb/issues/35", "https://rustsec.org/advisories/RUSTSEC-2020-0029.html"]
+informational = "unsound"
+
+[versions]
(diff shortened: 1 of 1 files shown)


HeroicKatora (commented):
Factually looks good. Style nit: The text refers to an explanation from RUSTSEC-2020-0029 but there's no link to it, I don't know how this is usually handled. As  an alternative, the explanation could be inlined. Suggestion:

> Affected versions of mozjpeg crate allow creating instances of any type `T` from bytes,
and do not correctly constrain `T` to the types for which it is safe to do so.
> Examples of safety violation possible for a type `T`:
> * `T` contains a reference type, and it constructs a pointer to an invalid, arbitrary memory address.
> * `T` requires a safety and/or validity invariant for its construction that may be violated.

pinkforest:
Thanks - we use `references = ["url", ..]` for this but added the inlined suggestion thanks!

PR #1393, opened 2022-08-28
Remove reference mention from description

amousset:
It is already the url of the advisory metadata, and the internal github link does not render outside of github.


Shnatsel:
Thanks!

PR #1407, opened 2022-09-02
Mention `structopt` advice under `ansi_term`

pinkforest:
Linking structopt/clap2 specific actionable advice: https://github.com/clap-rs/clap/discussions/4172 under `ansi_term`

Not sure if this is feasible but see what others think :woman_shrugging: 

As seen here:

- https://github.com/clap-rs/clap/discussions/4172
- https://github.com/TeXitoi/structopt/issues/525
- https://github.com/mrl5/vulner/issues/59#issuecomment-1221388712
- https://docs.rs/structopt/latest/structopt/#maintenance

I've asked @epage and @TeXitoi whether it would be also feasible to advice informational under `structopt` to optionally move to `clap` v3 but it will be @TeXitoi's sole decision if they wish to do that - the crate should be still fine to use as-is I think though.

### DIFF
--- a/crates/ansi_term/RUSTSEC-2021-0139.md
+++ b/crates/ansi_term/RUSTSEC-2021-0139.md
@@ -29,4 +29,4 @@ Last release seems to have been three years ago.
 
 ## Dependency Specific Migration(s)
 
- - [structopt, clap2](https://github.com/TeXitoi/structopt/issues/525)
+ - [structopt, clap2](https://github.com/clap-rs/clap/discussions/4172)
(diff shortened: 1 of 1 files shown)


epage (approved):

TeXitoi (approved):

TeXitoi:
Technically, structopt is maintained. Just that breaking changes are out of scope as clap v3 is the breaking change of structopt.

pinkforest:
Cool, we'll just flag some on point advice related to `ansi_term` this time then - thanks for the help.

PR #1416, opened 2022-09-08
Add umaintained `num-format`

### LINKED ISSUE #1405: `num-format` Status
_3,232,716 downloads, ~8k a day_

Last release was over three years ago

It is using the old version of itoa: https://github.com/rustsec/advisory-db/issues/1404

Ralf was helpful to ping earlier: https://github.com/bcmyers/num-format/issues/29

Maintenance status was asked on 9 Jan 2022
https://github.com/bcmyers/num-format/issues/27

@bcmyers - I wonder if people should be still using this crate today and whether it would be helpful to bump the itoa dependency

pinkforest:
Fixes #1405 

Mind reviewing this avisory @LovecraftianHorror @RalfJung 

Would be nice to know if there are maintained fork(s) or alternative implementation(s) around to perhaps list

Cheers

https://github.com/bcmyers/num-format/issues/21, https://github.com/bcmyers/num-format/issues/27, https://github.com/bcmyers/num-format/issues/29

### DIFF
--- a/crates/num-format/RUSTSEC-0000-0000.md
+++ b/crates/num-format/RUSTSEC-0000-0000.md
@@ -1,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "num-format"
+date = "2022-01-09"
+url = "https://github.com/bcmyers/num-format/issues/27"
+references = ["https://github.com/bcmyers/num-format/issues"]
+informational = "unmaintained"
+
+[versions]
+patched = []
+```
+# num-format is Unmaintained
+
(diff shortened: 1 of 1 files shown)


LovecraftianHorror:
My memory is hazy, but I think there are generally two aspects of `num-format` that people care about (and typically people only seem to care about one or the other). Localization and speed

I don't know of any alternatives that have a focus on speed, but the `icu` crate has support for [localized number formatting](https://docs.rs/icu/latest/icu/decimal/index.html) (although that crate has much broader scope than `num-format`)

pinkforest:
https://github.com/bcmyers/num-format/issues/27#issuecomment-1272626722

PR #1418, opened 2022-09-11
Add `alloc_no_stdlib`

Ben-Lichtman:
alloc_no_stdlib contains a memory disclosure vulnerability (reading uninitialized memory)

### DIFF
--- a/crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
+++ b/crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
@@ -1,0 +1,37 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+
+package = "aalloc-no-stdlib"
+
+date = "2022-11-09"
+
+url = "https://github.com/dropbox/rust-alloc-no-stdlib/issues/12"
+
+informational = "unsound"
+
+categories = ["memory-exposure"]
+
+keywords = ["uninitialized", "information-leak", "memory-exposure"]
+
+[affected]
+
+[versions]
+patched = []
+```
+
+# alloc_no_stdlib gives out references to uninitialized data to library users
+
+The following code demonstrates this library giving out a slice of uninitialized memory as its allocation 
+
+```rust
+let mut malloc_global_buffer =
+	unsafe { define_allocator_memory_pool!(4096, u8, [0; 200 * 1024 * 1024], malloc) };
+let mut ags =
+	MallocAllocatedFreelist4096::<u8>::new_allocator(&mut malloc_global_buffer.data, bzero);
+let mut uninit = ags.alloc_cell(4);
+let uninit_slice = uninit.as_mut();
+let boom = uninit_slice[0]; // reading uninitized data from the allocation
+```
+
+This is dangerous because it would allow information disclosure by reading data that was previously allocated where the memory pool currently is, as well as causing UB and a risk of miscompilation.
--- a/crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
+++ b/crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
@@ -33,3 +33,5 @@ let boom = uninit_slice[0]; // reading uninitized data from the allocation
 ```
 
 This is dangerous because it would allow information disclosure by reading data that was previously allocated where the memory pool currently is, as well as causing UB and a risk of miscompilation.
+
+It has had an issue open regarding this unsoundness for almost a year with no actions taken, and the crate appears somewhat unmaintained also.
--- a/crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
+++ b/crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
@@ -4,7 +4,7 @@ id = "RUSTSEC-0000-0000"
 
 package = "alloc-no-stdlib"
 
-date = "2022-11-09"
+date = "2022-01-31"
 
 url = "https://github.com/dropbox/rust-alloc-no-stdlib/issues/12"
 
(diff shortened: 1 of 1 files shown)


Ben-Lichtman:
This crate seems to be a dependency of both dropbox, and brotli (the compression crate)

pinkforest:
https://github.com/dropbox/rust-alloc-no-stdlib/issues/12

Upstream should acknowledge and ideally fix before we merge -

Looks like the maintainer has not acknowledged this yet ?

It seems that the crate may be unmaintained and @lesniewski and @danielrh - or someone from dropbox org - should be chased up first and see if there is path to fix as the issue.

At this stage proposing not to to merge until we have actionable fix or status found out e.g. whether the crate is simply unmaintained and / or whether the vendor acknowledges the problem and potentially fixes it - which would be ideal.

crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,37 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+
```
pinkforest, line 4: nit: please can we remove these empty lines

crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,37 @@
...
+package = "aalloc-no-stdlib"
```
pinkforest, line 5:
```suggestion
package = "alloc-no-stdlib"
```
This should also be in `alloc-no-stdlib` directory not in `alloc_no_stdlib`

crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,37 @@
...
+
+date = "2022-11-09"
```
pinkforest, line 7:
```suggestion
date = "2022-01-31"
```

this date is the date when it was reported originally e.g. when saethlin raised it on

crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,37 @@
...
+url = "https://github.com/dropbox/rust-alloc-no-stdlib/issues/12"
+
+informational = "unsound"
+
+categories = ["memory-exposure"]
+
+keywords = ["uninitialized", "information-leak", "memory-exposure"]
+
+[versions]
+patched = []
```
pinkforest, line 18: upstream should fix it and acknowledge this first.

crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,37 @@
...
+let mut ags =
+	MallocAllocatedFreelist4096::<u8>::new_allocator(&mut malloc_global_buffer.data, bzero);
+let mut uninit = ags.alloc_cell(4);
+let uninit_slice = uninit.as_mut();
+let boom = uninit_slice[0]; // reading uninitized data from the allocation
+```
+
+This is dangerous because it would allow information disclosure by reading data that was previously allocated where the memory pool currently is, as well as causing UB and a risk of miscompilation.
+
+It has had an issue open regarding this unsoundness for almost a year with no actions taken, and the crate appears somewhat unmaintained also.
```
pinkforest, line 37: If the maintainer does not acknowledge the issue (it is best to tag the maintainers in issues) then we can flag `informational = "unmaintained"` instead

If there is no fix e.g. it's unmaintained then we need to recommend alternative maintained crate(s) if any as actionable potential fix(es) which may or may not be alternative(s).

pinkforest:
I pinged several dropbox people here: https://github.com/dropbox/rust-alloc-no-stdlib/issues/16 re: maintenace status

I have a feeling this may be abandoned .. but will let the maintainer(s) if any to chime in -

If we were to go `informational = "unmaintained"` route then we would just round up all the non-std alternatives I guess

crates/alloc_no_stdlib/RUSTSEC-0000-0000.md
Ben-Lichtman: Done

Ben-Lichtman:
Hmm, now that I look at it more, It looks like you at least need to call unsafe to create the initial allocator backing memory... So maybe It should be considered sound...?

Ben-Lichtman:
Looks like only the "malloc"  version of the define memory pool macro can lead to handing out uninitialized memory, and brotli doesn't seem to use that at all, so I think it's probably safe to reject the advisory

pinkforest:
Ok the maintainer has appropriately responded and done a release fix 2.0.4 so this crate is clarified as maintained status.

The commit to fix it was here that ended up to 2.0.4: https://github.com/dropbox/rust-alloc-no-stdlib/commit/6032b6a9b20e03737135c55a0270ccffcc1438ef

To me personally regardless-  of anything - if a memory allocator gives out random stuff out of memory besides anything is concerning but whether we can use `informational = "unsound"` to classify it has it's problems when the maintainer does not agree that this is anything but a bug we can't use `informational = "notice"` either :woman_shrugging: 

And yes since this crate API requires you to use use unsafe to use it this unsound way .. where the safety guarantees don't extend to..  indeed stop us flagging `informational = "unsound"` that is correct intepretation :ok_woman: 

So I would agree and veering towards closing this since there isn't concrete downstream impact either that is affected that would lead us to raise issue from downstream basis - as a precedent we've had unicase - https://github.com/rustsec/advisory-db/pull/1176 - not to flag advisories like this where we don't have a category and neither the maintainer doesn't want us to flag advisory to warn on it's use or where we don't have anything concrete issues from downstream to flag on anything.

@5225225 you've been deep ends on to these and and assisted before on these so interested to hear your opinion as well.

Ben-Lichtman:
Ok. I think I will close this PR for now. Can always reopen if there are any further developments

pinkforest:
Thanks for the effort anyways - it was a good to clear and clarify this out and the maintainer fixed it anyways so win-win for everyone.

PR #1429, opened 2022-09-28
fix hyper patched version number

RalfJung:
That said, the known risks of miscompilation are actually completely mitigated by https://github.com/rust-lang/rust/pull/99182. So I wonder if we should remove the advisory?

### DIFF
--- a/crates/hyper/RUSTSEC-2022-0022.md
+++ b/crates/hyper/RUSTSEC-2022-0022.md
@@ -7,7 +7,7 @@ informational = "unsound"
 url = "https://github.com/hyperium/hyper/pull/2545"
 
 [versions]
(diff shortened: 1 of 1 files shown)


pinkforest:
Thanks. We don't take into account rustc version mitigations.

pinkforest:
This is already fixed ?
https://github.com/rustsec/advisory-db/pull/1250/files

RalfJung:
I did a git pull --rebase and the patch showed as unapplied. And github says there is a diff.

Did you rename the default branch? That that would obviously wreck havok on all existing checkouts...

pinkforest:
Yeah I think we renamed it at one stage to main I think, many thanks for the effort anyways.

PR #1430, opened 2022-09-28
add inconceivable

RalfJung:
This is similar to [plutonium](https://github.com/rustsec/advisory-db/tree/main/crates/plutonium).

### DIFF
--- a/crates/inconceivable/RUSTSEC-0000-0000.md
+++ b/crates/inconceivable/RUSTSEC-0000-0000.md
@@ -1,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "inconceivable"
+date = "2022-09-28"
+informational = "notice"
+url = "https://crates.io/crates/inconceivable"
+
+[versions]
+patched = []
+```
+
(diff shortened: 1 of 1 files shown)


pinkforest:
Thanks. I'll merge tomororow in case anyone has comments.

crates/inconceivable/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "inconceivable"
+date = "2022-09-28"
+informational = "notice"
```
Shnatsel, line 6: Perhaps `informational = unsound` would be more appropriate, since it allows injecting `unreachable_unchecked!` in safe code?

I don't have a strong opinion on this though.

RalfJung: Yeah that was my first inclination, too. But plutonium is also just 'notice'.

pinkforest: Yeah we decided / have precedent in the past notice for these

PR #1433, opened 2022-10-11
Add CVE-2022-3358 for openssl-src

amousset:


Shnatsel:
The upstream CVE states

> Fixed in OpenSSL 3.0.6 (Affected 3.0.0-3.0.5).

But the Rust patched specification is `[">= 300.0.10"]`. Is that intentional, with the Rust package remapping the version numbers?

amousset:
Yes, the versions are not an exact mapping and there are increments not linked to an openssl version bump.
image

Shnatsel:
Got it. Thanks!

PR #1436, opened 2022-10-15
Add unmaintained `orbtk`

xTibor:
As per their ["OrbTk is Sunsetting"](https://github.com/redox-os/orbtk/blob/eba9e77821551076bbf1d9f7ab44d788150e3446/README.md#orbtk-is-sunsetting) announcement.

CC @jackpot51, @rzerres, @FloVanGH to confirm this.

### DIFF
--- a/crates/orbtk/RUSTSEC-0000-0000.md
+++ b/crates/orbtk/RUSTSEC-0000-0000.md
@@ -1,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "orbtk"
+date = "2022-10-13"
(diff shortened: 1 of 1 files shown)


pinkforest:
Yep correct we would need a confirmation from the maintainer what the "non-active" maintenance status means -

e.g. Whether the maintainers monitor any security issues - if any pop and potentially address them - and / or accept Pull Requests that address security issues + update dependencies which is passive maintenance stuff.

Also if the maintainer advises that we can gently nudge the users to migrate this way then we can also merge :+1:

jackpot51:
I approve the advisory as it is. There is no guarantee of security issues being resolved in the future. I am recommending the use of either `iced` or `slint` instead of `orbtk` at this time.

PR #1441, opened 2022-10-21
Add unmaintained `parity-wasm`

athei:
The crate was explicitly deprecated by the author. The repo is archived and there is a note on top of the readme: https://github.com/paritytech/parity-wasm

### DIFF
--- a/crates/parity-wasm/RUSTSEC-0000-0000.md
+++ b/crates/parity-wasm/RUSTSEC-0000-0000.md
@@ -2,7 +2,7 @@
 [advisory]
 id = "RUSTSEC-0000-0000"
 package = "parity-wasm"
-date = "2022-01-10"
+date = "2022-10-01"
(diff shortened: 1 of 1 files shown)


pinkforest:
What would be the alternatives to use ? Are there any forks around ?

crates/parity-wasm/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,16 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "parity-wasm"
+date = "2022-01-10"
```
pinkforest, line 5:
```suggestion
date = "2022-10-01"
```

We follow RFC 3339 for this meaning YYYY-MM-DD

https://www.rfc-editor.org/rfc/rfc3339#section-5.1

PR #1444, opened 2022-10-24
Add advisory for logging of access tokens in matrix-sdk

jplatte:
Upstream issue: https://github.com/matrix-org/matrix-rust-sdk/issues/1110

### DIFF
--- a/crates/matrix-sdk/RUSTSEC-0000-0000.md
+++ b/crates/matrix-sdk/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "matrix-sdk"
+date = "2022-10-24"
+url = "https://github.com/matrix-org/matrix-rust-sdk/issues/1110"
+
+[versions]
+patched = [">= 0.6.2"]
+unaffected = ["< 0.6.0"]
+```
+
+# matrix-sdk 0.6.0 logs access tokens
+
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "matrix-sdk"
+date = "2022-10-24"
+url = "https://github.com/matrix-org/matrix-rust-sdk/issues/1110"
+
+[versions]
+patched = [">= 0.6.2"]
```
amousset, line 9: shouldn't this be `>= 0.6.1` which [seems to contain the fix](https://github.com/matrix-org/matrix-rust-sdk/releases/tag/matrix-sdk-0.6.1)?

jplatte: It contains a partial fix only, which I actually only realized after being half way done with the advisory. I'll update the GitHub releases.

jplatte: Does it make sense like this? https://github.com/matrix-org/matrix-rust-sdk/releases

amousset: Yes, thanks for the exlanation :+1:

amousset:
Thanks for reporting!

PR #1448, opened 2022-10-29
Add CVE-2022-36086 for linked_list_allocator

newAM:


Shnatsel:
Thank you for the report! Everything looks good.

Since Github Security Advisories are under CC-BY-SA licence while RustSec is in the public domain, we'll need a confirmation from @phil-opp that they're OK with releasing this information in the public domain before we can merge it.

phil-opp:
Sure!

Shnatsel:
Merging, thanks!

PR #1450, opened 2022-10-31
Add advisory for elf_rs crate

atulkharerivos:
This adds an advisory for the elf_rs crate.

### DIFF
--- a/crates/elf_rs/RUSTSEC-0000-0000.md
+++ b/crates/elf_rs/RUSTSEC-0000-0000.md
@@ -1,0 +1,44 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "elf_rs"
+date = "2022-10-31"
+url = "https://github.com/vincenthouyi/elf_rs"
+categories = ["memory-corruption"]
+keywords = ["elf", "header"]
+
+[versions]
+patched = []
+
+[affected]
+```
+
+# ELF header parsing library doesn't check for valid offset
+
+The crate has several unsafe sections that don't perform proper pointer validation.
+
+An example can be found in the following function:
+
+```
+fn section_header_raw(&self) -> &[ET::SectionHeader] {
+    let sh_off = self.elf_header().section_header_offset() as usize;
+    let sh_num = self.elf_header().section_header_entry_num() as usize;
+    unsafe {
+        let sh_ptr = self.content().as_ptr().add(sh_off);
+        from_raw_parts(sh_ptr as *const ET::SectionHeader, sh_num)
+    }
+}
+```
+
+While this will work perfectly fine *if* the ELF header is valid, malicious or
+malformed input can contain a section header offset of an arbitrary size, meaning
+that the resultant pointer in the unsafe block can point to an artibrary address
+in the address space of the process.
+
+This can result in unpredictable behaviour, and in our fuzz testing, we discovered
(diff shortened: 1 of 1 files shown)


pinkforest:
Has the maintainer asked about this ? I don't see any communication with the maintainer to either confirm or deny ?

atulkharerivos:
@pinkforest I wasn't sure of the protocol, but will follow-up with the maintainer.

FWIW, here's the test program that was used with the Rust afl to reproduce the issue:

```
use elf_rs::*;
#[macro_use]
extern crate afl;

fn elf_reader(buffer: &[u8]) -> Result<(), ()> {
    let elf = Elf::from_bytes(buffer).map_err(|e| {
        println!("failed to extract elf file {:?}", e);
        ()
    })?;

    println!("{:?} header: {:?}", elf, elf.elf_header());

    for p in elf.program_header_iter() {
        println!("{:x?}", p);
    }

    for s in elf.section_header_iter() {
        println!("{:x?}", s);
    }

    if let Some(s) = elf.lookup_section(b".text") {
        println!(".test section {:?}", s);
    }
    Ok(())
}

fn main() {
    fuzz!(|data : &[u8]| {
        let _ = elf_reader(data);
    });
}
```
To reproduce, create the "in" and "out" directories, copy a small ELF binary to the "in" directory, then:

```
cargo afl fuzz -i in -o out target/debug/testelf
```

pinkforest:
Cool no worries :) I created an issue in the repo: https://github.com/cole14/rust-elf/issues/23

Thanks a lot for the contrib in any case.

glg-rv:
Actually this refers to https://github.com/vincenthouyi/elf_rs rather than the rust-elf create.

cole14:
Yeah, I was just about to comment that this looks like it's fuzzing against a different crate. I'd love to see some fuzzing on the elf crate too, though :)

atulkharerivos:
@pinkforest Just noticed what @glg-rv pointed out...the crate in question is indeed elf_rs. I will run something against rust-elf as well :).

pinkforest:
Good catch - Thanks @glg-rv and @cole14 :+1: 

@vincenthouyi Ping

I raised https://github.com/vincenthouyi/elf_rs/issues/11

atulkharerivos:
Apologies for the confusion. I have updated the .md to point to the correct URL (`https://github.com/vincenthouyi/elf_rs`).

crates/elf_rs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,44 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "elf_rs"
+date = "2022-10-31"
+url = "https://github.com/vincenthouyi/elf_rs"
```
pinkforest, line 6:
```suggestion
url = "https://github.com/vincenthouyi/elf_rs/issues/11"
```

pinkforest:
@atulkharerivos would you mind sending a PR for elf-rs to fix it ?

https://github.com/vincenthouyi/elf_rs/issues/11#issuecomment-1300418210

Then the upstrean can roll a release we can point to in patched = []

pinkforest:
Ok. this has been open for a while with no fix whilst the maintainer has been made aware.

Merging and will update when there is a fix / if any in the future.

PR #1452, opened 2022-11-01
CVE-2022-3602 in openssl

amousset:


Shnatsel:
I see the fixed version is not yet up on crates.io. Did the maintainers confirm that this is the version in which it will ship, or is the version inferred from the versioning pattern?

amousset:
It is inferred from versioning scheme, the update and releases is waiting on the tag to be published.

We may want to wait until actual release before publishing the advisories.

Shnatsel (approved):

amousset:
[Published](https://crates.io/crates/openssl-src/versions)

Issue #1457, opened 2022-11-02
`atty` potential unaligned read


niluxv:
I found some discussions on a potential unaligned read in `atty` (10,564,450 recent downloads; 3 years since the last release). I think it legitimately looks like a soundness bug (though I guess it won't cause issues with the usual global allocators?). The issue was filed more than a year ago.

If I understand correctly, `FILE_NAME_INFO` boils down to a struct like
```
#[repr(C)]
struct FILE_NAME_INFO {
    FileNameLength: u32,
    FileName: [u16; 1],
}
```
which has align 4 ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=013c56cb97219d18bda98758824e31e0)). The vector `name_info_bytes` in <https://github.com/softprops/atty/blob/7b5df17888997d57c2c1c8f91da1db5691f49953/src/lib.rs#L131-L141> need not be 4 byte aligned (though with any sane allocator it will be?).

Links:
* [original issue on `atty`](https://github.com/softprops/atty/issues/50)
* [PR fixing the issue; same author as the issue](https://github.com/softprops/atty/pull/51)
* [discussion over at `clap`](https://github.com/clap-rs/clap/pull/4249)
* [`atty` maintenance status issue](https://github.com/softprops/atty/issues/57)

pinkforest:
Hey thanks :) Would you mind filing a PR with `informational = "unsound"` ?

To me it also looks like it's not maintained so we can go with that too :+1:

pinkforest:
People also at Wg-unsafe-guidelines agree that this indeed is unsound
https://rust-lang.zulipchat.com/#narrow/stream/136281-t-lang.2Fwg-unsafe-code-guidelines/topic/.60.23.5Bglobal_allocator.5D.60.20academic.20dilemma/near/311530432

Thanks to @thomcc who had picked up when this was being ported to std and pointed to:
https://github.com/rust-lang/rust/pull/98033#discussion_r958409731

Also thanks to @digama0 to also confirm this

pinkforest:
Resolved via #1462

PR #1462, opened 2022-11-04
Add `atty` potential unsoundness (unaligned read) advisory

niluxv:
See issue #1457.

@pinkforest Something like this?

### DIFF
--- a/crates/atty/RUSTSEC-0000-0000.md
+++ b/crates/atty/RUSTSEC-0000-0000.md
@@ -4,6 +4,7 @@ id = "RUSTSEC-0000-0000"
 package = "atty"
 date = "2021-07-04"
 url = "https://github.com/softprops/atty/issues/50"
+references = ["https://github.com/softprops/atty/pull/51", "https://github.com/softprops/atty/issues/57"]
 keywords = ["unaligned-read"]
 informational = "unsound"
 
@@ -19,3 +20,7 @@ patched = []
 On windows, `atty` dereferences a potentially unaligned pointer.
 In practice however, the pointer won't be unaligned unless a custom global allocator is used.
 In particular, the `System` allocator on windows uses `HeapAlloc`, which guarantees a large enough alignment.
+
+# atty is Unmaintained
+
+A Pull Request with a fix has been provided but the maintainer seems to be unreachable.
--- a/crates/atty/RUSTSEC-0000-0000.md
+++ b/crates/atty/RUSTSEC-0000-0000.md
@@ -1,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "atty"
+date = "2021-07-04"
+url = "https://github.com/softprops/atty/issues/50"
+keywords = ["unaligned-read"]
+informational = "unsound"
+
+[affected]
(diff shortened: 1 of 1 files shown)


crates/atty/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
...
+os = ["windows"]
+```
+
+# Potential unaligned read
+
+On windows, `atty` dereferences a potentially unaligned pointer.
+In practise however, the pointer won't be unaligned unaligned unless a custom global allocator is used.
+In particular, the `System` allocator on windows uses `HeapAlloc`, which guaranties large enough alignment.
+
+## References
```
pinkforest, line 20: We can remove these references here as it gets rendered via `url = []` and `references = []` - thanks :)

crates/atty/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
...
+package = "atty"
+date = "2021-07-04"
+url = "https://github.com/softprops/atty/issues/50"
+categories = ["unaligned-read"]
+informational = "unsound"
+
+[affected]
+os = ["windows"]
+```
+
```
pinkforest, line 13:
```suggestion
[versions]
patched = []

```

Needs this - it would be nice if the maintainer merges the fix before we merge this

crates/atty/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
...
+# Potential unaligned read
+
+On windows, `atty` dereferences a potentially unaligned pointer.
+In practise however, the pointer won't be unaligned unaligned unless a custom global allocator is used.
+In particular, the `System` allocator on windows uses `HeapAlloc`, which guaranties large enough alignment.
```
pinkforest, line 18:
```suggestion
In particular, the `System` allocator on windows uses `HeapAlloc`, which guarantees a large enough alignment.
```

crates/atty/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
...
+informational = "unsound"
+
+[affected]
+os = ["windows"]
+```
+
+# Potential unaligned read
+
+On windows, `atty` dereferences a potentially unaligned pointer.
+In practise however, the pointer won't be unaligned unaligned unless a custom global allocator is used.
```
pinkforest, line 17:
```suggestion
In practice however, the pointer won't be unaligned unless a custom global allocator is used.
```

pinkforest:
Thanks for the contribution - just couple of tiny cosmetic things please and one format fix :) thanks :heart: 

I'll leave this for a week open for anyone to comment as this will cause a lot of traffic

niluxv:
Thank you for the suggestions, and sorry for the delay! Addressed them now.

crates/atty/RUSTSEC-0000-0000.md
niluxv: Yeah, would be nice, but given that the last commit to `atty` was made more than a year ago and has many open PRs fixing issues (including this soundness issue), it seems unlikely to happen soon.

PR #1465, opened 2022-11-06
Add lzf `use-of-uninitialized-value` advisory

badboy:


pinkforest:
Thanks for the contribution :heart:

PR #1472, opened 2022-11-27
Patched slock

BrandonDyer64:
The [slock](https://crates.io/crates/slock) crate no longer has the issue described in the advisory


Shnatsel:
Thanks!

PR #1475, opened 2022-11-30
Add hyper-staticfile file disclosure on Windows

stephank:
This adds an advisory for https://github.com/stephank/hyper-staticfile/issues/35.

The contents are heavily based on an existing `tower-http` advisory: https://github.com/rustsec/advisory-db/blob/a66a3049c98395410a2afadf0382882b0a04d8b1/crates/tower-http/RUSTSEC-2022-0043.md


Shnatsel:
Thanks for the report!

I've fixed the version specification issue that caused CI to fail. Merging!

Issue #1474, opened 2022-11-30
Add new buffer overflow in capnproto-rust


Shnatsel:
https://github.com/capnproto/capnproto/blob/master/security-advisories/2022-11-30-0-pointer-list-bounds.md

PR #1476, opened 2022-11-30
add capnp bug: CVE-2022-46149

dwrensha:


dwrensha:
Closes #1474.

Shnatsel:
Thanks! Looks good to me, just one question - do you mind if I put https://github.com/capnproto/capnproto/blob/master/security-advisories/2022-11-30-0-pointer-list-bounds.md as the canonical URL, and add your post to the `references` field?

dwrensha:
Sure. There is also https://github.com/capnproto/capnproto/security/advisories/GHSA-qqff-4vw4-f6hx, which may be considered a more canonical URL.

PR #1480, opened 2022-12-07
Add soundness bug in `secp256k1` API

Kixunil:
Summary: Unsound API in `secp256k1` allows use-after-free and invalid deallocation from safe code. This was fixed and backported to multiple versions.

I'm not sure whether the `date` field should have todays date, will change if it needs something else.

Cc @apoelstra @tcharding

### DIFF
--- a/crates/secp256k1/RUSTSEC-0000-0000.md
+++ b/crates/secp256k1/RUSTSEC-0000-0000.md
@@ -2,7 +2,7 @@
 [advisory]
 id = "RUSTSEC-0000-0000"
 package = "secp256k1"
-date = "2022-12-07"
+date = "2022-11-30"
 
 url = "https://github.com/rust-bitcoin/rust-secp256k1/issues/543"
 references = ["https://github.com/rust-bitcoin/rust-secp256k1/pull/548"]
--- a/crates/secp256k1/RUSTSEC-0000-0000.md
+++ b/crates/secp256k1/RUSTSEC-0000-0000.md
@@ -1,0 +1,36 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "secp256k1"
+date = "2022-12-07"
+
+url = "https://github.com/rust-bitcoin/rust-secp256k1/issues/543"
+references = ["https://github.com/rust-bitcoin/rust-secp256k1/pull/548"]
+informational = "unsound"
+categories = ["memory-corruption"]
+keywords = ["use-after-free", "unsound-api", "invalid-free"]
+
+[affected]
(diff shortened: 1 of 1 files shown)


crates/secp256k1/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,39 @@
...
+url = "https://github.com/rust-bitcoin/rust-secp256k1/issues/543"
+references = ["https://github.com/rust-bitcoin/rust-secp256k1/pull/548"]
+informational = "unsound"
+
+categories = ["memory-corruption"]
+
+keywords = ["use-after-free", "unsound-api", "invalid-free"]
+
+[affected]
+
```
tcharding, line 16: nit: Could remove this space to be uniform. This section gets displayed when then rendered markdown on github so its kinda nice to have it uniform https://github.com/Kixunil/advisory-db/blob/secp256k1_preallocated_gen_new/crates/secp256k1/RUSTSEC-0000-0000.md

tcharding (approved):
I have zero experience with filing RUSTSEC advisories but FWIW I checked all links are valid and read over for grammar and correctness.

crates/secp256k1/RUSTSEC-0000-0000.md
Kixunil: Oh, yes, I deleted some other newlines too to make it a bit shorter without affecting readability too much.

Shnatsel:
Looks good to me, thank you!

Is this easy to trigger unintentionally? If so, we might want to drop `informational = "unsound"` which would promote it from a warning to an error. What do you think?

Kixunil:
@Shnatsel hard to say, it's possible that the specific function isn't even used at all in any outside code. Neither github search nor grep.app show anything interesting so if it's used anywhere it's probably closed source.

However it kinda bypasses borrow checking so the consequences are probably quite bad. If `cargo-audit` can already filter based on function and does not error if the function is not called, I'd keep `unsound`. If it unconditionally errors for everyone I'd drop it.

apoelstra:
>Is this easy to trigger unintentionally?

I don't think so -- you need to be using a special no-alloc version of our API (which already excludes the vast majority of users), and then you need to pass it a backing store with a finite lifetime (which would be a little weird for a backing store), and then you'd need to use the object from our API beyond the backing store's lifetime (which would likely be caught in code review -- unless the reviewer was working from a "Rust will protect us from such things" mental model -- and if not, would likely reliably cause a double-free crash).

Kixunil:
> unless the reviewer was working from a "Rust will protect us from such things" mental model

I think not wasting brain power on reviewing those things is part of the point of Rust so my guess is most people do. (Including myself; and you know how crazy my reviews are... ;))

tarcieri:
Based on @apoelstra's comment this sounds like `informational = "unsound"` to me

Shnatsel:
Yep, sounds good. Thanks for the explanation. Merging!

PR #1493, opened 2022-12-15
Add unsound `ostree`

shinmao:


shinmao:
The method to process the results of `catch_unwind` lead to unsoundness of unwinding-handling function:

1. When C calls `filter_trampoline_unwindsafe` function, developers used `catch_unwind` to avoid the panics in `filter_trampoline` propagating to C function.  However, it was still possible for `eprintln!` from `print_panic` to lead to panic outside the scope of `catch_unwind`, which was undefined behavior even in current nightly version.
2. Basically, it could break the functionality of `print_panic`.

Link: https://github.com/ostreedev/ostree/issues/2775

crates/ostree/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "ostree"
+date = "2022-11-18"
+url = "https://github.com/ostreedev/ostree/issues/2775"
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:C/C:N/I:N/A:L"
+categories = ["memory-corruption"]
```
pinkforest, line 8: Could we please explain how this would lead to memory corruption ?

crates/ostree/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "ostree"
+date = "2022-11-18"
+url = "https://github.com/ostreedev/ostree/issues/2775"
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:C/C:N/I:N/A:L"
```
pinkforest, line 7: How was this cvss constructed given the impact is undefined ?

pinkforest:
Hi. Thanks for the contribution. This would need more details - especially if memory corruption is suggested - would you have minimally reproducible code samples given the soundness issues would require contrieved code to trigger.

shinmao:
> Hi. Thanks for the contribution. This would need more details - especially if memory corruption is suggested - would you have minimally reproducible code samples given the soundness issues would require contrieved code to trigger.

```rust
pub(super) unsafe extern "C" fn filter_trampoline_unwindsafe(
    ...,
) -> ffi::OstreeRepoCheckoutFilterResult {
    ...
    let result = catch_unwind(move || filter_trampoline(repo, path, stat, user_data));
    result.unwrap_or_else(|panic| {
        print_panic(panic);
        abort()
    })
}

fn print_panic(panic: Box<dyn Any>) {
    eprintln!("A Rust callback invoked by C code panicked.");
    eprintln!("Unwinding across FFI boundaries is Undefined Behavior so abort() will be called.");
    let msg = {
        ...
    };
    eprintln!("Panic value: {}", msg);
}
```
Hi @pinkforest , thanks for the response. For memory corruption and cvss, I think I have a misunderstanding when I refer to other similar issues. It is not type of memory corruption, but type of panic-safety bug. The above is the workflow of the unsound function. `filter_trampoline_unwindsafe` is a FFI function which could invoke by C. Developers used `catch_unwind` to avoid the case that `filter_trampoline` has panic and propagation, and developers would print out error message with `print_panic` for the case. Based on the doc of [eprintln](https://doc.rust-lang.org/std/macro.eprintln.html), it is possible for panic to happen with `io::stderr` failure.

This incorrect panic handling leads to the unsoundness of `filter_trampoline_unwindsafe`: `print_panic` could not work normally. Program should abort when panic, but panic in `print_panic` would lead to interaction between C stack frame and Rust unwinding which is still UB.

crates/ostree/RUSTSEC-0000-0000.md
shinmao: Could you help me to delete the fields of `cvss` and `categories`? I have some misunderstanding on the definition. Or I can delete them and resend a pull request?

pinkforest:
Hiya. The example for a tricky case like this ideally should be something that demonstrates / can happen by using the crate and more ideally repro's the situation if possible - if we are using any education re: panic safety not just the code that is there in which just confuses most people who read the advisories.

Please add any of the example/s or descriptions into the markdown. You can just modify a file, git add, git commit and git push to your fork branch - the same goes for clarifying that this has nothing to with memory safety.

shinmao:
HI, thanks for the detailed response. I think I would not be able to include repro in this report. The reported part is not the external api which could be called by the users of libraries. I reported this bug just because I thought the design to catch panic with `catch_unwind` is not **sound** (scanned by static analysis). However, it's true that this behavior with this scenario couldn't cause to any vulnerabilities such as memory safety issues. I think I need a more **concrete** case to show that this **unsoundness could introduce a bug**, and this is what I miss in this case (Hope my understanding is correct). If you agree, I can close my pull request!

pinkforest:
Informational advisories don't need concrete vulns -

There was a suggestion in the zulip thread how it could be triggered by using output redirect /dev/full

I mean it would be nice - it's not generally required but it would be real nice to have given the educational value.

shinmao:
> There was a suggestion in the zulip thread how it could be triggered by using output redirect /dev/full

So what I miss here is to trigger panic in `eprintln!()` and show what would happen?

pinkforest:
When there is a saying that it doesn't need concerete vulns - this is usually in the case of library means there is no consumer bin currently using the library that can implicate the library itself where the soundness bug is.

This is usually the way for us to flag libraries that have some certain soundness issues yet we can't entirely pinpoint the binary consumer that is affected - thus concrete vulnerability which tend to exist in bin side only.

Thinking this now from that PoV since there is / are no suggested use scenarios - real / hypothetical - how this exactly can be realistically triggered or any description(s) together with hypothetical real impact(s) what could / would happen if there would be a hypothetical consumer binary usign this library e.g. using it in the wrong way -

I just don't see or maybe my imagination is just broken how can we merge this given this is not mapped out from consumer side.

Therefore I recommend closing this for now and / or come up with the details that would be required to have properly mapped out soundness issue from library PoV where this is consumed hypothetically.

We have to consider on the basis of value what the advisory can archieve given both the mandate as advisory database and where the maintainer @cgwalters & @lucab have helpfully fixed the bug where our advisory can send the wrong signals to the ecosystem if it's not done throughfully to map out both the issue, potential consumer scenario(s) and real potential impact given when the said consumer scenario(s) mapped out.

Anyways thank you again for contributing and I hope this does not dissuade you from contributing again.

The balance and completeness here is really tricky to get right and I appreciate for the patience.

Also if someone is keen to add input / details for completeness this could be certainly helpful here.

cgwalters:
There's only one [direct public dependency](https://crates.io/crates/ostree/reverse_dependencies) and I maintain that too and it doesn't use the affected API.  (Even if it did, the scenario of needing writing to stderr to fail makes this very much a corner case)

Having an advisory would likely trigger a lot of unnecessary churn.

But all that said I'm still conflicted because...well, I can see a bit of value in being a warning for other crates which also try to `catch_unwind` like this.

pinkforest:
Thanks for the input - agreed - re-open if / when there is a /are specific consumer scenario/s to think about and evaluate.

PR #1494, opened 2022-12-16
Add unmaintained `rusoto_credential`

iliana:
Hi. I'm ostensibly one of the maintainers of Rusoto, here to report its lack of maintenance.

[Rusoto has been deprecated for a while](https://github.com/rusoto/rusoto/issues/1651#issuecomment-1108012975); this is a more formalized notice of such to help churn downstream automation to make it clear that users need to migrate to aws-sdk-rust.

This is a somewhat unique case for the RustSec advisory DB, I think. The Rusoto project is made up of a little over 200 separate crates, but every crate in the tree either is or depends on `rusoto_credential`. Most users don't directly depend on `rusoto_credential` though, so I feel that listing an advisory for only that crate is a bit confusing, but hopefully downstream systems show the full description for context.

### DIFF
--- a/crates/rusoto_credential/RUSTSEC-0000-0000.md
+++ b/crates/rusoto_credential/RUSTSEC-0000-0000.md
@@ -1,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rusoto_credential"
+date = "2022-04-24"
+url = "https://github.com/rusoto/rusoto/issues/1651"
+informational = "unmaintained"
+
+[versions]
+patched = []
+```
+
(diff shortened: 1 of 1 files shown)


pinkforest:
Thanks a lot for rusoto - it was good as long as it lasted :heart: 

And thanks for contribution and keeping everyone up-to-date -

And yup lightning up the common crate is the way we've gone :)

kennethuil:
aws_sdk_rust still has this note in its repo: 

Please Note: The SDK is currently released as a developer preview and is intended strictly for feedback purposes only. Do not use this SDK for production workloads.

So that leaves us with no recommended way to use AWS APIs in production?

PR #1499, opened 2022-12-23
Add open redirect issue in hyper-staticfile

stephank:
This adds an advisory for an open redirect issue discovered in `hyper-staticfile`. Hopefully it's okay that this doesn't have a URL attached. There's no issue for this in the issue tracker; it was reported to us directly by a security partner analyzing an application using this crate.


alex:
Can you use the resolving commit as the URL?

stephank:
I added a link for the commit on the alpha branch, because it's a single commit there. On the current stable, I had to do another fixup, because Windows support broke in a way CI didn't catch. 😑

alex (approved):

pinkforest:
Thanks for the contribution :tada:

PR #1503, opened 2022-12-27
Add unsound `prettytable-rs`

david0u0:


crates/prettytable-rs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "prettytable-rs"
+date = "2022-12-27"
```
pinkforest, line 5: The date is set when it was originally reported where we align with other databases doing this backdating.

pinkforest:
Thanks, will merge today/tomorrow after 1.0..0 is out.

crates/prettytable-rs/RUSTSEC-0000-0000.md
david0u0: fixed

pinkforest:
fyi - ended up releasing 0.10.0 - fixed tihs one to refer to 0.10.0

PR #1507, opened 2023-01-06
Add License CC-BY 4.0 Import exemption

pinkforest:
Allows importing advisories with attribution from GHSA

EDIT: This is so that it will be ready for the tooling https://github.com/rustsec/rustsec/pull/682

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -13,7 +13,7 @@ repository, even for commercial purposes, without asking permission.
 Additional content from GitHub Security Advisory ("GHSA") database
 
 Additional content may be adapted from GHSA with attribution requirements, but
-no additional clauses like copyleft.
+with no additional clauses like copyleft.
 
 Any such license and attribution will be explicitly covered on an advisory by
 advisory basis directly within the applicable advisories.
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,5 +1,5 @@
-All code and data in the RustSec advisory database repository is dedicated
-to the public domain:
+All code and data in the RustSec advisory database repository is dedicated to
+the public domain:
 
 https://creativecommons.org/publicdomain/zero/1.0/
 
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,8 +1,3 @@
-Unless otherwise explicitly mentioned in any individual advisories imported
-from other security advisory databases generally as CC-BY 4.0:
-
-https://creativecommons.org/licenses/by/4.0/
-
 All code and data in the RustSec advisory database repository is dedicated
 to the public domain:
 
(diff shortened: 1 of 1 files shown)


Shnatsel:
I think we will need a more clear way to distinguish what data is under what license. For example a `license` field, or an explicitly specified attribution link since CC-BY requires attribution.

pinkforest:
Yes we cna do it after we have the license but that will require tooling changes.

In the meantime we can simply do a source attribution that seems to be the way nvd does it.

Shnatsel:
That is, I agree with the basic direction, but the implementation will be more complex than this.

FWIW I've already tried talking to Github about a licensing exception for RustSec and they said they aren't going to do it.

pinkforest:
Yeah I get the tooling is complex but having the license exemption approved & ready is a good start I would think ?

Shnatsel:
What I think could be valuable is a note that we're looking into adding advisories under CC-BY in the future, to ensure we can import from GHSA.

pinkforest:
But is that a license exemption issue that needs to be in license.txt ? This is just a legal foundational exemption blurb.

I've noted that importing which can generally by in CC-BY 4.0 term that is defined in the license.txt now

pinkforest:
For anyone ICYMI the tooling change PR is here: https://github.com/rustsec/rustsec/pull/682

tarcieri (approved):
This seems ok as a start, although it would be good to have a proper attribution for imported advisories with structured data about the license

Shnatsel (changes requested):
The data up until now is exclusively public domain, and I think it's important to capture that.

Please rephrase this to state that CC-BY (or other) content may be added in the future, and that it only applies to licenses with attribution requirements but no additional clauses like copyleft.

Shnatsel (approved):

LICENSE.txt
```diff
@@ -9,3 +9,12 @@ the extent allowed by law.
 
 You can copy, modify, distribute, and retransmit any information in this
 repository, even for commercial purposes, without asking permission.
+
+Additional adapted content from GitHub Security Advisory ("GHSA") database
+
+CC-BY (or other) additional content may be adapted from GHSA that only applies
+to licenses with attribution requirements but no additional clauses like
+copyleft.
```
alex, line 17: I'm having a bit of trouble parsing this sentence. I think there might be a missing or extra word?

pinkforest: How about now?

alex (approved):

amousset (approved):

amousset:
This is a good first step, the next one is probably to reach a consensus on how to store this information in a structured way (in https://github.com/rustsec/rustsec/pull/682).

Issue #1483, opened 2022-12-15
`wasmtime` CVEs


pinkforest:
CVE-2022-39392: wasmtime - out of bounds read/write with zero-memory-pages configuration
- https://github.com/rustsec/advisory-db/pull/1508

CVE-2022-39393: wasmtime - may have data leakage between instances in the pooling allocator
- https://github.com/rustsec/advisory-db/pull/1509

pinkforest:
@alexcrichton Could we please re-release the GHSA additional content as CC0 ? GHSA has CC-BY 4.0

alexcrichton:
Sorry but I'm not sure how that would be done? It's fine to do so, but I don't know what I could do to classify it as such or something like that.

pinkforest:
Yeah cool your reply here covers that - thanks a lot :)

pinkforest:
Completed via #1508 and #1509

PR #1509, opened 2023-01-09
Add `wasmtime` CVE-2022-39393

pinkforest:
Re: https://github.com/rustsec/advisory-db/issues/1483

https://nvd.nist.gov/vuln/detail/CVE-2022-39393


Shnatsel (approved):
LGTM.

I understand only information from NVD is used, so this information is in the public domain.

PR #1534, opened 2023-01-18
Add patched info for RUSTSEC-2022-0079 (elf_rs)

jschwe:
The issue was confirmed to be fixed in v0.3.0 by the person who reported the issue: https://github.com/vincenthouyi/elf_rs/issues/11#issuecomment-1384624749


pinkforest:
Thanks for the contribution :tada:

PR #1535, opened 2023-01-18
Add pointers to Rust-lang security policy

pinkforest:
- Issue template to point to the Security policy
- Top level note in README to point to the Security policy

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yaml
+++ b/.github/ISSUE_TEMPLATE/config.yaml
@@ -1,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Rust-lang Project Security Policy
+    url: https://www.rust-lang.org/policies/security
+    about: Please report issues about rustc and cargo to security@rust-lang.org.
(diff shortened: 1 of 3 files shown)


cuviper:
This is not the place for first security contact *in general*, right? It's great to mention the security policy for the Rust Project, but others in the ecosystem have their own policies as well, e.g. [tokio](https://github.com/tokio-rs/tokio/security/policy). So I think the gist of the message should be something like, "Here's the Rust Project policy, and for others please check if they have their own policy."

pinkforest:
Sure - For this we already have the very clear wording for that in the how-to that the maintainer is the first point of contact and where we have played the essential role of bridging the maintainers with the repoters.

There is nothing in the git2 repository about any security policy what so ever: https://github.com/rust-lang/git2-rs

By saying see if git2-rs has a security policy none of this would have helped unless we specify that every rust-lang repository is looked after by the Security Response WG.

pinkforest:
I further adjusted the wording about responsible disclosure that includes woring about looking at the security policies.
It would be great if people use SECURITY.md more and docs.rs crates.io would include that as a standard we could refer to instead of some GitHub-only proprietary information.

PR #1536, opened 2023-01-18
Report json as unmaintained

ecton:
A thread on Reddit yesterday led me to investigate a few different JSON crates in the ecosystem. One of the most popular ones beyond serde-json that reported fast benchmarks is [`json`](https://crates.io/crates/json). I investigated it and discovered [users have asked](https://github.com/maciejhirsz/json-rust/issues/205) if the project is abandoned without reponse. Given how popular this crate appears to be given the recent downloads, I felt an unmaintained advisory was warranted.

Scanning the issues list, I also found [this issue](https://github.com/maciejhirsz/json-rust/issues/196) that may introduce undefined behavior. This triggers a Miri error under the stacked borrows model, but I'm not sure if that's enough to warrant a mention in the advisory or not.

### DIFF
--- a/crates/json/RUSTSEC-0000-0000.md
+++ b/crates/json/RUSTSEC-0000-0000.md
@@ -17,7 +17,7 @@ Last release was almost 3 years ago.
 
 The maintainer is unresponsive with outstanding issues.
 
-These outstanding issues include [a soundness one](https://github.com/maciejhirsz/json-rust/issues/196).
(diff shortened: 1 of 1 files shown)


pinkforest:
Hey ecton thanks a lot for the contribution :) I gotta just align the date to when it's reported as we align to to other security database standard on that + refer to the soudness issue :partying_face:

PR #1538, opened 2023-01-20
MITM attack against dfile

alexanderkjall:


pinkforest:
Thanks, libgit2-sys advisory will flag this one up as well so no need for a separate one.

PR #1539, opened 2023-01-20
MITM attack against gitsync

alexanderkjall:


pinkforest:
Thanks, libgit2-sys advisory will flag this one up as well so no need for a separate one.

Shnatsel:
I believe https://github.com/rustsec/advisory-db/pull/1540 should cover this, which is filed against `libgit2-sys` package.

If that's not the case, please let us know and we'll reopen the PR.

Thanks for the report!

PR #1542, opened 2023-01-21
Remove unmaintained status of `markdown`

wooorm:
Related-to: GH-1191.

### DIFF
--- a/crates/markdown/RUSTSEC-2022-0044.md
+++ b/crates/markdown/RUSTSEC-2022-0044.md
@@ -7,7 +7,7 @@ informational = "unmaintained"
 url = "https://github.com/johannhof/markdown.rs/issues/48"
 
 [versions]
-patched = [">= 1.0.0-alpha.1]
+patched = [">= 1.0.0-alpha.1"]
 ```
 
 # `markdown` (1.0.0 and higher) is maintained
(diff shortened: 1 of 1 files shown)


wooorm:
lgtm :)

pinkforest:
Thanks for the contribution :partying_face: 

I think we'll just mark from 1.0.0-alpha.1 as resolved so it will note on anything stuck at 0.3.0 so that will incentivise upgrading - widthdrawn wouldn't continue note 0.3.0 users who are using the unmaintained one - lmk if any issues with this - in any case the advisories never really get deleted but marked with these meta tags.

Cheers

wooorm:
oh that sounds better (though I don‘t know much about all these tools!)

pinkforest:
The ones I tend to recommend:
- [cargo deny](https://crates.io/crates/cargo-deny) - for PR's
- [cargo audit](https://crates.io/crates/cargo-audit) - for cron style checks and
- dependabot from GitHub to send PR's for any dep bumps :)

as an example:
- https://github.com/lunatic-solutions/lunatic/blob/main/.github/dependabot.yml
- https://github.com/lunatic-solutions/lunatic/blob/main/.github/workflows/cargo-audit.yml
- https://github.com/lunatic-solutions/lunatic/blob/main/.github/workflows/cargo-deny-pr.yml

PR #1544, opened 2023-01-28
Add advisory for warp#937

### LINKED ISSUE #1524: `warp` missing advisory
I just realized that warp is missing an advisory for the issue similar to https://github.com/rustsec/advisory-db/pull/1159. The tracking issue for this bug was https://github.com/seanmonstar/warp/issues/937, and it was fixed by https://github.com/seanmonstar/warp/pull/997, which was released as 0.3.3. I think it is important to carry an advisory as it appears that the fix was not backported to the 0.2.x or 0.1.x branches. The maintainer was asked to file an advisory in https://github.com/seanmonstar/warp/issues/937, but it looks like they forgot to after fixing the issue. What are the next steps to be taken?

adumbidiot:
Fixes #1524. I put the date as 2023, since that is the creation date of this advisory. However, this security issue was filed back in 2022. Let me know if I should change the dates.

### DIFF
--- a/crates/warp/RUSTSEC-0000-0000.md
+++ b/crates/warp/RUSTSEC-0000-0000.md
@@ -1,0 +1,24 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "warp"
+date = "2023-01-28"
+url = "https://github.com/seanmonstar/warp/issues/937"
+categories = ["file-disclosure"]
+keywords = ["directory traversal", "http"]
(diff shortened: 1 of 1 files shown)


pinkforest:
Hey thanks a lot - yep could we please use the date when it was reported originally as we align with the other advisory databases standard on that

adumbidiot:
Updated the date

adumbidiot:
Thanks

PR #1553, opened 2023-02-01
Add unmaintained `multi_mut`

### LINKED ISSUE #1531: `multi_mut` is unsound
https://github.com/golddranks/multi_mut has been written before Rust had an aliasing model, and before `&mut` had `noalias` annotations in LLVM. It violates Stacked Borrows and is likely UB under current LLVM behavior:

https://github.com/golddranks/multi_mut/issues/1

pinkforest:
Fixes #1531 

It has no dependants so I think unmaintained can do it ?

Ping: https://github.com/rust-lang/rust/issues/39155

### DIFF
--- a/crates/multi_mut/RUSTSEC-0000-0000.md
+++ b/crates/multi_mut/RUSTSEC-0000-0000.md
@@ -5,7 +5,7 @@ package = "multi_mut"
 date = "2020-02-07"
 url = "https://github.com/golddranks/multi_mut/issues/1"
 references = ["https://github.com/rust-lang/rust/issues/39155", "https://www.reddit.com/r/rust/comments/5ofuun/multi_mut_multiple_mutable_references_to_hashmap/"]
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thanks for putting this together!

I'd consider "unsound" a stronger warning than "unmaintained", so I'd prefer to go with that. "Unmaintained" was originally created for a crate with no known issues but maintainers MIA, `rustcrypto`.

Any objections? Should we update the maintainer's guide with that?

pinkforest:
tbh we could have combined status for unmaintained and suspect unsound we can't get the maintainer input on ?

Shnatsel:
I think an advisory without a patched version is a strong enough signal that the crate isn't maintained well.

@tarcieri thoughts?

tarcieri:
You can file separate advisories for both issues

kornelski (approved):

PR #1579, opened 2023-02-07
Add CVE-2023-0215 for openssl-src

amousset:


Shnatsel (approved):
FWIW doesn't seem to be exposed through the `openssl` crate, at least at a glance

PR #1591, opened 2023-02-08
Update RUSTSEC-2020-0071.md

DK26:
I found this solution to work for me and I hope it is correct as cargo-audit stopped complaining about this vulnerability once this workaround is applied.

### DIFF
--- a/crates/time/RUSTSEC-2020-0071.md
+++ b/crates/time/RUSTSEC-2020-0071.md
@@ -71,7 +71,7 @@ Users of time 0.1 do not have a patch and should upgrade to an unaffected versio
 
 ### Workarounds
 
-A possible workaround is to avoid using the default `oldtime` feature dependency of the `chrono` crate by disabling its `default-features` and manually specifying the required features instead.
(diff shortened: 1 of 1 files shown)


crates/time/RUSTSEC-2020-0071.md
```diff
@@ -71,4 +71,26 @@ Users of time 0.1 do not have a patch and should upgrade to an unaffected versio
 
 ### Workarounds
 
-No workarounds are known.
+A possible workaround is to avoid using the default `oldtime` feature dependency of the `chrono` crate by disabling its `default-features` and manually specifying the required features instead.
```
amousset, line 74: As this is an advisory for the `time` crate I think this should explain that the workaround works for crates affected only through the transitive dependency in `chrono`.

DK26: How about:

"A possible workaround for crates affected through the  transitive dependency in `chrono`, is to avoid using...[same as now]" ?

amousset: :+1:

PR #1595, opened 2023-02-08
Add advisory for pnet_packet

### LINKED ISSUE #1473: Buffer overflow in old versions of `pnet_packet`
[MutableTcpPacket::set_payload](https://docs.rs/pnet_packet/0.22.0/src/pnet_packet/home/cratesfyi/cratesfyi/debug/build/pnet_packet-5e25de1d4dfd4299/out/tcp.rs.html#705-714) does not validate if the payload fits the packet buffer it writes into. In 0.22 the code looks like the following:

    pub fn set_payload(&mut self, vals: &[u8]) {
        use std::ptr::copy_nonoverlapping;
        let mut _self = self;
        let current_offset = 20 + tcp_options_length(&_self.to_immutable());
        unsafe {
            copy_nonoverlapping(vals[..].as_ptr(),
                                _self.packet[current_offset..].as_mut_ptr(),
                                vals.len())
        }
    }

Later versions generate this function via a proc macro, but the vulnerability is still there. There are other functions affected by the same vulnerability, since this macro seems to be used for all byte-array fields.

A correct caller will ensure that the buffer is big enough, and fixed versions will panic if this condition isn't satisfied. So I'm not sure if this should be marked as an actual vulnerability or just a soundness issue.

Fixed in https://github.com/libpnet/libpnet/pull/455 which went into `v0.27.0`. The issue is already present in the first version released on crates.io, `v0.20.0`.

----

[It looks like](https://github.com/libpnet/libpnet/issues/449#issuecomment-663355987) this this is the underlying soundness issue behind [RUSTSEC-2019-0037](https://github.com/rustsec/advisory-db/blob/main/crates/pnet/RUSTSEC-2019-0037.md) as mentioned in https://github.com/rustsec/advisory-db/pull/335/files. Since this is a different crate that can be used independently, it's probably still a good idea to have separate advisory.

sidunder:
Closes #1473 

I took many of the metadata details from [RUSTSEC-2019-0037](https://github.com/rustsec/advisory-db/blob/main/crates/pnet/RUSTSEC-2019-0037.md) as #1473 (and the original [issue](https://github.com/libpnet/libpnet/issues/449)) indicates that this is its root cause.

### DIFF
--- a/crates/pnet_packet/RUSTSEC-0000-0000.md
+++ b/crates/pnet_packet/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "pnet_packet"
+date = "2020-06-19"
+url = "https://github.com/libpnet/libpnet/issues/449"
+categories = ["memory-corruption"]
+cvss = "CVSS:3.0/AV:L/AC:L/PR:H/UI:N/S:U/C:N/I:H/A:H"
+
+[versions]
+patched = [">= 0.27.2"]
+```
+
+# `pnet_packet` buffer overrun in `set_payload` setters
+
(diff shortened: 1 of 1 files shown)


Shnatsel:
Looks good. Thanks!

PR #1612, opened 2023-02-20
Add unmaintained ftp crate

veeso:
Report for unmaintained `ftp` crate. Use `suppaftp` instead

### DIFF
--- a/crates/ftp/RUSTSEC-0000-0000.md
+++ b/crates/ftp/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "ftp"
+date = "2023-02-20"
+url = "https://github.com/mattnenterprise/rust-ftp/pull/92"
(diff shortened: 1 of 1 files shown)


pinkforest (commented):
Usually if the upstream dependency is insecure then the advisory would belong to the upstream dependency and this way any downstream dependency gets flagged as well making any other advisory redundant - as otherwise we would be raising a lot of effectively duplicate advisories.

However seeing it uses ^0.10 openssl there shouldn't be an issue since OpenSSL 1.1.1 tends to be secure by itself if properly patched (caveats apply) - there is relatively nothing wrong using 1.1.1 over 3.0 at the time of writing.

I understand that some Linux distros have started providing libs only on OpenSSL 3.0 series but that isn't security issue alone that could be flagged via downstream crate or even via OpenSSL given it's considered secure (as above)

Problem is that the maintainer has advised that they would be actively maintaining the crate then we need either explicit communication that the maintainer is not planning to maintain the crate or where the maintainer is completely unreachable for 90 days.

I've pinged the maintainer re: maintenance status enquiry and see what they come back with.

veeso:
@pinkforest 90 days have elapsed. Can rust-fpt be marked as unmaintained now?

Issue #1602, opened 2023-02-14
buf_redux might be unmaintained


oherrala:
Ping the repository owner and maintained @abonander. No reaction to any open pull request or issues since 2019 in the repository. Last crates.io release also from 2019.

My question about maintenance status has been unanswered so far https://github.com/abonander/buf_redux/issues/22

And there's also future incompatibility warning https://github.com/abonander/buf_redux/issues/23.

Ltrlg:
According to GitHub, @abonander archived the `buf_redux` repository yesterday, as well as others:

  - [`multipart`](https://github.com/abonander/multipart): 55 dependents according to crates.io
  - [`multipart-async`](https://github.com/abonander/multipart-async): No dependents according to crates.io
  - [`const-cstr`](https://github.com/abonander/const-cstr): Not the one on crates.io
  - [`safemem`](https://github.com/abonander/safemem): 12 dependents according to crates.io, including `multipart`.

At least `multipart` should probably get an unmaintained notice too.

oherrala:
I created unmaintained advisory PRs for 
 * const-cstr: https://github.com/rustsec/advisory-db/pull/1613
 * safemem: https://github.com/rustsec/advisory-db/pull/1615
 * buf_redux: https://github.com/rustsec/advisory-db/pull/1614

dignifiedquire:
Created a fork of `buf_redux` which fixed the issues: https://crates.io/crates/buffer-redux, which should be fully API compatible.

Issue #1438, opened 2022-10-16
`multipart` Status


pinkforest:
6,789,801 downloads all time, ~12k a day.

Whilst going through `iron` - https://github.com/rustsec/advisory-db/issues/1424 - and some other crates around the space I forgot to flag `multipart` to clarify it's status.

- https://crates.io/crates/multipart - Release over a year ago
- https://github.com/abonander/multipart

@abonander - Just wondering is `multipart` still being actively maintained or would be it be deprecated ?

These issues caught my eye - some of the I think are `hyper` a bit like in `iron`:

- https://github.com/abonander/multipart/issues/143 - 
- There is open PR on `twoway` - https://github.com/rustsec/advisory-db/pull/1435
- https://github.com/abonander/multipart/issues/139 - "\r\n" straddling the buffer end (sounds like outdated hyper crate)
- https://github.com/abonander/multipart/issues/131 - Got MissingContentDisposition Error
- https://github.com/abonander/multipart/issues/122 - Panic
- https://github.com/abonander/multipart/issues/106

Hyper/Iron are optional dependencies though -

```toml
hyper = { version = ">=0.9, <0.11", optional = true, default-features = false }
iron = { version = ">=0.4,<0.7", optional = true }
```

Considering `hyper` has several advisories - some related to the above.

Normally this wouldn't be an issue with the optional deps as the advisory pops up via hyper picked up -

But I'm just wondering whether this all maintained considering outdated deps e.g. hyper w/ advisories -

**There is no upgrade path to hyper current track 0.14**
**Issue essentially seems that the crate forces to use very old version of hyper from 0.11 track from 5 yrs ago ?**

**NOTE: Some of these MAY NOT be applicable - No further analysis has been done yet**
**NOTE.2: Maintainer is the first point to verify these issues whether affected or not**

**Parser creates invalid uninitialized value - >= 0.14.12**
https://github.com/rustsec/advisory-db/blob/main/crates/hyper/RUSTSEC-2022-0022.md

**Integer overflow in hyper's parsing of the Transfer-Encoding - >= 0.14.10**
https://github.com/rustsec/advisory-db/blob/main/crates/hyper/RUSTSEC-2021-0079.md

**Lenient hyper header parsing of Content-Length - >= 0.14.10**
https://github.com/rustsec/advisory-db/blob/main/crates/hyper/RUSTSEC-2021-0078.md

**Unaffected - Only between 0.12.0 < 0.14.3** 
[...]

PR #1613, opened 2023-02-20
Add unsound `const-cstr`

oherrala:
const-cstr is crate by @abonander who has archived bunch of his Rust crates in GitHub.

See communication attempts on other crates:
 * https://github.com/rustsec/advisory-db/issues/1602
 * https://github.com/rustsec/advisory-db/issues/1438

### DIFF
--- a/crates/const-cstr/RUSTSEC-0000-0000.md
+++ b/crates/const-cstr/RUSTSEC-0000-0000.md
@@ -36,5 +36,5 @@ This is however unlikely but the the crate should not be used for untrusted data
 
 The below may or may not provide alternative(s)
 
-- [const_cstr::cstr!](https://docs.rs/const-str/latest/const_str/macro.cstr.html)
-- [cstr::cstr!](https://docs.rs/cstr/latest/cstr)
+- [const_str::cstr!](https://docs.rs/const-str/latest/const_str/macro.cstr.html)
+- [cstr::cstr!](https://crates.io/crates/cstr)
--- a/crates/const-cstr/RUSTSEC-0000-0000.md
+++ b/crates/const-cstr/RUSTSEC-0000-0000.md
@@ -2,7 +2,7 @@
 [advisory]
 id = "RUSTSEC-0000-0000"
 package = "const-cstr"
-date = "2023-02-14"
+date = "2023-03-12"
 url = "https://github.com/abonander/const-cstr"
 informational = "unsound"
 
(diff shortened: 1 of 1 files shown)


Nugine:
Alternatives:
+ [`const_str::cstr!`](https://docs.rs/const-str/latest/const_str/macro.cstr.html): decl macro + const evaluation
+ [`cstr::cstr!`](https://docs.rs/cstr/latest/cstr/): proc macro

pinkforest:
btw `const_cstr` doesn't check for any interior nul bytes

Both the constructor and the canonical macro allows this:

```
const_cstr! {
    FOO = "\u{0}💖\u{0}";
}

let foo = FOO.as_cstr();
```

This violates the safety contract of [ffi::CStr::from_bytes_with_nul_unchecked](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_with_nul_unchecked) that is used in `as_cstr()`

[ConstCStr::as_cstr](https://docs.rs/const-cstr/latest/const_cstr/struct.ConstCStr.html#method.as_cstr) assert only checks the last byte is nul that is relied for safety:
https://github.com/abonander/const-cstr/blob/master/src/lib.rs#L122

Also because unicode(tm) vs bytes - as_str goes:

`"\x00\xf0\x9f\x92\x96\x00"` vs  `ConstCStr { val: "\0💖\0" }`

pinkforest:
Ok I've filled the advisory with some details.

crates/const-cstr/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,40 @@
...
+
+Additionally the crate may cause runtime panics if statically compiled and ran with any untrusted data that is not nul-terminated.
+
+This is however unlikely but the the crate should not be used for untrusted data in context where panic may create a DoS vector.
+
+## Possible Alternatives
+
+The below may or may not provide alternative(s)
+
+- [const_cstr::cstr!](https://docs.rs/const-str/latest/const_str/macro.cstr.html)
```
Nugine, line 39: `const_str::cstr!`

oherrala:
@pinkforest LGTM 👍

crates/const-cstr/RUSTSEC-0000-0000.md
pinkforest: Oh oops thanks :purple_heart:

PR #1617, opened 2023-02-20
Add informational advisory for partial_sort out-of-bounds read

niklasf:
Should not really be a vulnerability, because callers would have to validate the parameter anyway to avoid panics, but I figured I'd file an informational advisory regarding https://github.com/sundy-li/partial_sort/issues/7 anyway, now that a fixed version is released.

cc @sundy-li

### DIFF
--- a/crates/partial_sort/RUSTSEC-0000-0000.md
+++ b/crates/partial_sort/RUSTSEC-0000-0000.md
@@ -1,0 +1,29 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "partial_sort"
+date = "2023-02-20"
+url = "https://github.com/sundy-li/partial_sort/issues/7"
+informational = "unsound"
+categories = ["memory-exposure"]
+keywords = ["out-of-bounds read"]
+
+[versions]
+patched = [">= 0.2.0"]
+```
+
+# Possible out-of-bounds read in release mode
+
+Affected versions of this crate were using a debug assertion to validate the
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you for the report, and sorry for the delay!

I'd like to understand the issue better before we make the judgement of making it informational or a full-on advisory. I understand that passing a very large number as a first argument to `partial_sort` will cause more than one element to be read out of bounds and potentially written out of bounds as well during sorting, is that correct? 

If this can be used to trample out-of-bounds memory during sorting process, or read a large amount of data out of bounds, I'd rather have a regular advisory.

niklasf:
Some reads use `v.get_unchecked(...)`, but writing consistently uses `&mut v[...]` or `v.swap(...)`, so all writes have bounds checking.

Which offsets can be read is not completely arbitary, because a bounds-checked read at `(index - 2) / 2` is attempted first. But basically for large vectors, an equally large region of memory behind the vector becomes accessible.

Depending on the read contents, the algorithm will attempt a panicking write, or continue. So I would conservatively assume that observing this behavior allows inferring the contents. In practise, this is probably very tricky or impossible, especially if the panic ends the program.

Shnatsel:
Thank you for the explanation! Yes, in that case I agree that this is not a sufficiently big deal to warrant a regular advisory.

Could you also add a bit of info about this to the advisory text, for anyone else wondering about the potential impact?

niklasf:
Sure. Added more details.

Shnatsel:
Thanks a lot! Merging.

PR #1618, opened 2023-02-25
Add ascii advisory

KisaragiEffective:
related: https://github.com/tomprogrammer/rust-ascii/issues/64


Shnatsel:
I fixed up some fields, CI should pass now. Looks good to go. Thank you!

Shnatsel:
Actually I see that 0.9.2 is yanked, I assume it's still problematic, so I am going to change the version specification to `>= 0.9.3`

PR #1623, opened 2023-03-01
Add unmaintained advisory for ncollide

damienrg:
There is no activity since March 2022 and the maintainer has advised that this crate is passively-maintained since January 2021.

### DIFF
--- a/crates/ncollide2d/RUSTSEC-0000-0000.md
+++ b/crates/ncollide2d/RUSTSEC-0000-0000.md
@@ -1,0 +1,15 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "ncollide2d"
+date = "2021-01-29"
+url = "https://github.com/dimforge/ncollide"
+informational = "unmaintained"
+
+[versions]
+patched = []
+```
+# ncollide2d is unmaintained
+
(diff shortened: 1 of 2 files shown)


pinkforest:
Thanks for the contribution -

We reserve unmaintained PR's only for completely unreachable maintainers or where there is prospect that security fixes would not get merged.

Passive maintenance alone as such does not directly imply that security fixes don't get merged.

We did previously have sodiumoxide crate on which was in passive maintenance mode but since it was a crypto crate we treated it differently - but only after lengthy observation period and other issues it was facing.

Therefore I recommend for now closing this since the maintainer has for now promised to passively maintain this crate which does not involve any security critical components e.g. cryptography etc. and where the maintainer has not been unreachable and where the maintainer has not explicitly signaled that the crate is completely deprecated.

If the maintainer wishes to signal that we can announce this via warning via RustSec then we can also do this but we need the maintainer consent in this case or period of observation (90 days) where the maintainer does not confirm whether security issues via maintenance are not handled - e.g. it would be a good idea to post explicit question and @ping the maintainer in the repository.

sebcrozet:
Hey! Maintainer of  `ncollide` here. I dont consider `ncollide` as being maintained now that my focus is on its replacement `parry`. I’ll update the crates status as unmaintained soon.

Issue #1209, opened 2022-03-08
`maligned::align_first` is unsound


kyrias:
The [`align_first`](https://docs.rs/maligned/latest/maligned/fn.align_first.html) function of the `maligned` crate manually allocates with an alignment larger than `T`, and then uses `Vec::from_raw_parts` on that allocation to get a `Vec<T>`.

As per the [stdlib documentation](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts) this violates the `GlobalAlloc::dealloc` safety contract and causes undefined behaviour, since the allocation has to have the exact same allocation as `T`.

pinkforest:
_This crate has 15,247 downloads all time, ~25 average daily downloads supposedly affected versions._

Seems there are couple of open issues too and the repo hasn't moved for a while

https://github.com/tylerhawkes/maligned/issues/5 - Unsoundness in heap allocation - 1 Jan 2022
https://github.com/tylerhawkes/maligned/issues/6 - Padding bytes UB - 1 Mar 2021

Would you mind writing an issue about this into:
https://github.com/tylerhawkes/maligned

And if you could ping the maintainers to get these checked out and respond so we have some response potentially.

Guessing based on who has merged into main branch the maintainers may be:
@tylerhawkes and @thomstewart 

I see that nobody has raised the maintenance status with the maintainer(s) before either

It would be nice to try to nudge the maintainers a bit to see if they can fix these issues so we have a fix to action.

tylerhawkes:
I'll take a look at it this week. This is the first time I've heard about
this, so I'm sorry about missing notifications.

pinkforest:
@kyrias would you like to create a PR with `informational = unsound` around this ?

@tylerhawkes wonder if you have you been able to look into a fix ? Would love to point to a fixed version.

jayvdb:
https://github.com/tylerhawkes/maligned/pull/7 not actioned for 2.5 years

jayvdb:
There is already https://github.com/rustsec/advisory-db/blob/main/crates/maligned/RUSTSEC-2023-0017.md about https://github.com/tylerhawkes/maligned/issues/5 which is the `aligned_first` problem.

We need another `unsound` advisory about https://github.com/tylerhawkes/maligned/issues/6

https://github.com/sarah-quinones/aligned-vec looks like a crate that could be recommended as an alternative.
ping @sarah-quinones , could you confirm that?

sarah-quinones:
[...]

PR #1625, opened 2023-03-04
`maligned::align_first` causes incorrect deallocation

Nugine:
+ #1209


Shnatsel:
Thank you for writing it up!

The original author seems to be unresponsive, so we'll have to publish an advisory without a fix.

PR #1627, opened 2023-03-04
`out_reference::Out::from_raw` should be `unsafe`

### LINKED ISSUE #613: out_reference 0.1.x is unsound, needs advisory
https://github.com/RustyYato/out-ref/issues/1

There was no change for the past two years however I'm not eager to declare a small crate with a single specific job unmaintained if it works. So I'd suggest waiting a bit if the issue gets reply before claiming so.

Nugine:
Resolves #613

### DIFF
--- a/crates/out-reference/RUSTSEC-0000-0000.md
+++ b/crates/out-reference/RUSTSEC-0000-0000.md
@@ -2,7 +2,7 @@
 [advisory]
 id = "RUSTSEC-0000-0000"
 package = "out-reference"
-date = "2020-03-04"
+date = "2021-01-20"
 url = "https://github.com/RustyYato/out-ref/issues/1"
 informational = "unsound"
 categories = ["memory-corruption"]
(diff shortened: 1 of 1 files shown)


crates/out-reference/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "out-reference"
+date = "2020-03-04"
```
pinkforest, line 5:
```suggestion
date = "2021-01-20"
```
We use date when the issue was originally reported which aligns with other security databases.

pinkforest (commented):
Just one field needs adjusting - thanks a lot :partying_face:

pinkforest (approved):

crates/out-reference/RUSTSEC-0000-0000.md
Nugine: I was writing "2023-03-04" :laughing:

PR #1628, opened 2023-03-04
remove_dir_all have TOCTOU race condition

alexanderkjall:
reported in GHSA-mc8h-8q98-g5hr ( https://github.com/advisories/GHSA-mc8h-8q98-g5hr )


Shnatsel:
Thank you! We cannot use GHSA data directly due to license incompatibility (GHSA license is more restrictive than ours), so I've replaced the info with an excerpt from the upstream changelog entry.

@XAMPPRocky are you OK with releasing the 0.8.0 changelog entry for `remove_dir_all` so we could reuse the text in the security advisory?

XAMPPRocky:
Hello, I did not write it, so it would be up to @rbtcollins to release the text into public domain.

rbtcollins:
I am happy to place that changelog into public domain

Shnatsel:
Thanks! Merged.

PR #1643, opened 2023-03-13
Add unmaintained advisory for diesel-adapter

rockstar:
`diesel-adapter` has some dependencies with CVEs open it, and an attempt to update things from a crate consumer shows that (a) there really isn't anyone assigned to maintenance of the crate, e.g. issue fixes are requested from the community rather than having someone able to response, and (b) the source has a broken CI that can't realistically be addressed from a few drive-by patches.

### DIFF
--- a/crates/diesel-adapter/RUSTSEC-0000-0000.md
+++ b/crates/diesel-adapter/RUSTSEC-0000-0000.md
@@ -1,0 +1,22 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "diesel-adapter"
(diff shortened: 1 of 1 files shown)


pinkforest:
The maintenance status issue got closed and made a statement to effect that it would get maintained for security issues and no deprecation is to be done.

Closing - re-open if the maintainer wishes to deprecate. Thanks for the contribution in any case figuring this out.

pinkforest:
Also the dependencies when these have advisories these will get flagged downstream via those crates down the dependency tree so there is no need to flag additional advisories regarding those.

PR #1644, opened 2023-03-14
Add "unmaintained" advisories for all old Gitoxide crates.

khuey:
Gitoxide mass renamed its crates from git-<crate> to gix-<crate>, and the old crate names are no longer receiving updates. Create advisories for all of them with messages pointing to the new crates.

### DIFF
--- a/crates/git-actor/RUSTSEC-0000-0000.md
+++ b/crates/git-actor/RUSTSEC-0000-0000.md
@@ -1,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "git-actor"
+date = "2023-03-14"
+url = "https://github.com/Byron/gitoxide/pull/741"
+informational = "unmaintained"
+
+[versions]
+patched = []
+```
+
+# Gitoxide has renamed its crates.
+
+All crates in the gitoxide project have been renamed from git-<crate> to
+gix-<crate>. The git- prefixed crates are no longer being updated. Switch
+to using gix-actor to continue receiving updates.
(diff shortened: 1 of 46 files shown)


pinkforest:
Is there a single crate that is used by all of those ? As we usually flag one crate that lights up the others.

Also maintainer would need to explicitly signal that the old crates are not going to receive security fixes -

Cc/ @Byron - Would you be okay for us to flag a nudge for anyone usign the old crates to migrate ?

If any potential security issues in the future would pop up that affect the old crates -

Would the old crates be given consideration / re-published on ?

Byron:
> Would you be okay for us to flag a nudge for anyone usign the old crates to migrate ?

Yes, that would be of great help. `git-features` is used by most of the crates, but not all of them. Further, some of these are placeholders (in the README.md, search for `placeholder`). `git-url` and `git-path` come to mind as leaf crates who don't depend on `git-features`, and with that the majority of these would be covered.

> Would the old crates be given consideration / re-published on ?

No, there is no intention to maintain the old names. 

----

I also see now that I should have used the [badges](https://doc.rust-lang.org/cargo/reference/manifest.html#the-badges-section) section to indicate these crates are unmaintained, instead with the last publish under the old name I only changed the crate description. Interestingly, I always discarded the 'badges' section by its name alone as I wasn't interested in badges, even though I would be interested in the maintenance status of a crate and would be happy to maintain such a field. From the docs, it seems to be visual in nature only (at least by intention), maybe I am missing something that would help understand how to use these fields correctly.

pinkforest:
Yeah it seems `git-features|path|hash` should be sufficient - I've reduced to those three.

PR #1645, opened 2023-03-15
Update RUSTSEC-2022-0076.md with v1 patch

MOZGIII:
The details page at https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-44mr-8vmm-wjhg says the v1 also has a patched release.

PR #1646, opened 2023-03-15
A follow-up after #1645

MOZGIII:
This is technically required for correctness, as v1 doesn't have this fn at all.


MOZGIII:
@pinkforest mind merging this one too? I didn't manage to push in time :D

pinkforest:
What is the accessor in v1 to update the configuration ?

pinkforest:
This I guess https://docs.rs/wasmtime/1.0.1/wasmtime/struct.Config.html#method.allocation_strategy

MOZGIII:
> This I guess [docs.rs/wasmtime/1.0.1/wasmtime/struct.Config.html#method.allocation_strategy](https://docs.rs/wasmtime/1.0.1/wasmtime/struct.Config.html#method.allocation_strategy)

Wow, good job finding it, I couldn't locate it quickly, but it seems like just what we're looking for!

PR #1647, opened 2023-03-19
NULL pointer dereference in stb_image

mbrubeck:
A bug in error handling in the `stb_image` C library could cause a NULL pointer dereference when attempting to load an invalid or unsupported image file.  This is fixed in version 0.2.5 and later of the `stb_image` Rust crate, by patching the C code to correctly handle NULL pointers.

See https://github.com/servo/rust-stb-image/pull/102 for more details.

### DIFF
--- a/crates/stb_image/RUSTSEC-0000-0000.md
+++ b/crates/stb_image/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "stb_image"
+date = "2023-03-19"
+url = "https://github.com/servo/rust-stb-image/pull/102"
+categories = ["memory-corruption"]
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you! Merging.

8573:
This fix seems liable to get lost the next time the vendored C file is updated to a new upstream version, as it appears to have been [the last time it was changed](https://github.com/servo/rust-stb-image/commit/f3bb83b783b53c1ecdbf4bbe0aed3cb65f0ce3c8).  Would it be reasonable/cost-effective for Servo to take steps to mitigate this risk, such as patching the C file at build-time instead (like how Linux distros patch upstream source code) or at least adding a warning in the comment at the top of the file?

8573:
I guess opening [a ticket in the Servo repository](https://github.com/servo/rust-stb-image/issues/103) is more appropriate.

PR #1652, opened 2023-03-23
Add ntru unsoundness advisory

jcaesar:
I suspect that this unsoundness can also result in a crypto-fail with just the right kind of wrong parameters, but that'd kind of be the users fault.

It also [seems like](https://github.com/FrinkGlobal/ntru-rs/issues/7) ntru is unmaintained. Should I submit a separate advisory / separate PR for that?

### DIFF
--- a/crates/ntru/RUSTSEC-0000-0000.md
+++ b/crates/ntru/RUSTSEC-0000-0000.md
@@ -12,8 +12,8 @@ informational = "unsound"
 patched = []
 
 [affected.functions]
-"ntru::types::PrivateKey::export" = [">= 0.4.3, < 0.5.6"]
-"ntru::types::PublicKey::export" = [">= 0.4.3, < 0.5.6"]
+"ntru::types::PrivateKey::export" = [">= 0.4.3"]
+"ntru::types::PublicKey::export" = [">= 0.4.3"]
 ```
 
 # Unsound FFI: Wrong API usage causes write past allocated area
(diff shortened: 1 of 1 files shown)


pinkforest:
I've pinged the maintainer to see - last release was six years let'see whats the plan if any.

tarcieri:
I'd call this `unsound` unless there's an actual PoC for a `crypto-failure`.

Separately it'd be good to file an unmaintained crate advisory.

crates/ntru/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
...
+keywords = ["ffi", "buffer overflow"]
+informational = "unsound"
+
+[versions]
+patched = []
+unaffected = []
+
+[affected.functions]
+"ntru::types::PrivateKey::export" = [">= 0.4.3, < 0.5.6"]
+"ntru::types::PublicKey::export" = [">= 0.4.3, < 0.5.6"]
```
pinkforest, line 17:
```suggestion
"ntru::types::PrivateKey::export" = [">= 0.4.3"]
"ntru::types::PublicKey::export" = [">= 0.4.3"]
```

Since there is not going to be a release most I would guess - these need to be existing version/s

crates/ntru/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
...
+package = "ntru"
+date = "2023-03-22"
+url = "https://github.com/FrinkGlobal/ntru-rs/issues/8"
+categories = ["memory-corruption"]
+keywords = ["ffi", "buffer overflow"]
+informational = "unsound"
+
+[versions]
+patched = []
+unaffected = []
```
pinkforest, line 13:
```suggestion
```

PR #1661, opened 2023-03-24
async-nats MitM vulnerability

paolobarbolini:
The following vulnerability has been reported by me to https://advisories.nats.io/ and fixed in `async-nats` 0.29. Once this is merged I'll copy-paste the vulnerability and also apply it to the `nats` crate, which presents the exact same issue.

~~I'm opening this as draft because there are some things I'd like to cleanup in the report which I don't have the time to do right now. For now I simply copy-pasted part of the report I also sent to Synadia.~~

~~I wasn't sure **which category to put this in**, as none seemed to apply.~~
All versions of `async-nats` are affected, while very early versions of the `nats` crate aren't because they don't support TLS.
If a CVE is requested by the Synadia Security Team I'll update the advisory with the CVE ID.

### DIFF
--- a/crates/async-nats/RUSTSEC-0000-0000.md
+++ b/crates/async-nats/RUSTSEC-0000-0000.md
@@ -15,11 +15,11 @@ patched = [">= 0.29.0"]
 
 The NATS official Rust clients are vulnerable to MitM when using TLS.
 
-The common name of the TLS certificate presented by the server is validated against the
-`host`name provided by the NATS server `INFO` message during the initial connection setup phase.
-Because the `INFO` parameters are sent as plaintext, a MitM proxy can replace the value of
-the `host` field to the common name of a valid certificate it controls, fooling the client into
-accepting it.
+The common name of the server's TLS certificate is validated against
+the `host`name provided by the server's plaintext `INFO` message
+during the initial connection setup phase. A MitM proxy can tamper with
+the `host` field's value by substituting it with the common name of a
+valid certificate it controls, fooling the client into accepting it.
 
 ## Reproduction steps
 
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thanks for the report!

The `crypto-failure` category looks applicable, but leaving it blank is valid too.

Other than that looks good to me, we can update this with the CVE ID once it's assigned. Let me know once you're comfortable with us merging this.

Shnatsel (approved):

crates/async-nats/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "async-nats"
+date = "2023-03-24"
+url = "https://github.com/nats-io/nats.rs/commit/817a7b942c462fa9d9938dcb62124173634132fb#diff-767d442397fcaaf2f83e8f924d4a70317a2ce4703a49964d6007707949cfa5f5L303-R304"
+categories = []
```
tarcieri, line 7: I'd suggest `crypto-failure` for MitM:

```suggestion
categories = ["crypto-failure"]
```

crates/async-nats/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
...
+keywords = ["tls", "mitm"]
+
+[versions]
+patched = [">= 0.29.0"]
+```
+
+# rustls integration vulnerable to MitM
```
alex, line 14: I think the title is a bit confusing, it makes it sound like this is a rustls vulnerability

paolobarbolini: Should be better now

crates/async-nats/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,37 @@
...
+3. The proxy makes a separate connection to the NATS server
+4. The NATS server replies with an `INFO` message
+5. The proxy reads the `INFO`, alters the `host` JSON field of it and passes
+   the altered INFO back to the client
+6. The proxy upgrades the client connection to TLS, presenting a certificate issued
+   by a certificate authority present in the client's keychain.
+   In the previous step the `host` was set to the common name of said certificate.
+7. `rustls` accepts the certificate, having verified that the common name matches the
+   attacker-controlled value given to it
+9. We fooled the client into accepting **our** certificate
```
pinkforest:
```suggestion
9. Client was fooled to accept attacker-controlled certificate.
```
nit - should indicate better who fooled or make it party-less
Since this is a copy of the original need to credit authoe

paolobarbolini: Thanks. I thought the same while you were posting it :smile:

Shnatsel:
Merging. Thanks everyone!

PR #1670, opened 2023-03-25
CI: Remove redundant linter noise

pinkforest:
It seems to take long time to print 500+ lines @ 1m 35s when we could just print the failures and total checked:
https://github.com/rustsec/advisory-db/actions/runs/4519170146/jobs/7959502197#step:5:1

Testing if this is the case and if this help to get to bottom of it.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -26,4 +26,4 @@ jobs:
         fi
 
     - name: Lint advisories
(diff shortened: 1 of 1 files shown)


pinkforest:
Hmm as I mildly suspected this doesn't seem to be the case - some runner performance is just not liking the linter :woman_shrugging: 
https://github.com/rustsec/advisory-db/actions/runs/4520169749/jobs/7961109342?pr=1670#step:5:5

PR #1671, opened 2023-03-31
Added unsound `spin`

zesterer:
This should only be merged when the fix in https://github.com/mvdnes/spin-rs/pull/149 has merged and a new version (`0.9.8`) is released.

### DIFF
--- a/crates/spin/RUSTSEC-0000-0000.md
+++ b/crates/spin/RUSTSEC-0000-0000.md
@@ -5,7 +5,6 @@ package = "spin"
 date = "2023-03-31"
 informational = "unsound"
 url = "https://github.com/mvdnes/spin-rs/issues/148"
-withdrawn = "2020-10-08"
 
 [versions]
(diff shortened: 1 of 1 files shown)


crates/spin/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "spin"
+date = "2023-03-31"
+informational = "unsound"
+url = "https://github.com/mvdnes/spin-rs/issues/148"
+withdrawn = "2020-10-08"
```
pinkforest, line 8:
```suggestion
withdrawn = "2020-10-08"
```

We would use this field if we withdraw advisories but I suspect this was just artifact from template gen

pinkforest:
Thanks for the contribution - we'll merge when there is a release

zesterer:
`0.9.8` has now been released, so this can be merged.

PR #1672, opened 2023-04-01
Add unmaintained advisory for syslog crate

bheylin:
The syslog crate is no longer maintained by it's current owner.

The PR referenced in the advisory has had no response for a year a three months.
I tried to contact the owner through Github and the email associated with their Github account.

### DIFF
--- a/crates/syslog/RUSTSEC-0000-0000.md
+++ b/crates/syslog/RUSTSEC-0000-0000.md
@@ -10,7 +10,7 @@ informational = "unmaintained"
 patched = []
 ```
 
-The syslog crate is no longer maintained by it's current owner.
-
-The PR referenced above has had no response for a year a three months.
-I tried to contact the owner through Github and the email associated with their Github account.
+# The syslog crate is no longer maintained by it's current owner.
+# 
+# The PR referenced above has had no response for a year a three months.
+# I tried to contact the owner through Github and the email associated with their Github account.
--- a/crates/syslog/RUSTSEC-0000-0000.md
+++ b/crates/syslog/RUSTSEC-0000-0000.md
@@ -1,0 +1,16 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "syslog"
+date = "2023-04-01"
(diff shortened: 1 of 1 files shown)


crates/syslog/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,16 @@
...
+informational = "unmaintained"
+
+[versions]
+patched = []
+```
+
+# The syslog crate is no longer maintained by it's current owner.
+# 
+# The PR referenced above has had no response for a year a three months.
+# I tried to contact the owner through Github and the email associated with their Github account.
```
pinkforest, line 16: I don't see why this advisory should get merged as of now.

I don't see maintainer being communicated explicitly from security PoV.

Our unmaintained advisories are reserved for completely unreachable maintainers and I don't see the necessary explicit communication that would support flagging an advisory.

Please can we fill the necessary detail accordingly per this example:
https://github.com/rustsec/advisory-db/blob/main/crates/buf_redux/RUSTSEC-2023-0028.md

The maintainer seems very active and I don't see that maintainer would not merge security fixes if any.

@Geal just wondering if it would be worthwhile to nudge people to use other implementations ? Personally I don't see big reason to flag warning but just in case you would feel a gentle note would be warranted from your maintainer PoV that this crate may be deprecated possibly and you would be happy for people to not use the crate ?

Re; alternatives there seems another syslog implementation by name of `syslog-rs` but I'm not going to paint any alternatives at this stage since the maintainer hasn't had the word out.

pinkforest (changes requested):
Please see comment - this needs more detail and more communication.

crates/syslog/RUSTSEC-0000-0000.md
Geal: well that was a rude way to get my attention :grin: 
That crate is not unmaintained, it's just that as a very thin wrapper over syslog, it does not warrant a lot of releases (it's mostly 1 per year by now), unless there are serious issues or vulnerabilities, and that PR was neither.
@bheylin I'll get to it, chill a bit.

Issue #1543, opened 2023-01-23
`daemonize` unmaintained


oherrala:
There has been issue asking for new release since 2021-09: https://github.com/knsd/daemonize/issues/46 No commits, no communication, etc. from the author since 2021-07.

There's also issue https://github.com/knsd/daemonize/issues/44#issuecomment-887287257 from 2021-07 where author promised new version in two weeks, however that never happened.

Also ping @knsd, is `daemonize` crate still maintained?

pinkforest:
Yes I think this is a clear cut case - thanks :+1: wanna sketch a quick PR per above ^ ?

sidunder:
I'll write up a PR.

breard-r:
A new version of `daemonize` has been released since this advisory, therefore I don't think this crate should still be considered unmaintained anymore. Moreover, the listed alternatives have not seen any updates for respectively 1 and 2 years, so this advisory could misleads people into replacing a maintained crate by an unmaintained one.

Is it possible to take down this advisory? I see procedures to add new ones but not remove them.

tarcieri:
Advisories can be removed by adding `withdrawn` with the date.

See example here: https://github.com/rustsec/advisory-db/blob/main/README.md#advisory-format

breard-r:
@tarcieri Thank you, I'll create a PR in order to add `withdrawn`!

PR #1676, opened 2023-04-08
Withdraw RUSTSEC-2021-0147

breard-r:
Withdraw the unmaintained advisory for `daemonize` requested in #1543.

On Feb 19 2023, the author started to maintain the `daemonize` crate again.
Version 0.5.0 has been released on Feb 25 2023.
Therefore, this crate in no longer unmaintained and advisory `RUSTSEC-2021-0147` should be withdrawn.


amousset:
Thanks!

PR #1680, opened 2023-04-11
Add unmaintained colored crate

sanpii:


amousset:
It seems maintained now, [latest release is 25 days old](https://crates.io/crates/colored/versions). Closing.

PR #1682, opened 2023-04-12
Add notice for borsh issue

maxammann:
This adds a notice for a long known unsound problem in borsh.

### DIFF
--- a/crates/borsh/RUSTSEC-0000-0000.md
+++ b/crates/borsh/RUSTSEC-0000-0000.md
@@ -17,4 +17,6 @@ patched = []
 
 Affected versions of borsh cause undefined behaviour when zero-sized-types (ZST) are parsed and the Copy/Clone traits are not implemented/derived.
 For instance if 1000 instances of a ZST are deserialized, and the ZST is not copy (this can be achieved through a a singleton), 
-then accessing/writing to deserialized data will cause a segmentation faul.
+then accessing/writing to deserialized data will cause a segmentation fault.
+
+There is currently no way for borsh to read data without also providing a Rust type. Therefore, it there are not ZST used for serialization, then you are not affected by this issue. 


crates/borsh/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,22 @@
+```toml
+[advisory]
+# Identifier for the advisory (mandatory). Will be assigned a "RUSTSEC-YYYY-NNNN"
+# identifier e.g. RUSTSEC-2018-0001. Please use "RUSTSEC-0000-0000" in PRs.
+id = "RUSTSEC-0000-0000"
+package = "borsh"
+date = "2023-04-12"
+url = "https://github.com/near/borsh-rs/issues/52"
```
Shnatsel, line 8:
```suggestion
url = "https://github.com/near/borsh-rs/issues/19"
```

I believe this is the correct issue to link? It's talking about the unsoundness more directly, and provides a reproducing example.

crates/borsh/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,22 @@
...
+references = ["https://github.com/near/borsh-rs/pull/136"]
+informational = "unsound"
+categories = ["denial-of-service"]
```
Shnatsel, line 11: Issue 19 (about unsoundness) seems to be causing an out-of-bounds access, so I think `memory-corruption` is more appropriate.

Shnatsel (commented):
I've left a few comments, but looks good to me otherwise. Thanks!

crates/borsh/RUSTSEC-0000-0000.md
maxammann: Fine for me :)

crates/borsh/RUSTSEC-0000-0000.md
maxammann: I'm confused here because denial-of-service and memory-corruption are different categories inthemselves?

Sometimes memory-corruption leads to denial-of-service.
Sometimes memory-corruption leads to RCE. 

So idk, maybe put both?

Shnatsel: We usually put the more severe one.

The (partial) mitigation is a denial-of-service vector itself, which is described in issue 52, but I'd consider that a separate issue.

crates/borsh/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,22 @@
...
+[affected]
+[versions]
+patched = []
+```
+
+# Parsing borsh messages with ZST which are not-copy/clone is unsound
+
+Affected versions of borsh cause undefined behaviour when zero-sized-types (ZST) are parsed and the Copy/Clone traits are not implemented/derived.
+For instance if 1000 instances of a ZST are deserialized, and the ZST is not copy (this can be achieved through a a singleton), 
+then accessing/writing to deserialized data will cause a segmentation faul.
```
Shnatsel, line 22: Does the application code need to opt-in to deserializing ZSTs, e.g. explicitly define a struct with a ZST in it, or can this affect any deserializer?

In other words, is deserialization only done into structs defined at compile time, or is there an interface to deserialize arbitrary data?

crates/borsh/RUSTSEC-0000-0000.md
maxammann: Great, I'll change!

crates/borsh/RUSTSEC-0000-0000.md
maxammann: Good question, I will look into it!

maxammann: > Since Borsh is not a self-descriptive format we have a way to describe types serialized with Borsh so that we can deserialize serialized blobs without having Rust types available. 

There is a BorshSchema, but even that requires a Rust type to be available: https://github.com/near/borsh-rs/blob/79097e3c71ae469a101b4828457792bcf8be7f5f/borsh/src/schema_helpers.rs#L11

crates/borsh/RUSTSEC-0000-0000.md
maxammann: Done

crates/borsh/RUSTSEC-0000-0000.md
maxammann: I added a note

Shnatsel: Thanks! Looks good, and the maintainer is also on board as per https://github.com/near/borsh-rs/pull/136

Merging.

PR #1684, opened 2023-04-13
Add advisory for h2: resource exhaustion vulnerability may lead to DoS

KisaragiEffective:
(should not be merged as is at writing this comment, see following comment)

### DIFF
--- a/crates/h2/RUSTSEC-0000-0000.md
+++ b/crates/h2/RUSTSEC-0000-0000.md
@@ -13,6 +13,8 @@ patched = [">= 0.3.17"]
 
 # h2 causes stream stacking, may cause Denial of Service (DoS)
 
-h2 causes stream stacking, may cause Denial of Service. Please see [GitHub Security Advisory entry](https://github.com/advisories/GHSA-f8vr-r385-rh5r) for more info.
+If an attacker is able to flood the network with pairs of `HEADERS`/`RST_STREAM` frames, such that the `h2` application is not able to accept them faster than the bytes are received, the pending accept queue can grow in memory usage. Being able to do this consistently can result in excessive memory use, and eventually trigger Out Of Memory.
+
+Please see the [GitHub Security Advisory entry](https://github.com/advisories/GHSA-f8vr-r385-rh5r) for more info.
 
 This flaw is corrected in [hyperium/h2#668](https://github.com/hyperium/h2/pull/668), which restricts remote reset stream count by default.
--- a/crates/h2/RUSTSEC-0000-0000.md
+++ b/crates/h2/RUSTSEC-0000-0000.md
@@ -13,6 +13,6 @@ patched = [">= 0.3.17"]
 
 # h2 causes stream stacking, may cause Denial of Service (DoS)
 
-h2 causes stream stacking, may cause Denial of Service.
+h2 causes stream stacking, may cause Denial of Service. Please see [GitHub Security Advisory entry](https://github.com/advisories/GHSA-f8vr-r385-rh5r) for more info.
 
 This flaw is corrected in [hyperium/h2#668](https://github.com/hyperium/h2/pull/668), which restricts remote reset stream count by default.
--- a/crates/h2/RUSTSEC-0000-0000.md
+++ b/crates/h2/RUSTSEC-0000-0000.md
@@ -1,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "h2"
+date = "2023-04-14"
+url = "https://github.com/hyperium/hyper/issues/2877"
+categories = ["denial-of-service"]
+keywords = ["http", "http2", "h2"]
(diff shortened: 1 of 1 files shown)


crates/h2/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,18 @@
...
+categories = ["denial-of-service"]
+keywords = ["http", "http2", "h2"]
+aliases = ["CVE-2023-26964", "GHSA-f8vr-r385-rh5r"]
+[versions]
+patched = [">= 0.3.17"]
+```
+
+# h2 causes stream stacking, may cause Denial of Service (DoS)
+
+h2 causes stream stacking, may cause Denial of Service.
```
tarcieri, line 16: The description in the [GHSA advisory was a bit more extensive](https://github.com/advisories/GHSA-f8vr-r385-rh5r).

tarcieri: I'm not saying to verbatim copy it, but it would be nice to have the same information

KisaragiEffective: I don't believe I can give appropriate summary for this entry.

tarcieri: Okay, well the added info is an improvement at least

Noah-Kennedy:
Let's get @seanmonstar's sign off on this to make sure the wording is correct regarding what the vuln is.

The original CVE was not very descriptive of the issue, and mentioned only an implementation detail which of itself isn't necessarily exploitable. It was only in conjunction with other things that there could be an issue, we found.

crates/h2/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,18 @@
...
+categories = ["denial-of-service"]
+keywords = ["http", "http2", "h2"]
+aliases = ["CVE-2023-26964", "GHSA-f8vr-r385-rh5r"]
+[versions]
+patched = [">= 0.3.17"]
+```
+
+# h2 causes stream stacking, may cause Denial of Service (DoS)
+
+h2 causes stream stacking, may cause Denial of Service. Please see [GitHub Security Advisory entry](https://github.com/advisories/GHSA-f8vr-r385-rh5r) for more info.
```
seanmonstar, line 16: I'd never heard of stream stacking, and a quick Google search only brings up a copy of the same CVE in all the registries (😅).

```suggestion
If an attacker is able to flood the network with pairs of `HEADERS`/`RST_STREAM` frames, such that the `h2` application is not able to accept them faster than the bytes are received, the pending accept queue can grow in memory usage. Being able to do this consistently can result in excessive memory use, and eventually trigger Out Of Memory.

Please see the [GitHub Security Advisory entry](https://github.com/advisories/GHSA-f8vr-r385-rh5r) for more info.
```

Fair?

Noah-Kennedy:
It might actually be better for him or myself to write this. This is a good place for us to pull the useful information out of our heads and put it to paper.

The original CVE wasn't helpful for maintainers or users, and I'd really like this one to be better.

KisaragiEffective:
I believe so, I don't understand whole point at currenet moment nor have read entire thread around it.

crates/h2/RUSTSEC-0000-0000.md
Noah-Kennedy: This is somewhat of an unusual case given the circumstances of that CVE.

seanmonstar: At the same time, we're guessing. The reporter hasn't confirmed that was the issue _they_ saw. So, I dunno 🤷

Noah-Kennedy: I believe the reporter of the issue was a different person than the reporter of the CVE. I think the CVE was just a copy-paste of the issue.

tarcieri:
@seanmonstar can you confirm the advisory looks good now?

seanmonstar:
LGTM! Thanks for all the work to keep this db running!

crates/h2/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,20 @@
...
+[versions]
+patched = [">= 0.3.17"]
+```
+
+# Resource exhaustion vulnerability in h2 may lead to Denial of Service (DoS)
+
+If an attacker is able to flood the network with pairs of `HEADERS`/`RST_STREAM` frames, such that the `h2` application is not able to accept them faster than the bytes are received, the pending accept queue can grow in memory usage. Being able to do this consistently can result in excessive memory use, and eventually trigger Out Of Memory.
+
+Please see the [GitHub Security Advisory entry](https://github.com/advisories/GHSA-f8vr-r385-rh5r) for more info.
+
```
tarcieri, line 19: This advisory now contains more detailed and correct information than the GHSA, so we probably shouldn't refer people to that.

```suggestion
Please see the [GitHub Security Advisory entry](https://github.com/advisories/GHSA-f8vr-r385-rh5r) for more info.

```

PR #1693, opened 2023-05-11
Update aliases from GHSA OSV export

amousset:


tarcieri:
@amousset if you're confident these look OK we can merge this (but remove draft first)

amousset:
This looks good now, ready for merge.

PR #1697, opened 2023-05-19
Add advisory for buffered-reader

nwalfield:
Attacker-controlled input can lead to an out-of-bounds index, which causes buffered-reader to panic.  This has been fixed in versions 1.2.0, 1.1.5, and 1.0.2 of buffered-reader.

### DIFF
--- a/crates/buffered-reader/RUSTSEC-0000-0000.md
+++ b/crates/buffered-reader/RUSTSEC-0000-0000.md
@@ -1,0 +1,24 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "buffered-reader"
+date = "2023-05-16"
+url = "https://lists.sequoia-pgp.org/hyperkitty/list/announce@lists.sequoia-pgp.org/thread/SN2E3QRT4DMQ5JNEK6VIN6DJ5SH766DI/"
+references = ["https://gitlab.com/sequoia-pgp/sequoia/-/tags/buffered-reader%2Fv1.2.0"]
+categories = ["denial-of-service"]
+# Attacker-controlled input can result in a panic due to an
+# out-of-bounds array index.
+keywords = ["panic"]
+
+[versions]
+patched = [">= 1.0.2, < 1.1.0", ">= 1.1.5, < 1.2.0", ">= 1.2.0"]
+```
+
+# Out-of-bounds array access leads to panic
+
+Affected versions of the crate have a bug where attacker-controlled
+input can result in the use of an out-of-bound array index.  Rust
+detects the use of the out-of-bound index and causes the application
+to panic.  An attacker may be able to use this to cause a
+denial-of-service.  However, it is not possible for an attacker to
+read from or write to the application's address space.


nwalfield:
CI is failing with:

```
12 | [versions]
   | ^^^^^^^^^^
bad parameter: Overlapping version ranges: [1.0.2, 2.0.0-0) and [1.1.5, 2.0.0-0)
```

I have:

```
[versions]
patched = [">= 1.2.0"]

# The fixes were backported to these versions (1.0.1 is in Debian
# Stable and 1.1.4 is in Debian Testing).
unaffected = ["1.0.2", "1.1.5"]
```

I guess that is wrong.  I'd appreciate help for articulating that all versions are affected except 1.2.0, 1.1.5 and 1.0.2, which I just released and include the required fixes.

crates/buffered-reader/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Out-of-bounds array access leads to panic
+
+Affected versions of the crate have a bug where
+attacker-controlled input can result in the use of an out-of-bound
+array index.  Rust detects the use of the out-of-bound index and
+causes the application to panic.  An attacker can use this to cause
+a denial-of-service.  However, it is not possible for an attacker
+to read from or write to the application's address space.
+```
```
amousset, line 29:
```suggestion
```

crates/buffered-reader/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,29 @@
...
+date = "2023-05-16"
+url = "https://lists.sequoia-pgp.org/hyperkitty/list/announce@lists.sequoia-pgp.org/thread/SN2E3QRT4DMQ5JNEK6VIN6DJ5SH766DI/"
+references = ["https://gitlab.com/sequoia-pgp/sequoia/-/tags/buffered-reader%2Fv1.2.0"]
+categories = ["denial-of-service"]
+# Attacker-controlled input can result in a panic due to an
+# out-of-bounds array index.
+keywords = ["panic"]
+
+[versions]
+patched = [">= 1.2.0"]
```
amousset, line 14:
```suggestion
patched = [">= 1.0.2, < 1.1.0", ">= 1.1.5, < 1.2.0", ">= 1.2.0"]
```

amousset (changes requested):
Thanks for reporting these! This should be the correct way to indicate these version ranges.

nwalfield:
Thanks for the help.  I indeed misunderstood the meaning of `unaffected`.  I think your suggestions are reasonable.  If no one else gets to it by Tuesday, then I'll do it then, I'll submit an updated patch.

crates/buffered-reader/RUSTSEC-0000-0000.md
nwalfield: Good catch, thanks.

crates/buffered-reader/RUSTSEC-0000-0000.md
nwalfield: Thanks for the suggestion.

PR #1698, opened 2023-05-23
Suggest kuchikiki as an alternative to kuchiki

rillian:
The `kuchiki` crate has been marked unmaintained. We're continuing to support a fork under then name `kuchikiki` which we intend to remain semver-compatible with our former upstream.

Suggest this as an alternative in RUSTSEC-2023-0019 since it is a direct replacement; the other alternatives involve significant porting effort.


tarcieri:
Thanks!

PR #1703, opened 2023-06-03
Vulnerability in trust-dns and trust-dns-server

jonasbb:
An attacker can form packet loops between vulnerable instances, leading to a denial-of-service for both network and CPU resources.

My supervisor (@crossow) and I found and reported the issue. @bluejekyll and @djc have acknowledged and resolved the issue.

Questions:

1. The same advisory should apply to both `trust-dns` and `trust-dns-server`. Is there a way to have a single advisory (i.e., single identifier) target multiple crates?
2. Affected is the type `trust_dns_server::server::ServerFuture`, not a specific public function. It seems that cannot be encoded via `[affected]`.

### DIFF
--- a/crates/trust-dns-server/RUSTSEC-0000-0000.md
+++ b/crates/trust-dns-server/RUSTSEC-0000-0000.md
@@ -1,0 +1,31 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "trust-dns-server"
+date = "2023-06-01"
+url = "https://github.com/bluejekyll/trust-dns/pull/1952"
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you!

On 1: There is currently no way to make a single advisory affect two different crates. If both crates contain vulnerable code, then there should be two advisories. It's fine if they contain the same text.

On 2: That is indeed correct, so `[affected]` is not applicable here.

jonasbb:
Ok, thanks. Then I will submit a second PR for trust-dns. Could you maybe check if and how `aliases` should be used here.

PR #1705, opened 2023-06-03
Vulnerability in trust-dns and trust-dns-server (Part 2)

jonasbb:
Same as #1703 but for `trust-dns`

### DIFF
--- a/crates/trust-dns/RUSTSEC-0000-0000.md
+++ b/crates/trust-dns/RUSTSEC-0000-0000.md
@@ -1,0 +1,31 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "trust-dns"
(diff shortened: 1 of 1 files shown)


Shnatsel:
I understand this is about the `trust-dns` binary, which unconditionally depends on the `trust-dns-server` crate. Is that correct? If that's the case, a separate advisory should not be needed - `cargo audit` will inspect the dependency tree and locate the vulnerable `trust-dns-server` crate.

If you have other use cases, please let me know, but as it stands I think an advisory about `trust-dns-server` should be sufficient.

jonasbb:
Yes it is about the binary, since it can be used "independed" of the `trust-dns-server`. But you are right, that there is a mandatory dependency between them. If you think the existing advisory is sufficient, then I close this one.

PR #1710, opened 2023-06-12
Unsound transmute in slint

shinmao:


crates/slint/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,18 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "slint"
```
ogoffart, line 4:
```suggestion
package = "slint-interpreter"
```

ogoffart (commented):
Slint dev here.

I don't believe this is worth a rustsec advisorty, since the given functions are in a private `ffi` module that can only be called from our C++ API. 
And even though our assertions were bad, this isn't a problem because the alignment and the size is actually matching. 
So this doesn't allow to invoke UB from rust using our public API.  
And there also should'nt be UB from C++ if the C++ API is used correctly.

Shnatsel:
@ogoffart thank you for the clarification! I'm going to go ahead and close this, then. 

I'm happy to reopen if there is evidence of this causing UB used from Rust.

PR #1713, opened 2023-06-14
serialport-rs memory leak advisory

huntc:
A memory leak had been found with serialport-rs, which has now been corrected.

### DIFF
--- a/crates/serialport/RUSTSEC-0000-0000.md
+++ b/crates/serialport/RUSTSEC-0000-0000.md
@@ -8,14 +8,12 @@ date = "2023-06-14"
 
 url = "https://github.com/serialport/serialport-rs/pull/98"
 
-categories = ["memory-corruption"]
+categories = ["denial-of-service"]
 
 keywords = ["memory", "leak"]
 
 [affected]
-arch = ["aarch64-apple-darwin", "i686-apple-darwin"]
-
-os = ["MacOS"]
+os = ["macos"]
 
 [versions]
 patched = [">= 4.2.1"]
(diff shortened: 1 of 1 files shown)


huntc:
I'm unsure why the target is failing. I checked the targets before. In essence, all Darwin targets present an issue. Can you please advise on what should be used for `x86_64-apple-darwin`?

huntc:
I’m unsuccessful in specifying the Darwin targets here. Thanks for any help.

Shnatsel:
Sure, I'll fix up the platform specification.

@eldruin could you confirm that the maintainers would like to treat this as a security issue? I understand this could eventually exhaust the memory available on the machine and trigger the OOM killer, but I see no obvious way for an attacker to trigger this.

sirhcel:
@huntc and I (as one of the maintainers) discussed this in length. I neither see exploiting this as a low hanging fruit and it might even be a corner case. But i see the benefit, that people using dependency scanners like `cargo audit` and Dependabot would be notified about it.

What are your points against making it a security issue?

huntc:
I was inclined to file this as a security issue given that “memory corruption” seemed to fit. However,perhaps making it informational/notice may also be sufficient? I agree that having people become informed of the problem is useful, even if there is no obvious attack vector.

Shnatsel:
"Memory corruption" refers to buffer overflows, use-after-free, and other bugs that lead to arbitrary code execution.

I agree that this is a potentially serious issue but not a security issue. I believe yanking the affected crate versions would be sufficient; `cargo audit` already warns about yanked crates even if there is no advisory filed (although this may change in the future).

huntc:
Shall we close this PR then, or shall I change it to an informational/notice category? Thanks.

Shnatsel:
Given that `informational = notice` is rather exotic and the databases we integrate with aren't set up to consume this, I suggest not publishing a RustSec advisory for this. So I'm going to go ahead and close this PR, but I'm happy to reopen if there is a plausible attack vector.

However, thank you for taking the issue seriously and taking the time to report and follow up on it!

PR #1726, opened 2023-07-05
Add shadowsocks misaligned read advisory

pinkisemils:
We found that on the latest stable compiler, the `shadowsocks` crate is reliably panicking under normal conditions due to a misaligned read, thus we've thought it'd be best to write up a potential advisory.

### DIFF
--- a/crates/shadowsocks/RUSTSEC-2023-0000.md
+++ b/crates/shadowsocks/RUSTSEC-2023-0000.md
@@ -1,0 +1,42 @@
+```toml
+[advisory]
+id = "RUSTSEC-2023-0000"
+package = "shadowsocks"
+date = "2023-07-05"
+url = "https://shadowsocks.org/"
+categories = ["denial-of-service"]
+keywords = []
+
+[versions]
+patched = [">= 1.15.4"]
+
+
+[affected]
+arch = ["x86_64", "aarch64"]
+os = ["linux", "macos"]
+```
+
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you for the report!

We would like to see confirmation from the Shadowsocks-rust maintainers that they'd like to treat this as a security issue. Additionally, it would be great to have a release fixing the issue - it seems easy enough to backport, so perhaps they could issue a patch release?

pinkisemils:
I've gone and made an issue on their [repository](https://github.com/shadowsocks/shadowsocks-rust/issues/1239). Let's see what they say :slightly_smiling_face:

pinkisemils:
Seems like the fix has been backported and [1.15.4](https://github.com/shadowsocks/shadowsocks-rust/releases/tag/v1.15.4) has been released with the fix.

pinkisemils:
Closing as this hasn't been acted upon.

Shnatsel:
Sorry for not acting on this, it seems this one fell through the cracks.

Since the observed behavior is an immediate crash on all common platforms, irrespective of the inputs, I'm not convinced that this is a security issue. The maintainers have not acknowledged it as a security issue either, so I think it is best if we do not merge this after all.

However, thank you for bringing this up with the author of the crate! It seems that your involvement did lead to the fix being backported to the current stable series of shadowsocks.

PR #1728, opened 2023-07-08
Bump rustsec-admin to 0.8.6

amousset:


Shnatsel:
Nice. Thanks for taking care of it so quickly!

PR #1735, opened 2023-07-30
Add unmaintained `dlopen_derive` advisory

smoelius:
`dlopen_derive` hasn't been updated since June 9, 2019.

`dlopen_derive` depends on [`quote = "0.6.12"`] and [`syn = "0.15.34"`]. Versions `1.0.0` of these dependencies were published on August 13, 2019. The `0.*` versions haven't received updates since.

## Recommended alternatives

- [`dlopen2_derive`]

[`dlopen2_derive`]: https://github.com/OpenByteDev/dlopen2
[`quote = "0.6.12"`]: https://github.com/dtolnay/quote/releases/tag/0.6.12
[`syn = "0.15.34"`]: https://github.com/dtolnay/syn/releases/tag/0.15.34

### DIFF
--- a/crates/dlopen_derive/RUSTSEC-0000-0000.md
+++ b/crates/dlopen_derive/RUSTSEC-0000-0000.md
@@ -1,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "dlopen_derive"
+date = "2023-07-30"
+url = "https://github.com/szymonwieloch/rust-dlopen/issues/47"
+informational = "unmaintained"
+
+[versions]
+patched = []
+```
+
+# `dlopen_derive` is unmaintained
+
+`dlopen_derive` hasn't been updated since June 9, 2019.
+
+`dlopen_derive` depends on [`quote = "0.6.12"`] and [`syn = "0.15.34"`]. Versions `1.0.0` of these dependencies were published on August 13, 2019. The `0.*` versions haven't received updates since.
+
+## Recommended alternatives
+
+- [`dlopen2_derive`]
+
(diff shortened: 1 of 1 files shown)


smoelius:
Is there anything else I need to do to justify this PR?

And, separately, I have the following question. [`dlopen`](https://crates.io/crates/dlopen), which [depends on `dlopen_derive`](https://github.com/szymonwieloch/rust-dlopen/blob/f9fed7168dfdcfd682a6c409595d4e2430b2dd2d/Cargo.toml#L17-L19),  is also unmaintained. I wrote this advisory thinking:
- If (hypothetically) a project depends directly on `dlopen_derive`, it will get a cargo-audit warning for `dlopen_derive`.
- If a project depends on `dlopen`, then it also (indirectly) depends on `dlopen_derive`, and so the prjoect will get a cargo-audit warning for `dlopen_derive`.

Is this a reasonable approach? Or, should this PR's description be adjusted to mention `dlopen`? Or, should there be a separate advisory for `dlopen` specifically?

amousset:
> Is there anything else I need to do to justify this PR?

No, according to [our policy](https://github.com/rustsec/advisory-db/blob/main/HOWTO_UNMAINTAINED.md) the unresponsiveness on the provided GitHub issue (for more than one year here) is enough.

> Is this a reasonable approach? Or, should this PR's description be adjusted to mention dlopen? Or, should there be a separate advisory for dlopen specifically?

I think your approach is reasonable. Maybe also mention the `dlopen` crate in the body of the advisory to make things clearer.

smoelius:
> I think your approach is reasonable. Maybe also mention the `dlopen` crate in the body of the advisory to make things clearer.

How is what I just pushed?

amousset:
Thanks!

PR #1736, opened 2023-08-02
fyrox: double free happens if panic occurs 

kuzeyardabulut:
Original issue report: https://github.com/FyroxEngine/Fyrox/pull/476

### DIFF
--- a/crates/fyrox/RUSTSEC-0000-0000.md
+++ b/crates/fyrox/RUSTSEC-0000-0000.md
@@ -1,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "fyrox"
+date = "2023-08-02"
+url = "https://github.com/FyroxEngine/Fyrox/pull/476"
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you for the report!

Do you know if a published version with the fix is forthcoming in the next few days? It would be better to file an actionable advisory with a fixed version already available.

Shnatsel:
Actually let me ask the maintainers on the original issue

adamreichold:
While using `ManuallyDrop` as done in the pull request is certainly an improvement, `Vec::from_raw_parts` does not panic (it does not even do arithmetic which could panic in debug builds) and hence I don't think there is an actual vulnerability here.

Nilstrieb:
I don't think this needs an advisory. The code was sound as-is, it was purely a refactoring that improved the code and made it nicer and more robust in the future.

Shnatsel:
Closing since this has been shown not to be a security issue in practice. Thank you @adamreichold and @Nilstrieb for the input!

PR #1742, opened 2023-08-08
Report Integer Overflow in imageproc 

kuzeyardabulut:
2 different bugs in original issue report: https://github.com/image-rs/imageproc/issues/537


Shnatsel:
Thank you for the report!

Unlike in C-like languages, in Rust integer overflow is not considered undefined behavior by the specification, so the compilers do not assume it will never happen. Integer overflow in Rust does cause a panic in debug mode or an incorrect result in release mode, but does not create opportunities for arbitrary code execution. This is a bug, but not a security issue, so I am going to go ahead and close this.

PR #1743, opened 2023-08-11
Add unmaintained ethabi advisory

maxammann:
See the readme of the project.

### DIFF
--- a/crates/ethabi/RUSTSEC-0000-0000.md
+++ b/crates/ethabi/RUSTSEC-0000-0000.md
@@ -1,0 +1,16 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "ethabi"
+date = "2023-08-11"
+informational = "unmaintained"
+url = "https://github.com/rust-ethereum/ethabi"
+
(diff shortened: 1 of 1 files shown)


amousset:
The README only states that the project is not _actively_ maintained, and the author is responsive. If still "passively maintained" (i.e. no active development but required security fixes would be made), then it does not match [our criteria](https://github.com/rustsec/advisory-db/blob/main/HOWTO_UNMAINTAINED.md) for the unmaintained advisories.

@vkgnosis Do you think an advisory makes sense or is the project still (passively) maintained?

maxammann:
I agree that we should ask again. This is the backstory on why I created this: https://github.com/rust-ethereum/ethabi/issues/301

Shnatsel:
The maintainer's stance doesn't seem to be substantially different from the typical "full disclosure" policy.

As Alexis has pointed out, this crate does not match [our criteria](https://github.com/rustsec/advisory-db/blob/main/HOWTO_UNMAINTAINED.md) for the unmaintained advisories. I'm going to go ahead and close this.

PR #1701, opened 2023-06-01
🦺 Advisory for unmaintained crate, `users`

zeenix:
The `users` crate hasn't seen any action since 2020-10-08. The developer seems [MIA] since.

I failed to find a direct replacement that's maintained but [`sysinfo`] crate provides the same
functionality since the last release.

[MIA]: https://github.com/ogham/rust-users/issues/54
[`sysinfo`]: https://crates.io/crates/sysinfo

PR #1751, opened 2023-08-21
Add uzers as alternative recommendation for users

gierens:
Following up on #1701

We use `users` in https://github.com/eza-community/eza so I decided to fork `users` to https://github.com/gierens/uzers-rs . I reviewed merged existing pull requests and plan to maintain it there. It's also on crates.io in version v0.11.1, see https://crates.io/crates/uzers 

This change adds `uzers` to the recommended alternatives for `users`.


Shnatsel:
Thanks!

PR #1752, opened 2023-08-22
CPU denial of service in rustls-webpki and webpki crates

ctz:


Shnatsel:
Thanks for the report!

It seems that `webpki` is unmaintained, with the maintainer MIA, so there will be no fix available. Do I understand correctly that switching to `rustls-webpki` is the only way to resolve the issue? If so, this should be incorporated into the advisory text for `webpki`.

ctz:
Yes I believe that is the current situation. The `rustls-webpki` 0.100.x release line is pretty close API-wise to the last `webpki` release. Will add a note.

Shnatsel:
I'm pretty sure remember it working fine in this case.

ctz:
> Will add a note.

This is done.

Shnatsel:
ID assignment worked fine: #1753

This advisory is now live. Thanks for reporting it!

briansmith:
> webpki is unmaintained, with the maintainer MIA, so there will be no fix available. Do I understand correctly

No, that's not correct.

Shnatsel:
That's great news! I see 0.22.1 with the fix is up on crates.io, I'll update the advisory accordingly.

Shnatsel:
Ah nevermind, I see it was already updated in #1762. Thanks!

PR #1770, opened 2023-09-07
Update reported issues in multiple packages

shinmao:


shinmao:
Sorry for sending PRs for several times. I finally find out how to check all the failed lints together.

crates/data-buffer/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,17 @@
...
+keywords = ["type-confusion"]
+
+[versions]
+patched = []
+```
+
+# Allows uninitialized memory exposure in safe function
+The safe function `push` allows users to cast arbitrary types as bytes. If user provides a `struct` type with padding bytes, it could violate the safety guarantee of `func` and expose the uninitialized memory.
+
+Note: The crate is not maintained anymore.
```
alex, line 17: If this is the case, a seperate unmaintained advisory would be more appropriate.

shinmao: Hi, I think my case is similar to this one: [RUSTSEC-2023-0055](https://rustsec.org/advisories/RUSTSEC-2023-0055.html), so I only mentioned it in advisories. Or I can update it again if you want?

shinmao:
Is there any update from the maintainers of the packages? 

Just want to keep track of the status following the timeline of two weeks, but no rush :)

I can also provide the information about which issues have been verified by the maintainer if it can help you 👍

shinmao:
hi @alex , I am wondering the progress of this PR so far. If the PR has some wrong format then it should be rejected so that I can correct the issues:)

kornelski:
It would be better to split this PR into one per crate, since in the current form nothing can be merged until all reports are verified and up-to-date.

For example xous has been patched, but the advisory doesn't contain the patched version.

shinmao:
Yes I could do it today. Thanks for reminding me.

PR #1773, opened 2023-09-10
Inventory 0.1 is unsound (allows std access before init of Rust runtime)

dtolnay:
I am reporting this against `inventory` even though the affected unsafe code lives in the `ctor` crate. I believe the majority of usages of `ctor` use it soundly. In fact inventory 0.2, which uses the same version of ctor as inventory 0.1, uses it in a manner that is not affected by this bug.

- Relevant standard library issue: https://github.com/rust-lang/rust/issues/110708
- Relevant inventory PR: https://github.com/dtolnay/inventory/pull/43


amousset:
Thanks!

PR #1782, opened 2023-09-13
CVE-2023-39914 in bcder.

partim:
The bcder crate has a number of issues resulting in panics when decoding invalid BER/CER/DER encoded data which may result in denial-of-service in applications using the crate.

### DIFF
--- a/crates/bcder/RUSTSEC-0000-0000.md
+++ b/crates/bcder/RUSTSEC-0000-0000.md
@@ -4,8 +4,6 @@ id = "RUSTSEC-0000-0000"
 package = "bcder"
 date = "2023-09-13"
 url = "https://nlnetlabs.nl/downloads/bcder/CVE-2023-39914.txt"
-# Valid categories: "code-execution", "crypto-failure", "denial-of-service", "file-disclosure"
-# "format-injection", "memory-corruption", "memory-exposure", "privilege-escalation"
 categories = ["denial-of-service"]
 keywords = ["example", "freeform", "keywords"]
 aliases = ["CVE-2023-39914"]
--- a/crates/bcder/RUSTSEC-0000-0000.md
+++ b/crates/bcder/RUSTSEC-0000-0000.md
@@ -3,19 +3,17 @@
 id = "RUSTSEC-0000-0000"
 package = "bcder"
 date = "2023-09-13"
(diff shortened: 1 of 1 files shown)


crates/bcder/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,30 @@
...
+# Valid categories: "code-execution", "crypto-failure", "denial-of-service", "file-disclosure"
+# "format-injection", "memory-corruption", "memory-exposure", "privilege-escalation"
+categories = ["denial-of-service"]
+keywords = ["example", "freeform", "keywords"]
+aliases = ["CVE-2023-39914"]
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+
+[versions]
+patched = ["<= 0.7.2"]
+unaffected = [">= 0.7.3"]
```
alex, line 16:
```suggestion
patched = [">= 0.7.3"]
```

The current way you have it doesn't make sense, it implies no versions are affected.

Shnatsel:
Thank you for the report!

Is there an issue report on the original repository? If so, it would be preferable to link to it in the `url` field. If not, let's link to the CVE. It would also be nice to put https://github.com/NLnetLabs/bcder/pull/74 into `references`, like this: `references = ["https://github.com/NLnetLabs/bcder/pull/74"]`

Also, we generally do not carry advisories for panics. But I see you are the top committer on the original repository. We are happy to accept this provided you feel that this is indeed a security issue for the typical uses of your crate.

tarcieri:
I think we make an exception on the DoS rule for security-critical format parsers that are routinely exposed to untrusted inputs

crates/bcder/RUSTSEC-0000-0000.md
partim: Apologies. Should read and not assume …

Shnatsel (approved):
Just one nit, looks good otherwise

partim:
I was on the fence whether this should be an advisory, but the case is explicitly mentioned as a criteria in https://rustsec.org/contributing.html. We created a separate CVE for it mostly because we wanted something to refer to in the related [CVE in Routinator](https://github.com/NLnetLabs/routinator/pull/891).

PR #1792, opened 2023-09-25
Add advisory for unsoundness in Hound < 3.5.1

ruuda:
See the diff for the description of the unsoundness.

### DIFF
--- a/crates/hound/RUSTSEC-0000-0000.md
+++ b/crates/hound/RUSTSEC-0000-0000.md
@@ -1,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "hound"
+date = "2023-09-25"
+url = "https://github.com/ruuda/hound/blob/9a962e1819102a33f0af2a2c60be6f0edcdfd7a8/changelog.md#351"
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you for backporting the fix and for the disclosure!

While this is technically undefined behavior under the current Rust specification, I understand there is no evidence that it is exploited by the compiler?

We generally use `unsound` marking for crates make it possible to misuse them and cause memory safety issues through the safe API. Since this is not the case here, and this does not cause any security issues in practice (yet?), I do not see any immediate security impact.

So I would advise against filing a security advisory in this case, but I'm happy to merge it if you are certain you want it.

ruuda:
> While this is technically undefined behavior under the current Rust specification, I understand there is no evidence that it is exploited by the compiler?

Yes, correct.

> We generally use unsound marking for crates make it possible to misuse them and cause memory safety issues through the safe API. Since this is not the case here, and this does not cause any security issues in practice (yet?), I do not see any immediate security impact.

I agree there is no security impact; I mainly wanted to preempt any confusion that might arise for people who see the word “unsound” in the changelog and jump to conclusions.

> So I would advise against filing a security advisory in this case

Makes sense, I’ll close it then. Thank you for the clarification!

PR #1793, opened 2023-09-26
Update info about CVE-2023-5129

kornelski:
There is finally a CVE specifically for libwebp that doesn't suggest it's a Chrome-only problem.


amousset (approved):

amousset:
Thanks!

PR #1795, opened 2023-09-29
Create advisory for tungstenite DoS

djc:
@daniel-abramov @agalakhov please have a look. I noticed this via GitHub's vulnerability reporting system, but would be nice to have this in the RustSec advisory database, too (because it triggers tools like cargo-audit and cargo-deny).


agalakhov:
Very good, thank you very much!

Shnatsel:
Looks good, merging. Thanks!

PR #1798, opened 2023-10-02
Document the privilege-escalation vulnerability in pleaser.

alexanderkjall:
Note that the reproducer doesn't work out of the box on a modern kernel, as the ioctl TIOCSTI is disabled by default nowadays.

### DIFF
--- a/crates/pleaser/RUSTSEC-0000-0000.md
+++ b/crates/pleaser/RUSTSEC-0000-0000.md
@@ -16,7 +16,7 @@ patched = []
 
 # Vulnerable to privilege escalation using ioctls TIOCSTI and TIOCLINUX
 
-please is vulnerabily to privilege escalation using ioctls TIOCSTI
+please is vulnerable to privilege escalation using ioctls TIOCSTI
 and TIOCLINUX on systems where they are not disabled.
 
 Here is how to see it in action:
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you for the report!

This is clearly a security issue with a published PoC that has not been addressed for 5 months, so I am going to merge this directly and notify the upstream issue after the fact.

Could you also add a note that many modern Linux distributions disable that IOCTL, and a way to test if your system is vulnerable? That would make the advisory more actionable. 

I believe simply running `sysctl dev.tty.legacy_tiocsti` should show it, and if set to `1` it would be exploitable. Could you confirm that this command outputs 0 on the systems that are not affected?

crates/pleaser/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,49 @@
...
+
+[versions]
+patched = []
+
+[affected]
+```
+
+# Vulnerable to privilege escalation using ioctls TIOCSTI and TIOCLINUX
+
+please is vulnerabily to privilege escalation using ioctls TIOCSTI
```
BlackHoleFox, line 19: Typo nit:
```suggestion
please is vulnerable to privilege escalation using ioctls TIOCSTI
```

alexanderkjall: Thanks, fixed :)

alexanderkjall:
I added a section on how to check if TIOCSTI is disabled, but I don't know if that also affects TIOCLINUX (and the poc doesn't use that, so hard to test without building a new poc)

Shnatsel:
Ah, good point about `TIOCLINUX`! 

If the steps only cover part of the attack surface, it is best to leave them out. No reason to give a false sense of security when `TIOCSTI` is disabled but `TIOCLINUX` is not.

PR #1800, opened 2023-10-12
fehler is unmaintained

SandaruKasa:


tarcieri:
Based on https://github.com/withoutboats/fehler/issues/66 I think this meets our "implicitly unmaintained" criteria

PR #1802, opened 2023-10-14
Add advisory for soundness bug in simd-json-derive < 0.12

Licenser:


Shnatsel:
Thank you for the report!

CI fails because there is no title, see the example advisory for reference: https://github.com/rustsec/advisory-db/blob/main/EXAMPLE_ADVISORY.md

It would be nice to clarify the exact impact of the issue. It looks like a use-after-free at a glance, in which case it might deserve a proper advisory, not just `informational = unsound` which is only surfaced as a warning. Let me take a closer look.

Licenser:
Heya,

thanks for the quick reply :). I fixed the header. On the direct impact I was unsure what to put.

My suspicion is that it's freeing memory that isn't allocated property, but I'm not sure how to validate that assumption, and could be totally wrong.

Shnatsel:
There are two ways to figure out what exactly is happening.

The first is [sanitizers](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html). Build the code with Address Sanitizer enabled, something like `RUSTFLAGS=-Zsanitizer=address cargo +nightly build -Zbuild-std --target x86_64-unknown-linux-gnu` and then run it. This will print any invalid memory accesses. There's also a separate Memory Sanitizer for catching reads from uninitialized memory. Great all around, but no Windows support.

The other is [miri](https://github.com/rust-lang/miri). This should go something like:
```
rustup default nightly
rustup component add miri
cargo miri test
```
It is an interpreter that can run on any OS. It does very in-depth checks for UB but it is very slow, only suitable for small snippets. Doesn't support I/O or most other calls to the operating system, only pure Rust code.

Licenser:
Thanks so much!

Miri failed for me since the underlying library uses SIMD instructions, and the sanitizer didn't flag any issues either the address one, the memory one, nor the leak one. 

I used this little test snippet (mostly from the original issue) as a target. 
```rust
use simd_json_derive::Deserialize;
fn main() {
    #[derive(Deserialize)]
    struct FooStr<'de> {
        foo: &'de str,
    }
    let mut json = br#"{"foo":"bar"}"#.to_vec();
    let res = FooStr::from_slice(&mut json);
    if let Ok(res) = res {
        assert_eq!(res.foo, "bar");
        println!("foo: {}", res.foo);
    } else {
        println!("this is bad, we hit the bug!");
    }
}
```


This is not to say that it's not an issue while testing, I got it to crash differently wich might give an indication:

```
❯ target/x86_64-unknown-linux-gnu/release/examples/foo
[examples/foo.rs:3] size = 1
[1]    142187 illegal hardware instruction  target/x86_64-unknown-linux-gnu/release/examples/foo
```

Shnatsel:
Well, apparently the compiler takes advantage of the UB and optimizes the code into some nonsense. Perhaps a debug build with sanitizers would reveal what exactly goes wrong?

If not, we can just say "has UB and optimizes into some nonsense" and leave it at that.

Licenser:
In debug builds it sadly doesn't trigger. I think we're stuck with `has UB and optimizes into some nonsense`. I updated the readme section to include that wonderful phrase ;)

Thanks for all the help @Shnatsel , I learned a few things and appreciate your time :)

PR #1803, opened 2023-10-14
Updating information about replacements for owning_ref

noamtashma:
I've just taken my original fix and pull request for `owning_Ref` and made it into a crate,
So now people using `owning_ref` can be directed to that as a possible replacement.


Shnatsel:
Thanks!

PR #1814, opened 2023-11-01
RUSTSEC advisory for sudo-rs: Path Traversal vulnerability

J3rry-1729:


Shnatsel:
Thanks!

Since this uses information available only in GHSA, we need confirmation from the submitter to include this.

@rnijveld are you OK with releasing this information into the public domain, so it could be included in the RustSec advisory database?

rnijveld:
@Shnatsel Sure! Go ahead!

PR #542, opened 2020-12-31
Update CVE numbers

Qwaz:
A lot of CVE advisories have been filed for RustSec bugs last night.
I tried my best not to make a mistake when updating the numbers, but please double check!

Thanks!

PR #1815, opened 2023-11-02
Remove CVE-2020-25575 from RUSTSEC-2019-0030 aliases

michaelkedar:
RUSTSEC-2019-0030 lists CVE-2020-25575 & GHSA-jq66-xh47-j9f3 as aliases, but they don't seem to be related.

CC: https://github.com/github/advisory-database/pull/2907


Shnatsel:
Thanks for the correction!

I was scared our automation isn't working properly, but this seems to have been introduced due to human error: https://github.com/rustsec/advisory-db/pull/542

PR #1818, opened 2023-11-10
Add Insufficient covariance check makes self_cell unsound

Voultapher:


Voultapher:
@steffahn thoughts?

Shnatsel (approved):
Looks good, I'm happy to merge as soon as @steffahn approves. Thanks!

steffahn:
There's a [discussion](https://github.com/Voultapher/self_cell/issues/49) to be had, whether a fixed `0.10.3` version (by means of re-exporting version `1.*` will also be published, in which case the `patched = [">= 1.0.2"]` would no longer be entirely accurate. Does RustSec support something like “everything `<= 0.10.2`, as well as `1.0.0 <= x <= 1.0.1` is affected, but `0.10.3`, and `>= 1.0.2` are patched” in the first place?

tarcieri:
Yes, that would look like:

```toml
affected = ["<= 0.10.2", ">1.0.0, <= 1.0.1"]
patched = ["0.10.3", ">= 1.0.2"]
```

steffahn:
The description is okay, though it reads a bit as if the covariance check used to not work at all. Maybe the sentence

> This allowed users to mark a dependent as `covariant` even though it allows interior-mutability.

could become

> This allowed users to mark a dependent as `covariant` even though its type was not covariant but invariant, for certain invariant types involving trait object lifetimes.

also removing the mention of “interior-mutability” which is hard to understand in this context, anyways.

steffahn:
> Yes, that would look like:
> 
> ```toml
> affected = ["<= 0.10.2", ">1.0.0, <= 1.0.1"]
> patched = ["0.10.3", ">= 1.0.2"]
> ```

Does `">1.0.0, <= 1.0.1"` exactly correspond to “`1.0.0 <= x <= 1.0.1`”, i.e including version `1.0.0` still as “affected”?

tarcieri:
The latter looks good, yeah

steffahn:
Looks like there was a confusion in the advice here. An `affected` key does not exists, and would be redundant anyways. An `unaffected` key would exist for earlier versions that are not affected *yet*. I don’t believe any earlier version didn’t have the issue. @Voultapher you just need to remove the `affected` line, it seems ^^

steffahn (approved):

steffahn:
@Shnatsel, this should be good to merge now. The `0.10.3` version question was resolved (it’s already published), and this advisory’s information updated accordingly.

PR #1824, opened 2023-11-23
`openssl` `X509StoreRef::objects` is unsound

alex:


alex:
One note: I marked this issue as resolved with the latest release. But really the fix is: Don't use this function. This function is still unsound in the latest release, it's just deprecated.

Is there a better way to indicate this?

PR #1836, opened 2023-12-11
Sync advisories ids from GitHub

amousset:
Required a [few changes](https://github.com/rustsec/rustsec/pull/656/commits/f56fb9b2a3a8272100ca8f31bb434f8e5b2be858) in the sync command.

Not sure what to do with the PYSEC ids. The occurrences in this pull request look like actual aliases (i.e. they refer to the same vulnerability in upstream code), but it may not always be the case.

### DIFF
--- a/crates/rsa/RUSTSEC-2023-0071.md
+++ b/crates/rsa/RUSTSEC-2023-0071.md
@@ -8,7 +8,7 @@ categories = ["crypto-failure"]
 url = "https://github.com/RustCrypto/RSA/issues/19#issuecomment-1822995643"
 references = ["https://people.redhat.com/~hkario/marvin/"]
 cvss = "CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:N"
-aliases = ["CVE-2023-49092", "GHSA-c38w-74pg-36hr"]
+aliases = ["CVE-2023-49092", "GHSA-4grx-2x9w-596c", "GHSA-c38w-74pg-36hr"]
 
 [versions]
 patched = []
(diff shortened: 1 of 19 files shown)


alex:
Looks like there's a conflict here.

crates/rsa/RUSTSEC-2023-0071.md
```diff
@@ -8,7 +8,7 @@ categories = ["crypto-failure"]
 url = "https://github.com/RustCrypto/RSA/issues/19#issuecomment-1822995643"
 references = ["https://people.redhat.com/~hkario/marvin/"]
 cvss = "CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:N"
-aliases = ["CVE-2023-49092", "GHSA-c38w-74pg-36hr"]
+aliases = ["CVE-2023-49092", "GHSA-4grx-2x9w-596c", "GHSA-c38w-74pg-36hr"]
```
tarcieri, line 11: #1838 put these in the opposite order:

```suggestion
aliases = ["CVE-2023-49092", "GHSA-c38w-74pg-36hr", "GHSA-4grx-2x9w-596c"]
```

Shnatsel:
I'd like to get this merged, it has stalled for too long.

I understand the blocker is the uncertainty around PYSEC IDs? Is there a good reason to expect they are different?

I would rather put them in `aliases` along with all the other IDs, otherwise that data will not make it into the OSV export where PYSEC advisories are also available. And if we are including the identifier, it would be a good idea to link it in a machine-readable way too.

The PYSEC advisories from this PR have GHSA and CVE IDs listed as OSV aliases, e.g. https://osv.dev/vulnerability/PYSEC-2021-321
Would PYSEC advisories without such a link be picked up by the import script?

Another option is to not carry PYSEC IDs at all, and just expect advisories to be the same thing if they have one ID in common, e.g. CVE or GHSA. This will likely be necessary to do regardless.

amousset:
I'll just remove the PYSEC IDs for now, relying on GHSA/CVE should be enough to avoid duplicates.

amousset:
Replaced by #1881.

PR #1846, opened 2024-01-07
Unsound sending of non-Send types in `threadalone`

dtolnay:
The repo has issues disabled, so I have not been able to report this bug upstream. https://github.com/cr0sh/threadalone

### DIFF
--- a/crates/threadalone/RUSTSEC-0000-0000.md
+++ b/crates/threadalone/RUSTSEC-0000-0000.md
@@ -1,0 +1,27 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "threadalone"
+date = "2024-01-07"
+url = "https://github.com/cr0sh/threadalone/issues/1"
+informational = "unsound"
+
+[versions]
+patched = []
+```
+
+# Unsound sending of non-Send types across threads
+
+Affected versions can run the `Drop` impl of a non-Send type on a different
+thread than it was created on.
+
+The flaw occurs when a stderr write performed by the `threadalone` crate fails,
+for example because stderr is redirected to a location on a filesystem that is
+full, or because stderr is a pipe that has been closed by the reader.
+
+Dropping a non-Send type on the wrong thread is unsound. If used with a type
+such as a pthread-based `MutexGuard`, [the consequence is undefined
+behavior][mutexguard]. If used with `Rc`, there would be a data race on the
+reference count, which is likewise undefined behavior.
+
+[mutexguard]: https://github.com/rust-lang/rust/issues/23465#issuecomment-82730326


dtolnay:
Repro:

```rust
mod thread_unsafe {
    use std::marker::PhantomData;
    use std::rc::Rc;
    use std::thread;

    pub struct Thing {
        not_send: PhantomData<Rc<()>>,
    }

    impl Thing {
        pub fn new() -> Self {
            println!("created on thread {:?}", thread::current().id());
            Thing {
                not_send: PhantomData,
            }
        }
    }

    impl Drop for Thing {
        fn drop(&mut self) {
            println!("dropped on thread {:?}", thread::current().id());
        }
    }
}

use std::thread;
use threadalone::ThreadAlone;

fn main() {
    let thing = thread_unsafe::Thing::new();

    // Does not compile: Thing cannot be sent between threads safely
    #[cfg(any())]
    let _ = thread::spawn(|| drop(thing)).join();

    let ta = ThreadAlone::new(thing);
    let _ = thread::spawn(|| drop(ta)).join();
}
```

```console
$ cargo build && target/debug/repro >&2 2>/dev/full
created on thread ThreadId(1)
dropped on thread ThreadId(2)
```

dtolnay:
FYI @cr0sh

dtolnay:
Repro of data race causing segfault in safe code:

```rust
use std::cell::Cell;
use std::thread;
use threadalone::ThreadAlone;

struct Corrupt<'a>(&'a Cell<String>);

impl<'a> Drop for Corrupt<'a> {
    fn drop(&mut self) {
        self.0.set(".".repeat(40));
    }
}

fn main() {
    let cell = Cell::new(String::new());
    thread::scope(|s| loop {
        let bad = ThreadAlone::new(Corrupt(&cell));
        s.spawn(|| drop(bad));
        cell.set(String::new());
    });
}
```

```console
$ cargo build && target/debug/testing >&2 2>/dev/full
Segmentation fault (core dumped)
```

cr0sh:
Oops, this is definitely not expected. Drop should be abort the runtime. I yanked the latest(and the only active) version.

Do you have a suggestion about why [my Drop impl](https://github.com/cr0sh/threadalone/blob/457fd19bd9dd8fabf48ade9885503bf25e23d9b5/src/lib.rs#L162-L165) does not work?

Anyway, your last repro does not segfault on my MacOS machine, it just aborts successfully. Maybe the `thread_id` semantics are different on each platform..? I tested against the latest commit(0.2.0).

```
cargo build && target/debug/testing >&2 2>/dev/null
[1]    57052 abort      target/debug/threadalone-test >&2 2> /dev/null
```

dtolnay:
Can you enable Issues in https://github.com/cr0sh/threadalone/settings? I can provide guidance there. This PR is not the right place.

tarcieri:
@cr0sh would you like for us to wait for you to publish a fixed version, or can we merge this advisory as-is?

We can always update it once you do release a fixed version.

cr0sh:
@dtolnay done, let's move on
@tarcieri I'll publish the fix once I fully understand what's wrong.

cr0sh:
Update: I acknowledged the error and published v0.2.1 with a fix. You're free to merge.

crates/threadalone/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,27 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "threadalone"
+date = "2024-01-07"
+url = "https://github.com/cr0sh/threadalone/issues/1"
+informational = "unsound"
+
+[versions]
+patched = []
```
tarcieri, line 10:
```suggestion
patched = ["0.2.1"]
```

Shnatsel:
I made a PR with the patched version, since the "Allow edits from maintainers" box was not ticked: #1858

Shnatsel:
I went ahead and merged https://github.com/rustsec/advisory-db/pull/1858, so I'm going to close this.

Thanks to @dtolnay for reporting this and to @cr0sh for the fix!

PR #1852, opened 2024-01-17
h2: Reset Flood vulnerability may lead to resource exhaustion and DOS

Noah-Kennedy:
Add Reset Flood advisory for older versions of h2.

### DIFF
--- a/crates/h2/RUSTSEC-0000-0000.md
+++ b/crates/h2/RUSTSEC-0000-0000.md
@@ -11,7 +11,7 @@ aliases = ["CVE-2019-9514"]
 related = ["CVE-2019-9511", "CVE-2019-9512", "CVE-2019-9513", "CVE-2019-9515", "CVE-2019-9516", "CVE-2019-9517", "CVE-2019-9518", "CVE-2023-44487"]
 
 [versions]
-patched = ["0.3.24", "0.4.2"]
+patched = ["^0.3.24", ">= 0.4.2"]
 ```
 
 # Resource exhaustion vulnerability in h2 may lead to Denial of Service (DoS)
--- a/crates/h2/RUSTSEC-0000-0000.md
+++ b/crates/h2/RUSTSEC-0000-0000.md
@@ -6,7 +6,7 @@ date = "2024-01-17"
 references = ["https://github.com/Netflix/security-bulletins/blob/master/advisories/third-party/2019-002.md"]
 categories = ["denial-of-service"]
 keywords = ["http", "http2", "h2"]
-related = ["CVE-2019-9511", "CVE-2019-9512", "CVE-2019-9513", "CVE-2019-9514", "CVE-2019-9515", "CVE-2019-9516", "CVE-2019-9517", "CVE-2019-9518", "CVE-2023-44487", "RUSTSEC-2023-0034"]
+related = ["CVE-2019-9514"]
 
 [versions]
 patched = ["^0.3.24", ">= 0.4.2"]
--- a/crates/h2/RUSTSEC-0000-0000.md
+++ b/crates/h2/RUSTSEC-0000-0000.md
@@ -4,7 +4,6 @@ id = "RUSTSEC-0000-0000"
 package = "h2"
 date = "2024-01-17"
 references = ["https://github.com/Netflix/security-bulletins/blob/master/advisories/third-party/2019-002.md"]
-informational = "notice"
 categories = ["denial-of-service"]
 keywords = ["http", "http2", "h2"]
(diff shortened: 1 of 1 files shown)


Noah-Kennedy:
Recently, I found that the h2 crate (and, by extension, Hyper) was vulnerable to Reset Flood attacks, and worked with @seanmonstar to patch the vulnerability.

It is patched in the newly released versions v0.3.24 and v0.4.2 of h2.

crates/h2/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
...
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "h2"
+date = "2024-01-17"
+references = ["https://github.com/Netflix/security-bulletins/blob/master/advisories/third-party/2019-002.md"]
+informational = "notice"
+categories = ["denial-of-service"]
+keywords = ["http", "http2", "h2"]
+aliases = ["CVE-2019-9514"]
+related = ["CVE-2019-9511", "CVE-2019-9512", "CVE-2019-9513", "CVE-2019-9515", "CVE-2019-9516", "CVE-2019-9517", "CVE-2019-9518", "CVE-2023-44487"]
```
tarcieri, line 11:
```suggestion
related = ["CVE-2019-9511", "CVE-2019-9512", "CVE-2019-9513", "CVE-2019-9515", "CVE-2019-9516", "CVE-2019-9517", "CVE-2019-9518", "CVE-2023-44487", "RUSTSEC-2023-0034"]
```

crates/h2/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "h2"
+date = "2024-01-17"
+references = ["https://github.com/Netflix/security-bulletins/blob/master/advisories/third-party/2019-002.md"]
+informational = "notice"
```
Shnatsel, line 7: Setting the `informational` field will make this advisory show up only as a warning. I think it is a legitimate vulnerability, so the field should be removed.
```suggestion
```

I've also gone ahead and fixed the version specification. Looks good to me otherwise!

seanmonstar (approved):
LGTM. Thanks Noah!

crates/h2/RUSTSEC-0000-0000.md
Noah-Kennedy: This suggestion is now out of date, but it makes sense so I'll commit it manually.

Shnatsel:
Do you think https://rustsec.org/advisories/RUSTSEC-2023-0034.html should be added to the `related` field? Once that's decided I'm happy to merge and have this go live.

crates/h2/RUSTSEC-0000-0000.md
Noah-Kennedy: Done

Noah-Kennedy:
@Shnatsel I already added it after seeing the comment from @tarcieri.

Noah-Kennedy:
I don't see why not, unless @seanmonstar has any objections

seanmonstar:
2023-0034 is a separate kind of attack. In that case, the attack is sending legit RST_STREAM frames, and the accept queue was growing.

In this case, the attack is to send junk frames that the server must then grow the send queue of outgoing reset frames.

Is that the point of related? I dunno. Just thought I'd clarify.

Noah-Kennedy:
I'm not really sure what the scope of that field should be

Noah-Kennedy:
I dropped in all of the 2019 netflix vulns as well as rapid reset

Noah-Kennedy:
and arguably rapid reset is also what RUSTSEC-2023-0034 is

Noah-Kennedy:
anyways, if we're using this for a loose collection of "http2 mechanisms are hard" vulns, then i think it's fine as is

otherwise, we should pull out everything other than the 2019 reset flood cve

Shnatsel:
> otherwise, we should pull out everything other than the 2019 reset flood cve

I think that is the best course of action. It's not immediately obvious for me, as a security responder, how any of these vulnerabilities are related if I'm given such a long list, and it would take a while to manually evaluate them.

Noah-Kennedy:
@Shnatsel let's go with that then, i'll pull the rest out

Noah-Kennedy:
@seanmonstar @Shnatsel done

Shnatsel:
Merging. Thanks!

Noah-Kennedy:
Thanks everyone!

PR #1855, opened 2024-01-19
rust-i18n-support: Use-after-free when setting the locale

Kijewski:
Version 3.0.0 introduced an `AtomicStr` type, that is used to store the current locale. It stores the locale as a raw pointer to an `Arc<String>`. The locale can be read with `AtomicStr::as_str()`. `AtomicStr::as_str()` does not increment the usage counter of the `Arc`.

If the locale is changed in one thread, another thread can have a stale -- possibly already freed -- reference to the stored string.

### DIFF
--- a/crates/rust-i18n-support/RUSTSEC-0000-0000.md
+++ b/crates/rust-i18n-support/RUSTSEC-0000-0000.md
@@ -1,17 +1,21 @@
 ```toml
 [advisory]
 id = "RUSTSEC-0000-0000"
-package = "rust-i18n"
+package = "rust-i18n-support"
 date = "2024-01-19"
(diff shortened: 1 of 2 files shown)


Shnatsel:
Thank you for the report!

I would like to give the maintainers some time to publish a fix, so that the advisory is actionable once it goes live. But if the maintainer doesn't respond, we can merge as-is to at least notify users of the crate.

Kijewski:
A fix was released in <https://github.com/longbridgeapp/rust-i18n/releases/tag/v3.0.1>.

I changed the scope from `rust-i18n` to `rust-i18n-support`, because that's the actually affected crate, and someone could use `rust_i18n_support::AtomicStr` outside of `rust_i18n`.

tarcieri (approved):

PR #1856, opened 2024-01-20
Mark crate `cosmwasm` as unmaintained

webmaster128:
Hi! I'm one of the original maintainers of the cosmwasm crate. This is unmaintained for a long time for for some reason, it is still downloaded. Would be great to be able to mark it as unmaintained.


tarcieri (approved):

tarcieri:
Thanks!

PR #1864, opened 2024-01-24
Add advisories for trillium-http and trillium-client

jbr:


Shnatsel:
Thanks!

CI is failing because it expects the files to be named RUSTSEC-0000-0000.

Shnatsel:
Do I understand correctly that both `trillium-client` and `trillium-http` required fixes and both crates have to be updated to resolve the issue?

jbr:
> Do I understand correctly that both `trillium-client` and `trillium-http` required fixes and both crates have to be updated to resolve the issue?

Yes, if both crates are used. Many trillium-http applications do not use the client, but all trillium-client applications depend on trillium-http.

PR #1866, opened 2024-01-24
snow: Unauthenticated Nonce Increment

mcginty:
Main report: https://github.com/mcginty/snow/security/advisories/GHSA-7g9j-g5jg-3vv3

In short: A logic bug was missed that allowed an internal nonce to be incremented even when a received payload was not successfully decrypted, opening up the possibility of an easy denial-of-service attack for those who have the ability to inject data into the channel that is used for Noise.


Shnatsel:
Sorry about the delay! Somehow this PR has fallen through the cracks. Looks good to me, merging!

PR #1868, opened 2024-01-24
README.md: update maintained image

tarcieri:
It's now 2024.


tarcieri:
Alternatively we could consider removing this, since the database has been actively maintained for several years

PR #1871, opened 2024-01-26
Add unmaintained advisory report for conrod and conrod_core

nickspurry:
The `conrod` crate is explicitly unmaintained. Core functionality was transferred to `conrod_core` and backend functionality split across a number of separate crates such as `conrod_wgpu` around December 2018.

The entire conrod project became explicitly unmaintained in January 2022 when the developer stepped away from the project (https://github.com/PistonDevelopers/conrod/issues/1454).

### DIFF
--- a/crates/conrod/RUSTSEC-0000-0000.md
+++ b/crates/conrod/RUSTSEC-0000-0000.md
@@ -1,0 +1,19 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "conrod"
+date = "2024-01-26"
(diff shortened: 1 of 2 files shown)


amousset:
Thanks for the report. This looks fine but as the author seems responsive could you ask in the issue if they are ok with an "unmaintained" advisory?

nickspurry:
Author of conrod has responded that they agree with unmaintained advisory (see [conrod#1456](https://github.com/PistonDevelopers/conrod/issues/1456#issuecomment-2060266151)).

PR #1882, opened 2024-02-10
Add automation for advisories ID sync

amousset:


Shnatsel:
Any reason not to make it daily? 

We can monitor the PRs for a while and then let it run unattended (as long as CI passes) once see that it works well.

amousset:
Made it daily, and should be ready now.

Shnatsel:
Let's give it a shot!

Issue #1845, opened 2023-12-26
buffer overflow in transpose


caibear:
https://github.com/ejmahler/transpose/issues/11

Shnatsel:
Thank you for the report, and sorry about the delay!

This crate sees a lot of downloads, mostly through `rustfft` which is in turn used by crates exposed to untrusted input such as `symphonia`. It looks like a rather serious issue. I'd be happy to merge an advisory for it.

PR #1890, opened 2024-02-15
Add advisory for buffer overflow in transpose

caibear:
#1845

### DIFF
--- a/crates/transpose/RUSTSEC-0000-0000.md
+++ b/crates/transpose/RUSTSEC-0000-0000.md
@@ -23,4 +23,4 @@ fn transpose<T: Copy>(input: &[T], output: &mut [T], input_width: usize, input_h
 The safety check `input_width * input_height == output.len()` can fail due to `input_width * input_height` overflowing in such a way that it equals `output.len()`.
 As a result of failing the safety check, memory past the end of `output` is written to. This only occurs in release mode since `*` panics on overflow in debug mode.
 
-Exploiting this issue requires the caller to pass input and output slices with mismatching and carefully crafted lengths.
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you! 

I've taken the liberty of editing the advisory directly to fix the `affected` specification and add a line about the circumstances under which it is exploitable. Please let me know if I misunderstood anything.

Shnatsel:
@alex do you agree we should downgrade this to a warning with `informational = "unsound"`? I have a hard time imagining an attacker managing to get mismatching buffer lengths passed just by manipulating the inputs in a reasonable program.

alex:
I would not downgrade it -- since this is a general purpose utility crate, it can be used in many many places, and this really only requires particular data, it doesn't require you to have written particularly unlikely code.

Shnatsel:
I believe the size of the output buffer not matching the input buffer in a carefully crafted way would be caused by unlikely code, and not unlikely input data. But if you still disagree, then let me know and I'll merge it as is.

alex:
Personally I'd rather be conservative and not "hide" vulnerabilities that plausibly can occur without unusual code.

I haven't used this crate myself, so perhaps I'm not familiar with the idioms of how it's generally used, but it seems possible to me.

caibear:
Actually the input and output buffers have to be the same size (or it will panic). The issue is that input_width * input_height can overflow causing data outside input to be copied to outside output.

caibear:
I would change the circumstances to:
```
Exploiting this issue requires the caller to pass input_width and input_height in which multiplying them overflows and and equals the length of input and output.
```

Shnatsel:
Ah, I see I misunderstood. Thank you for the clarification!

@alex I agree this is more serious than I thought, I'll keep it as a vulnerability advisory. Thanks!

PR #1900, opened 2024-02-27
Add non-informational crayon advisory

riking:
The crate already has an informational advisory, and this is clearly a separate issue despite affecting the same code.

Discovered via https://asan.saethlin.dev/ub?crate=crayon&version=0.7.1 . I considered removing `informational=unsound` on the previous advisory, but the issue it describes can be worked around with careful usage, while this issue is unconditional UB in extremely basic usage of the crate.

The crate does not appear to have any reverse dependents on crates.io except its own libraries, so the utility of this advisory is questionable, but clearly someone attempted to use it or else the previous advisory would not have been created.


Shnatsel:
Thank you!

PR #1902, opened 2024-02-28
cassandra-cpp: non-idiomatic use of iterators leads to use after free

kw217:
Hi - raising a RUSTSEC for an issue reported and fixed in cassandra-cpp crate. I am the maintainer of this crate. I've raised a GHSA and requested a CVE via GitHub but it hasn't yet been assigned.


Shnatsel:
Thank you for the report! 

Looks good, merging!

PR #1911, opened 2024-03-05
Add advisory for stack buffer overflow with whoami

sunshowers:
See:

* https://github.com/ardaku/whoami/issues/91
* https://github.com/ardaku/whoami/issues/97

Questions:

* Some other Unix platforms may also be affected. It's hard to enumerate a full list here, but happy to add them as they come to mind for people.

### DIFF
--- a/crates/whoami/RUSTSEC-0000-0000.md
+++ b/crates/whoami/RUSTSEC-0000-0000.md
@@ -1,0 +1,33 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "whoami"
+date = "2024-02-28"
+url = "https://github.com/ardaku/whoami/issues/91"
+informational = "unsound"
+categories = ["denial-of-service", "memory-corruption"]
+keywords = ["buffer-overflow", "stack-buffer-overflow", "cwe-121"]
+license = "CC0-1.0"
+
+[affected]
+# Solaris is also probably affected, as are other Unix OSes that aren't any of:
+# linux, macos, freebsd, dragonfly, bitrig, openbsd, netbsd
+os = ["illumos"]
+functions = { "whoami::username" = ["< 1.5.0"] }
+
+[versions]
+patched = [">= 1.5.0"]
+```
+
+# Stack buffer overflow with whoami on illumos
+
+With older versions of the whoami crate, calling the `username` function leads to an immediate stack
+buffer overflow on illumos. Denial of service and data corruption have both been observed in the
+wild, and the issue is possibly exploitable as well.
+
+This also affects any other Unix platforms that aren't any of: `linux`, `macos`, `freebsd`,
+`dragonfly`, `bitrig`, `openbsd`, `netbsd`.
+
+This issue has been addressed in whoami 1.5.0.
+
+For more information, see [this GitHub issue](https://github.com/ardaku/whoami/issues/91).
--- a/crates/whoami/RUSTSEC-0000-0000.md
+++ b/crates/whoami/RUSTSEC-0000-0000.md
@@ -4,7 +4,6 @@ id = "RUSTSEC-0000-0000"
 package = "whoami"
 date = "2024-02-28"
 url = "https://github.com/ardaku/whoami/issues/91"
-informational = "unsound"
 categories = ["denial-of-service", "memory-corruption"]
 keywords = ["buffer-overflow", "stack-buffer-overflow", "cwe-121"]
 license = "CC0-1.0"
--- a/crates/whoami/RUSTSEC-0000-0000.md
+++ b/crates/whoami/RUSTSEC-0000-0000.md
@@ -6,23 +6,22 @@ date = "2024-02-28"
 url = "https://github.com/ardaku/whoami/issues/91"
 categories = ["denial-of-service", "memory-corruption"]
 keywords = ["buffer-overflow", "stack-buffer-overflow", "cwe-121"]
-license = "CC0-1.0"
 
 [affected]
-# Solaris is also probably affected, as are other Unix OSes that aren't any of:
+# Other Unix OSes that aren't any of these may be affected as well:
 # linux, macos, freebsd, dragonfly, bitrig, openbsd, netbsd
-os = ["illumos"]
+os = ["illumos", "solaris"]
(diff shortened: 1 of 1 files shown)


crates/whoami/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,33 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "whoami"
+date = "2024-02-28"
+url = "https://github.com/ardaku/whoami/issues/91"
+informational = "unsound"
```
sunshowers, line 7: Is this line correct? I was confused by what "informational security advisory" means here. This advisory is definitely more severe than "informational", maybe "high"?

Shnatsel: The `informational` field is meant to be set for low-priority issues, and omitted otherwise.

sunshowers: Thanks, updated! (It would be good to clarify that in [the readme](https://github.com/rustsec/advisory-db?tab=readme-ov-file#advisory-format).)

crates/whoami/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,32 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "whoami"
+date = "2024-02-28"
+url = "https://github.com/ardaku/whoami/issues/91"
+categories = ["denial-of-service", "memory-corruption"]
+keywords = ["buffer-overflow", "stack-buffer-overflow", "cwe-121"]
+license = "CC0-1.0"
```
Shnatsel, line 9:
```suggestion
```

This is the default and can be omitted.

crates/whoami/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,32 @@
...
+
+[affected]
+# Solaris is also probably affected, as are other Unix OSes that aren't any of:
+# linux, macos, freebsd, dragonfly, bitrig, openbsd, netbsd
+os = ["illumos"]
```
Shnatsel, line 14:
```suggestion
os = ["illumos","solaris"]
```

FYI the full lists of recognized platforms can be found at https://docs.rs/platforms/latest/platforms/target/enum.OS.html

Shnatsel (commented):
Thanks for the report! Just a few minor tweaks and it's ready to go!

crates/whoami/RUSTSEC-0000-0000.md
sunshowers: Thanks. I just wasn't sure whether Solaris is affected because I don't have access to their man pages or a running machine.

From https://docs.rs/platforms/latest/platforms/target/enum.OS.html I can't tell what the other Unix platforms are.

I'm also unsure whether the other Apple platforms (iOS, tvOS, watchOS?) are affected. They would probably be affected if it was possible to build and run binaries with whoami on them, but I don't know if that's the case in reality.

sunshowers: I'll add Solaris because the preponderance of evidence suggests it is affected.

crates/whoami/RUSTSEC-0000-0000.md
sunshowers: Done, thanks.

crates/whoami/RUSTSEC-0000-0000.md
Shnatsel: The `platforms` crate doesn't expose the unix-ness right now, even though rustc is aware of it. So I've written a shell script to list all Unix OSs:

```bash
rustc --print=target-list | while read TARGET; do rustc --print=cfg --target="$TARGET" | grep -q 'unix' && rustc --print=cfg --target="$TARGET" | grep 'target_os' | cut -d '=' -f 2-; done | sort -u
```

This is the result:

```
"aix"
"android"
"dragonfly"
"emscripten"
"espidf"
"freebsd"
"fuchsia"
"haiku"
"horizon"
"hurd"
"illumos"
"ios"
"l4re"
"linux"
"macos"
"netbsd"
"nto"
"openbsd"
"redox"
"solaris"
"tvos"
"vita"
"vxworks"
"watchos"
```

sunshowers: Ahh, so some of those of those platforms would also be affected and I guess the advisory should be updated. Any platform where the `passwd` struct doesn't match [Linux's](https://man7.org/linux/man-pages/man3/getpwuid.3.html), and isn't part of the other excluded list of platforms, is affected.

I went through a few of them by hand:

* AIX is accidentally okay because [the passwd struct](https://www.ibm.com/docs/en/aix/7.2?topic=g-getpwent-getpwuid-getpwnam-putpwent-setpwent-endpwent-subroutine#getpwent__d6a64f2310jani__title__1) matches Linux.
* Fuchsia is okay, I think? I can't find the source code but [this struct](https://fuchsia.dev/reference/cpp/struct/passwd) has all the fields, though they're 
* Android is okay per [this struct](https://android.googlesource.com/platform/bionic/+/882b8af/libc/include/pwd.h#99).

It's hard to go through every single platform, so if you have suggestions for a better way to do this, or some kind of other catch-all, I'd love that.

Shnatsel: The advisory db format doesn't support platform selectors, only a list.

Sadly I am not aware of a resource that would aggregate and/or diff the C APIs from different platforms.

AldaronLau:
Would it be important to note that this doesn't affect whoami < 0.5.3 (from 5 years ago)?

This is also a problem on freebsd, dragonfly, bitrig, openbsd, netbsd with whoami >= 0.5.3, < 1.0.1 (not sure if that would count separately from this or not).

AldaronLau:
Also just noticed this affects not only `whoami::username()`, but additionally `whoami::realname()`, `whoami::username_os()`, and `whoami::realname_os()`.

Shnatsel:
Yes, that would be useful. We have the `unaffected` field for that.

sunshowers:
Thanks. I'll send an update shortly.

PR #1913, opened 2024-03-05
update RUSTSEC-2024-0020 with additional information

sunshowers:
Add information about more platforms affected, per https://github.com/rustsec/advisory-db/pull/1911#issuecomment-1978963801.


AldaronLau (commented):
Thanks for this update; looks good to me!

Shnatsel:
Thanks! Merged.

PR #1918, opened 2024-03-06
eyre: Parts of Report are dropped as the wrong type during downcast

dtolnay:
- https://github.com/eyre-rs/eyre/issues/141
- https://github.com/eyre-rs/eyre/pull/143


amousset:
Thanks for the report.

PR #1926, opened 2024-03-26
add CVE-2024-28854 for tls-listener

doy-materialize:


Shnatsel:
I see this originates from GHSA but only uses data from CVE which is released in the public domain, so we don't need an explicit license grant in this case.

Sorry for the review delay! Merging now.

PR #1934, opened 2024-04-08
Add unmaintained advisory for `mdbook-template`

sgoudham:
Hi :wave:, I'm the author of `mdbook-template` and no longer wish to maintain the package.

I believe this is the format for adding an unmaintained advisory, please let me know if details need to changed and/or are missing!


riking:
> This crate is not used as a dependency in any other crate on crates.io.

I don't see any ecosystem dependents, so I don't think there's any value in including this in the db?

sgoudham:
Ah sorry I didn't realise that was a requirement, happy to close this PR!

PR #1946, opened 2024-04-19
Add rustls GHSA-6g7w-8wpp-frhj

ctz:


cpu (approved):

khuey:
The advisory says `rustls-tokio` is unaffected but the correct name of the crate is `tokio-rustls`, no?

cpu:
> The advisory says rustls-tokio is unaffected but the correct name of the crate is tokio-rustls, no?

Indeed, this was an oversight. It's [tokio-rustls](https://github.com/rustls/tokio-rustls/blob/7fd44d9a884744eb6df0fbb930db3d8ea586f584/Cargo.toml#L2).

PR #1968, opened 2024-05-21
kalker advisory for parsing panic

nyw0102:


crates/kalker/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
...
+url = "https://github.com/PaddiM8/kalker/pull/153"
+categories = ["denial-of-service"]
+#related = ["CVE-2022-31100"]
+
+[versions]
+patched = [">=2.2.0"]
+```
+# Panic due to inappropriate string
+
+When trying to parse untrusted string through kalker, kalker may pancic,
```
bjorn3, line 15: *panic

nyw0102: Thank you! I revised the report!

riking:
I don't think that a panic in a library not used outside a single application binary (no dependents on crates.io) deserves an advisory. There's no indications that this is used as a networked service (and if only the developers host it as a network service, it can be handled internally).

In other words, use binary-focused advisory systems like CVE and customer mailing lists for this (and perhaps not even that, it's just a user input panic).

Shnatsel:
I agree with @riking, so I'm going to go ahead and close this.

Panics are generally not advisory-worthy on their own, and this is not a high-stakes environment like a networking server where it could potentially matter.

PR #1974, opened 2024-06-03
Add advisory for nano-id crate

ciffelia:
This patch adds an advisory for [nano-id](https://crates.io/crates/nano-id) crate. I reported this vulnerability to the maintainer of the crate (@fundon) and got permission to disclose this.


Shnatsel:
Thanks!

PR #1979, opened 2024-06-10
Create `mmap` unmaintained advisory

ppershing:
Note: Did not ask the author about the status explicitly but the repository is archived for over two years so it should be a good indication


Shnatsel:
I understand the main alternatives are https://crates.io/crates/memmap2 and https://crates.io/crates/memmapix, with memmap2 being the by far the more popular of the two.

ppershing:
Thanks, I clarified that memmap2 seems to be the main alternative.

amousset:
Thanks!

PR #1985, opened 2024-06-26
Add advisory for sequoia-openpgp DoS vulnerability.

teythoon:
There is a denial-of-service vulnerability in sequoia-openpgp, our crate providing a low-level interface to our OpenPGP implementation. When triggered, the process will enter an infinite loop.

The RawCertParser does not advance the input stream when encountering unsupported cert (primary key) versions, resulting in an infinite loop.


amousset:
Thanks!

amousset (approved):

Issue #1994, opened 2024-07-03
opentelemetry_api is removed from the github project


jayvdb:
https://crates.io/crates/opentelemetry_api doesnt mention that it is deprecated.

It was removed from their repo in https://github.com/open-telemetry/opentelemetry-rust/pull/1226 , merged August 2023

As a result, it is now unmaintained.

hdost:
> 
> As a result, it is now unmaintained.
> 
> 

It was migrated from `opentelemtry_api` to `opentelemetry` the transition should be a migration a simple Search and Replace (depending on which version you upgrade to)

PR #1995, opened 2024-07-03
opentelemetry_api merged into opentelemetry crate

jayvdb:
Closes https://github.com/rustsec/advisory-db/issues/1994


djc:
I was involved as a maintainer at the time, this makes sense to me.

PR #1997, opened 2024-07-03
Advisory for CVE-2024-35197 (device names) in gix-ref, gix-index, gix-worktree

EliahKagan:
This adds notices for the [Windows device name handling](https://cwe.mitre.org/data/definitions/67.html) vulnerability CVE-2024-35197 (https://github.com/Byron/gitoxide/security/advisories/GHSA-49jc-r788-3fc9). This is a separate vulnerability from the one that #1996 is about—and I cannot open a single PR for both because they both have `RUSTSEC-0000-0000.md` files in two of the same locations until IDs are assigned—but it is likewise discussed in https://github.com/Byron/gitoxide/discussions/1437 (cc @Byron).

The advisory text (long description) is what I wrote for https://github.com/Byron/gitoxide/security/advisories/GHSA-49jc-r788-3fc9 and is essentially the same as in [the global advisory](https://github.com/advisories/GHSA-49jc-r788-3fc9). This is analogous to the situation in #1996, albeit for different advisories/vulnerabilities. Both there and here, it is and has always been my intention that this text be dedicated to the public domain (with CC0).

Some of the same considerations there apply here as well, such as the possible need to create multiple RUSTSEC advisories since multiple crates are affected in a way that is not fully independent. However, here there is another factor: `gix-ref` is affected in a *very* different way from the other crates.

That is because this vulnerability has two clearly distinct aspects, or variants: the effect on references, which causes `gix-ref` to be a directly affected crate; and the effect on paths, which is wholly independent of `gix-ref` and which the advisory text describes behaviorally in terms of `gix-worktree-state`, but for which I consider the primary affected crates to be `gix-index` and `gix-worktree`.

This bifurcation may justify altering the RUSTSEC advisory text so that the different affected crates are described differently, with one long description for `gix-ref` that covers only the effect on references, and a separate long description for `gix-index` and `gix-worktree` that covers only the effect on paths. I am unsure if this is justified, but if so then I would be pleased to make that change. Unlike most advisory text changes, this would not require a corresponding change in the repo-local or global GHSA advisory text (since those notices would still need to combine the two aspects of the vulnerability into one description).

I wasn't sure what, if anything, to put here for `categories` or `keywords`. Although I'd prefer to list a category if one is clearly correct, I'm not sure any properly applies for this or most [CWE-67](https://cwe.mitre.org/data/definitions/67.html) vulnerabilities. (Few such vulnerabilities seem to have been reported in recent years; it looks like this one is [the only one in GHSA](https://github.com/advisories?query=cwe%3A67).) A possible impact is denial of service, either by disrupting interaction with external devices, or by writing a large amount of text to a terminal. But I don't think DoS is the main concern for this vulnerability. For now I have not listed any categories.

### DIFF
--- a/crates/gix-index/RUSTSEC-0000-0000.md
+++ b/crates/gix-index/RUSTSEC-0000-0000.md
@@ -1,0 +1,91 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "gix-index"
+date = "2024-05-22"
+url = "https://github.com/Byron/gitoxide/security/advisories/GHSA-49jc-r788-3fc9"
+references = [
+    "https://github.com/advisories/GHSA-49jc-r788-3fc9",
+    "https://nvd.nist.gov/vuln/detail/CVE-2024-35197",
+]
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:N/I:L/A:L"
+aliases = ["CVE-2024-35197", "GHSA-49jc-r788-3fc9"]
+license = "CC0-1.0"
+
+[affected]
+os = ["windows"]
+
+[versions]
+patched = [">= 0.33.0"]
+```
+
+# Refs and paths with reserved Windows device names access the devices
+
+### Summary
+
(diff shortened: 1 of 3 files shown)


EliahKagan:
I made a mistake in copying some of the Markdown. I'll fix that in a couple of minutes.

tarcieri (approved):
It's fine not to have categories or keywords. We can add them later if someone opens a PR to do so.

PR #1999, opened 2024-07-06
Add advisory for unsound problem in `wasmtime_jit_debug`

safe4u:
A soundness bug in `wasmtime_jit_debug` that wrongly marks the function `dump_code_load_record` as 'safe'.
The issue report: https://github.com/bytecodealliance/wasmtime/issues/8905

### DIFF
--- a/crates/wasmtime_jit_debug/RUSTSEC-0000-0000.md
+++ b/crates/wasmtime_jit_debug/RUSTSEC-0000-0000.md
@@ -1,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "wasmtime_jit_debug"
+date = "2024-07-06"
+url = "https://github.com/bytecodealliance/wasmtime/issues/8905"
+informational = "unsound"
+categories = ["memory-exposure"]
+
+[affected]
+functions = { "wasmtime_jit_debug::perf_jitdump::JitDumpFile::dump_code_load_record" = ["<= 22.0.0"] }
+
+[versions]
+patched = []
+```
+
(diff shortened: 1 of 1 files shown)


kornelski:
It looks like it's been patched in v27 

https://github.com/bytecodealliance/wasmtime/commit/b5e31a5c33725ab8a7dfbe8505c56b5cf282ffed

can you mark it as patched in the advisory?

kornelski (changes requested):
Needs to be updated with the patched version

safe4u:
> Needs to be updated with the patched version

Done. Thanks for your correction.

kornelski (approved):

djc:
@alexcrichton any feedback on this one?

alexcrichton:
Ok we discussed this yesterday and the discussion culminated in https://github.com/bytecodealliance/wasmtime/pull/10963. We don't plan on retroactively applying this change to older branches but it should at least hopefully mitigate things going forward. In the meantime this seems reasonable to merge to me, I don't expect the fallout to be that large.

djc:
@alexcrichton any feedback on the advisory contents? Should we add a note about this being an internal crate?

alexcrichton:
Yeah that would probably be best, something along the lines of:

> **Note**: this is an internal-only crate in the Wasmtime project not intended for external use and is more strongly signaled nowadays as of [bytecodealliance/wasmtime#10963](https://github.com/bytecodealliance/wasmtime/pull/10963). Please open an issue in Wasmtime if you're using this crate directly.

djc:
@safe4u can you add this note to the advisory?

safe4u:
Thank you all for your attention to this minor problem : )

PR #2000, opened 2024-07-06
Add advisory for unsound problem in `scanner-rust`

safe4u:
`ScannerU8SliceAscii` creates and parses non-utf8 str, which is unsound and might lead to undefined behaviors.
The details are described in https://github.com/magiclen/scanner-rust/issues/2.

### DIFF
--- a/crates/scanner-rust/RUSTSEC-0000-0000.md
+++ b/crates/scanner-rust/RUSTSEC-0000-0000.md
@@ -1,0 +1,24 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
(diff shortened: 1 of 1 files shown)


kornelski (approved):

alexanderkjall:
I see that this report is disputed by the crate author, and I didn't manage to reproduce any problem when I wrote a small fuzzer for the crate.

Maybe this is a false positive?

safe4u:
After further code review by the crate maintainer, the originally reported issue is considered a false positive.

Based on this clarification, we are closing this PR. Thank you for the review and discussion.

PR #2001, opened 2024-07-07
Add advisory for unsound problems in `workflow-core`

safe4u:
The util functions `buffer_as_slice` and `buffer_as_slice_mut` in crate `workflow-core` could create illegal slice.
The details are described in https://github.com/workflow-rs/workflow-rs/issues/11

### DIFF
--- a/crates/workflow-core/RUSTSEC-0000-0000.md
+++ b/crates/workflow-core/RUSTSEC-0000-0000.md
@@ -1,0 +1,26 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "workflow-core"
+date = "2024-07-07"
+url = "https://github.com/workflow-rs/workflow-rs/issues/11"
+informational = "unsound"
+categories = ["memory-exposure", "memory-corruption"]
+
+[affected.functions]
+"workflow_core::utils::buffer_as_slice" = ["<= 0.13.3"]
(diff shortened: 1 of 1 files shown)


kornelski:
Still unsound in 0.18.0

kornelski (approved):

kornelski (changes requested):
The report is technically accurate.

Sadly, there's no fixed version, and it looks like development on the crate has ground to a halt.

kornelski (approved):

pduhandeh-collab:
Thanks for all

djc:
@aspect @surinder83singh can you talk about the maintenance status of the workflow-core crate? If it's unmaintained, it would be good to communicate this.

aspect:
Thanks for tagging. This is great.

No, the crate is very much maintained and is critical to some well maintained mainstream applications.

This is my fault as I have basically disregarded this assuming that this is AI auto-detection and this crate contains general-purpose toolbox of different handy utils ... not really used by anyone (and apparently broken :)). They just sit in one of the submodules. 

I am unfortunately swamped and can't look at this right now or in the coming days. There is a maintenance pass that is needed in related crates (it's a large framework). These functions should be just killed off.

I will add this to my general todo list and address this eventually.

djc:
@aspect okay, so is it okay if we just publish this advisory without fixed versions for now? We can always add those later as they become available.

PR #2025, opened 2024-07-23
Add advisory for object_store credentials leak via logs

jdstrand:


Shnatsel:
looks great, thank you!

PR #2027, opened 2024-07-24
Unsoundness notice for gix-attributes (kstring integration)

EliahKagan:
gix-attributes was found by @ssbr to be unsound, as reported in https://github.com/Byron/gitoxide/issues/1460. This adds an informational notice for that, as discussed in comments there (cc @Byron).

The text of the notice is taken from that issue (083656c), with slight modification (1a50df1).

It looks like the affected code, having been introduced in https://github.com/Byron/gitoxide/pull/400, was most likely present in all published versions of the crate prior to the fix in 0.22.3 (which was one of the bugs fixed in https://github.com/Byron/gitoxide/pull/1462). So I have not specified a minimum affected version.

### DIFF
--- a/crates/gix-attributes/RUSTSEC-0000-0000.md
+++ b/crates/gix-attributes/RUSTSEC-0000-0000.md
@@ -1,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "gix-attributes"
+date = "2024-07-24"
+url = "https://github.com/Byron/gitoxide/issues/1460"
+informational = "unsound"
+
+[versions]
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you! Looks good to me.

To clarify, why is this a draft? Is there anything else you want to do before we merge this?

EliahKagan:
> *[...]* To clarify, why is this a draft? *[...]*

I think it didn't need to be a draft -- I had it as a draft while I was looking over the PR text and checking CI results.

> Looks good and I see the maintainer is on board. Merging. Thank you!

Thanks for the very quick response! 😄

PR #2039, opened 2024-08-15
Advisory for sqlx <= 0.8.0

abonander:
https://github.com/launchbadge/sqlx/issues/3440

I'm not sure whether to open separate advisories for each affected `sqlx-*` crate. Ultimately, only `sqlx` is meant to be used directly.

### DIFF
--- a/crates/sqlx/RUSTSEC-0000-0000.md
+++ b/crates/sqlx/RUSTSEC-0000-0000.md
@@ -1,0 +1,56 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "sqlx"
+date = "2024-08-15"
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thanks! Looks good to me.

Would you prefer to have this go live ASAP, or once the fix is published?

abonander:
That's an open question. This is my first time filing an advisory so I don't know what's preferred here.

I suppose if people's builds start bailing out due to `cargo deny` and there isn't a patch they can immediately upgrade to, they might be pretty annoyed.

But this is technically a 0-day so getting the information out there may be more important.

Shnatsel:
I'll go ahead and publish it then. Thank you!

PR #2042, opened 2024-08-21
report unsound issue in bcc

llooFlashooll:
For more information, see: [rust-bcc/issues/200](https://github.com/rust-bpf/rust-bcc/issues/200)
In the meantime, bcc will no longer be maintained. Users are encouraged to migrate to libbpf-rs.

### DIFF
--- a/crates/bcc/RUSTSEC-0000-0000.md
+++ b/crates/bcc/RUSTSEC-0000-0000.md
@@ -1,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "bcc"
+date = "2024-08-21"
+url = "https://github.com/rust-bpf/rust-bcc/issues/200"
+#informational = "unmaintained"
+categories = ["memory-corruption"]
+keywords = ["misalignment"]
+
+[versions]
+patched = []
+```
+
+# Unsound usage of unsafe implementation from `Vec<u8>` to `T`
+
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
```rust
fn null_or_mut_ptr<T>(s: &mut Vec<u8>) -> *mut T {
    if s.capacity() == 0 {
        ptr::null_mut()
    } else {
        s.as_mut_ptr() as *mut T
    }
}
```

This is safe function containing safe code. I do not see any credible way this could be unsound.

Skgland:
Your example while indeed incorrect rust does not discredit `null_or_mut_ptr`.
The source of the incorrectness is the unsafe block added in your example not `null_or_mut_ptr`.
Pointers created in a safe manner aren't inherently unsafe, even if they might be nonsensical or easy to misuse in unsafe code.

As fare as I can tell `null_or_mut_ptr` in `rust-bcc` is only used thrice.
Twice as an argument to `bpf_prog_load` and once as an argument to `bcc_prog_load`.
At a quick glance (didn't check all versions) the relevant argument is of type `c_char` in each case which is an alias for either `u8` or `i8` and as such layout compatible with `u8`.

Not involving a generic i.e. using `c_char` instead of a generic T would be better to not accidentally cause UB/unsoudness, but I don't see an immediate problem as is.

amousset:
An unmaintained advisory has been published for `bcc`, so I think we can close this issue now.

PR #2043, opened 2024-08-21
report unsafe transmute in allocator-api2

llooFlashooll:
For more information, see: [allocator-api2/issues/15](https://github.com/zakarumych/allocator-api2/issues/15).


Skgland:
This function exists (unstably) in the same way in the standard library ([core::ptr::without_provenance_mut](https://doc.rust-lang.org/core/ptr/fn.without_provenance_mut.html)).

PR #2048, opened 2024-08-23
Fill an advisory for protocol level injections for diesel

weiznich:
This is essentially the same as https://github.com/rustsec/advisory-db/pull/2039 but for diesel.

I plan to release the fixed version later today. I would like to hold back merging this advisory until the 2.2.3 version of diesel is actually released. As soon as I've performed the release I will leave another comment here.


weiznich:
Diesel 2.2.3 is published, so this PR is now ready to be merged

Shnatsel:
Thanks!

PR #2055, opened 2024-08-31
Advisory for GHSA-v26r-4c9c-h3j6 (config scopes) in gix-path

EliahKagan:
This adds a notice for GHSA-v26r-4c9c-h3j6 in `gix-path`.

cc @Byron

PR #2061, opened 2024-09-02
Add CVE number for RUSTSEC-2024-0367 (config scopes)

EliahKagan:
This adds CVE-2024-45305 to `aliases` for RUSTSEC-2024-0367.

No CVE had been issued for that vulnerability when it was added to the RUSTSEC database in #2055, but it has been assigned since. (This CVE number can, if desired, be confirmed at https://github.com/Byron/gitoxide/security/advisories/GHSA-v26r-4c9c-h3j6, under "CVE ID" on the sidebar.)

Although I suspect an entry in the GitHub Advisory Database will be created, and as GHSA-v26r-4c9c-h3j6 is the repo-level GHSA that URL will likely be https://github.com/advisories/GHSA-v26r-4c9c-h3j6, I have refrained from adding that here because the page does not yet exist and I don't want to add inaccurate information. I'll open another PR if/when there is more to add.

cc @Byron

### DIFF
--- a/crates/gix-path/RUSTSEC-2024-0367.md
+++ b/crates/gix-path/RUSTSEC-2024-0367.md
@@ -4,9 +4,9 @@ id = "RUSTSEC-2024-0367"
 package = "gix-path"
 date = "2024-08-31"
 url = "https://github.com/Byron/gitoxide/security/advisories/GHSA-v26r-4c9c-h3j6"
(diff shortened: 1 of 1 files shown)


EliahKagan:
The anticipated entry has been added to the GitHub Advisory Database, and I've opened #2063, which adds the reference.

PR #2068, opened 2024-09-05
Advisory for `GHSA-rwq6-crjg-9cpw` in `ic-cdk`

venkkatesh-sekar:
Creating a RUSTSEC advisory for [Github advisory](https://github.com/advisories/GHSA-rwq6-crjg-9cpw) due to a memory leak bug in `ic-cdk` crate


Shnatsel:
Unlike in Cargo, we need to specify the `^` operator explicitly to express "0.8.2 and later but before 0.9.0". I went ahead and made the change.

Everything else looks good. Thank you!

PR #2080, opened 2024-09-15
Lexical is maintained again, unsoundness is fixed

Shnatsel:
Announcement: https://redd.it/1fgyopo


Alexhuszagh:
The correctness issues should have been patched as well.
https://github.com/Alexhuszagh/rust-lexical/issues/91

PR #2082, opened 2024-09-16
Add lexical-core advisory.

Alexhuszagh:
This adds in an additional security advisory for `lexical` and adds in the same for `lexical-core`.

Closes #2080 2081


Shnatsel:
There is no need to list alternatives now that a patched version is available. I've removed that paragraph from the advisory text.

For technical reasons we need to have a separate ID for each crate, so I've changed the ID to 0000-0000 so that the ID assignment mechanism would work properly.

Shnatsel:
CI passes, I think this is good to go. Thank you, both for updating `lexical` and for filing this advisory!

Issue #2087, opened 2024-09-23
atty is unmaintained


jayvdb:
https://github.com/softprops/atty/issues/50 was https://osv.dev/vulnerability/RUSTSEC-2021-0145

https://github.com/softprops/atty/issues/57 was created Oct 2022 , no explicit response from @softprops on the issue, however the user is active, and has declared the project is unmaintained with a commit to the README.md

PR #2089, opened 2024-09-25
Mark atty as officially unmaintained

BlackHoleFox:
Should be fairly uncontroversial as it explicitly meets the [unmaintained policy's](https://github.com/rustsec/advisory-db/blob/main/HOWTO_UNMAINTAINED.md#definition-of-an-unmaintained-crate) of explicit:
> Explicitly unmaintained: a crate's author has declared that they are no longer maintaining a particular crate.

The author of `atty` published [this README update](https://github.com/softprops/atty/commit/5bfdbe9e48c6ca6a4909e8d5b04f5e843a257e93#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R5) making the maintenance status official:
>  This is crate is no longer maintained and users are encouraged to prefer the equivalant functionality of [std.io.IsTerminal]...

Since the previous maintainer was clear about alternatives, I made them more explicit in the advisory instead of the usual "these have not been looked and might not work, etc".

Closes https://github.com/rustsec/advisory-db/issues/2087

### DIFF
--- a/crates/atty/RUSTSEC-0000-0000.md
+++ b/crates/atty/RUSTSEC-0000-0000.md
@@ -1,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "atty"
+date = "2024-09-25"
+informational = "unmaintained"
+url = "https://github.com/softprops/atty/issues/57"
+
+[versions]
+patched = []
+```
+
+# `atty` is unmaintained
+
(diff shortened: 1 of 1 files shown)


BlackHoleFox:
cc @pinkforest since you opened the original issue upstream asking about the crate's maintenance status.

PR #2100, opened 2024-10-11
Add disclosure for hasbrown's borsh encoding

kayabaNerve:
I wrote this as fixed in 0.16 due to that being the plan documented here: https://github.com/rust-lang/hashbrown/pull/570#issuecomment-2406017398

Until that PR is merged and 0.16 is released, this cannot be merged/must be edited to be described as unpatched.

### DIFF
--- a/crates/hashbrown/RUSTSEC-0000-0000.md
+++ b/crates/hashbrown/RUSTSEC-0000-0000.md
@@ -8,7 +8,7 @@ categories = []
 keywords = ["borsh"]
 
 [versions]
-patched = []
+patched = [">= 0.15.1"]
 unaffected = ["<= 0.14"]
 
 [affected]
@@ -23,3 +23,5 @@ It also did not perform canonicty checks on decoding.
 
 This can result in consensus splits and cause equivalent objects to be
 considered distinct.
+
+This was patched in 0.15.1.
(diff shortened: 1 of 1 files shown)


clarfonthey:
There was a proposal to release the fix as 0.15.1 and just yank 0.15.0, so, might be worth verifying which of the two versions is going to be used for the fix.

kayabaNerve:
I've remove the patched notice so this can be filed now (warning other people off from adopting this functionality) while we continue to wait for a patched version.

kayabaNerve:
Sorry for my delay in updating this. It correctly notes when patched now.

kornelski (approved):

crates/hashbrown/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,27 @@
...
+id = "RUSTSEC-0000-0000"
+package = "hashbrown"
+date = "2024-10-11"
+url = "https://github.com/rust-lang/hashbrown/issues/576"
+categories = []
+keywords = ["borsh"]
+
+[versions]
+patched = [">= 0.15.1"]
+unaffected = ["<= 0.14"]
```
Turbo87, line 12: I'm not sure how this is supposed to be interpreted, but at least cargo-deny flags `0.14.5` as being affected now. Maybe `unaffected = ["< 0.15.0"]` would be better?

Turbo87: I've opened https://github.com/rustsec/advisory-db/pull/2157 with that proposal

PR #2103, opened 2024-10-21
report serial crates as unmaintained

solomoncyj:


crates/serial/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,14 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "serial"
+date = "2017-07-02"
+url = "https://github.com/dcuddeback/serial-rs/issues/75"
```
CommanderStorm, line 6: Your issue does not meet the bar set out by the [unmaintainedness-policy](https://github.com/rustsec/advisory-db/blob/main/HOWTO_UNMAINTAINED.md), but this issue does.

```suggestion
url = "https://github.com/dcuddeback/serial-rs/issues/71"
```

crates/serial/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,14 @@
...
+informational = "unmaintained"
+
+[versions]
+patched = []
+```
+
+# no updates to git rpo and crates
+no updates have been pushed to the serial family of crates for over 7 years
```
CommanderStorm, line 14:
```suggestion
# `serial` crate is unmaintained

The [`serial`](https://crates.io/crates/serial) crate is no longer maintained.

Last release was on 2017-07-02.

## Possible alternatives

Consider using an alternative, for instance the blocking librarys:

- [serial2](https://crates.io/crates/serial2)
- [serialport](https://crates.io/crates/serialport)

or async alternatives:

- [mio-serial](https://crates.io/crates/mio-serial)
- [tokio-serial](https://crates.io/crates/tokio-serial)
```

CommanderStorm (changes requested):
Hi, what do you think about these changes?

CommanderStorm:
Also note that the title of the advisory does not match its content. Is this intentional?

solomoncyj:
> Also note that the title (author vs crate) of the advisory does not match its content. Is this intentional?

i am aming to shutdown the root crate, which should implicitly mark the dependencies as unmaintained

CommanderStorm:
Just that we are clear on communication:

So
- [`serial-core`](https://crates.io/crates/serial-core),
- [`serial-posix`](https://crates.io/crates/serial-posix) and
- [`serial-windows`](https://crates.io/crates/serial-windows)? 

=> i.e. not the other "crates by dcuddeback" like for example
- [`CoreFoundation-sys`](https://crates.io/crates/CoreFoundation-sys) ([maintenance issue](https://github.com/dcuddeback/core-foundation-sys/issues/11))

> shutdown the root crate, which should implicitly mark the dependencies as unmaintained

I don't think this is how this works.
Say I only put `serial-posix` into my `Cargo.toml` file
=> an advisory on `serial` would not appear for me, as `serial-posix` is just a dependency of `serial`, not the other way around...

kornelski (approved):

PR #2104, opened 2024-10-24
Mark two pqcrypto- crates as unmaintained

thomwiggers:
Since the official standards of these draft algorithms were released, they are replaced by crates that track the standards versions.


amousset:
Thanks!

PR #2105, opened 2024-10-25
Add missing information about fixed versions

VorpalBlade:
According to https://github.com/Xudong-Huang/may/issues/88, this has been fixed since a long time. Add the missing `patched`


alex (approved):
thanks

VorpalBlade:
Whoa that was a quick merge. Too quick in fact:

Actually, there is no 0.3.19 (crate is on 0.1.x) and the issue was fixed in 2021. I think the version in the linked bug report was probably typoed. I will try to dig and find when it was *actually* fixed and post a followup PR.

VorpalBlade:
It seems it should have been version 0.1.8 of `may_queue` that fixed this (and the version number above refers to the main `may` crate).

PR #2106, opened 2024-10-25
Fix incorrect fixed version for may_queue

VorpalBlade:
According to https://github.com/Xudong-Huang/may/issues/88, this was fixed in 0.3.19, but that refers to the main "may" crate. The correct version for the "may_queue" sub crate is 0.1.8 (based on manual checking the bounds for the affected type on docs.rs)

My bad in the previous PR (#2105) for trusting the number written in the relevant bug report.


alex (approved):
Thanks for catching this. Maybe we need to be validating these versions programatically...

PR #2133, opened 2024-11-14
Stack overflow during decompression with malicious input in zlib-rs

rnijveld:


Shnatsel:
Thank you for the report!

Could you make it more clear that this is purely a denial-of-service issue? It would be nice to put "Denial of service" in the title and/or in the first sentence of the description.

"Stack overflow" is superficially similar to "stack buffer overflow" which may lead people not well versed in the distinction to conclude that this issue is more severe than it actually is.

Also please drop the comments from the TOML part of the file.

rnijveld:
I've updated the description and title to hopefully make that point a little more clear. And the comments should now all have been removed.

kornelski (approved):

PR #2144, opened 2024-11-25
Fix unaffected version in RUSTSEC-2024-0399.md

conradludgate:
https://github.com/rustsec/advisory-db/pull/2143#issuecomment-2497484693

PR #2145, opened 2024-11-25
RUSTSEC-2024-0399: fix 0.22 version constraint

ctz:
Apologies for the noise. Verified this covers the correct versions with `rustsec-admin list-affected-versions`


ctz:
midair collision with #2144

PR #2147, opened 2024-11-27
Add `ruzstd` uninit/out-of-bounds memory reads advisory

paolobarbolini:
Creates the advisory for the vulnerability reported at https://github.com/KillingSpark/zstd-rs/issues/75 and potentially fixed by https://github.com/KillingSpark/zstd-rs/pull/76.

Opening as draft while we wait for testing and response from the maintainer.


Shnatsel:
Thank you for the report!

We normally use the `informational = unsound` variant for unsound APIs that can be misused, but I see the function is not exposed via the public API.

AFAIK it's not settled whether it is OK to copy uninitialized u8 values without ever observing them, and the compiler currently does not exploit that for optimization purposes. 

However, I am concerned about the risk of exposing contents of uninitialized memory via a crafted input. I would like to hear from the maintainers about the potential impact of this issue.

KillingSpark:
Hi, maintainer of ruzstd here:

As far as I can tell this could, with specially crafted inputs, result in the decoded output containing up to 15 bytes of heap memory that didn't belong to the allocation of the ringbuffer used during decoding. More specifically up to 15 bytes directly after that allocation.

So this could maybe be used to exfiltrate data from a process, if the process
* Decodes an attacker controlled archive
* Passes the decoded contents back to the attacker

KillingSpark:
Can I ask what the recommended/common next steps are here? This is the first handling such a situation as a maintainer.

My guess would be:

* Find the released versions that include the error
* Yank them from crates.io
* Make new releases with a bumped minor that only fix this and change nothing else
* Delete the affected releases on github

Am I missing something?

paolobarbolini:
Cross posting here too that the [v0.7.3](https://github.com/KillingSpark/zstd-rs/releases/tag/v0.7.3) release fixing the issue is out

crates/ruzstd/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,23 @@
...
+```
+
+# `ruzstd` uninit and out-of-bounds memory reads
+
+Affected versions of `ruzstd` miscalculate the length of the allocated
+and init section of its internal `RingBuffer`, leading to uninitialized
+or out-of-bounds reads in `copy_bytes_overshooting` of up to 15 bytes.
+
+This may result in up to 15 bytes of memory contents being written
+into the decoded data when ecompressing a crafted archive.
```
paolobarbolini, line 22:
```suggestion
into the decoded data when decompressing a crafted archive.
```

PR #2160, opened 2024-12-04
Report unsoundness in pprof

shinmao:


crates/pprof/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,28 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "pprof"
+date = "2024-12-04"
+url = "https://github.com/tikv/pprof-rs/issues/232"
+references = ["https://github.com/tikv/pprof-rs/pull/255"]
+categories = ["denial-of-service"]
```
kornelski, line 8: Thanks for finding the issue.

I think the root cause here is a soundness issue, so it's not a DoS issue. The panic just happens to be from debug assertions that sometimes catch the issue, but that's a symptom, not the cause of the problem.

shinmao: Thanks for pointing it out. I just updated it to "unsound". Please check.

kornelski (approved):

PR #2162, opened 2024-12-05
report unsoundness in candle-core

shinmao:


Shnatsel:
Thank you for the report!

I see there has been recent activity on the upstream issue, and I'd prefer to publish an actionable advisory once the issue is fixed. I see the report itself is quite old, so if the issue still there in a week we may proceed regardless.

shinmao:
Yes. I have reported this issue for a while. I agree the report of unsoundness if it still keeps for one more week 👍

shinmao:
@Shnatsel I think we can release the unsound advisory since the issue is still open without update since last week. How do you think about it?

PR #2163, opened 2024-12-09
idna accepts Punycode labels that do not produce any non-ASCII when decoded

hsivonen:


hsivonen:
It's unclear to me what should be done about the category field. The field is optional but also seems to be part of inclusion criteria. This isn't about privilege escalation in `idna` itself, but the potential for privilege escalation is the point of issuing an advisory.

Shnatsel:
That's a very clear and thorough write-up. Thank you! Looks good, merging.

PR #2164, opened 2024-12-09
Add advisories for unmaintained gtk-rs GTK3 bindings

sdroege:


Shnatsel:
Thank you for the PR!

CI fails because there is no `gdk-x11-sys` package on crates.io.

sdroege:
> CI fails because there is no `gdk-x11-sys` package on crates.io.

Thanks, there was a `-` too much :)

Shnatsel:
Could you add a note that this is coming from the actual maintainers, to make the source of the notice more clear? Mentioning that the repositories have been archived might also be helpful. Thank you!

sdroege:
Done, good idea

Shnatsel:
Thanks! Looks good, merging.

PR #2167, opened 2024-12-09
Add advisories for unmaintained gtk-layer-shell GTK3 bindings

pentamassiv:
I am the maintainer of the `gtk-layer-shell` and `gtk-layer-shell-sys` crates. The foundation of `gtk-layer-shell` are the `gtk` crates. It does not make sense to continue maintaining the crates if the foundation is unmaintained (https://github.com/rustsec/advisory-db/pull/2164), so I also stopped the development of them and marked them as unmaintained. Users should use GTK4 and the `gtk4-layer-shell` crate instead.


Shnatsel:
Thanks!

PR #2169, opened 2024-12-12
Protobuf DoS

DarkaMaul:
This (public) advisory follows two emails sent on August 9 and October 3rd.

The crate is affected by the same vulnerability as described in https://github.com/protocolbuffers/protobuf/security/advisories/GHSA-735f-pc8j-v9w8

### DIFF
--- a/crates/protobuf/RUSTSEC-0000-0000.md
+++ b/crates/protobuf/RUSTSEC-0000-0000.md
@@ -15,7 +15,7 @@ patched = []
 functions = { "protobuf::coded_input_stream::CodedInputStream::skip_group" = ["<= 3.4.0"] }
 ```
 
-# RustSec Advisory Template - Advisory Title Goes Here
+# Crash due to uncontrolled recursion in protobuf crate
 
 Affected version of this crate did not properly parse unknown fields when parsing a user-supplied input.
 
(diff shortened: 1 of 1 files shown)


jayvdb:
ping @stepancheg

crates/protobuf/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,22 @@
...
+informational = "unmaintained"
+
+[versions]
+patched = []
+
+[affected]
+functions = { "protobuf::coded_input_stream::CodedInputStream::skip_group" = ["<= 3.4.0"] }
+```
+
+# RustSec Advisory Template - Advisory Title Goes Here
```
woodruffw, line 18:
```suggestion
# Crash due to uncontrolled recursion in protobuf crate
```

CC @DarkaMaul

DarkaMaul:
> > The crate is affected by the same vulnerability as described in [GHSA-735f-pc8j-v9w8](https://github.com/protocolbuffers/protobuf/security/advisories/GHSA-735f-pc8j-v9w8)
> 
> Given that this affects multiple implementations across different languages, do we know whether other Rust protobuf implementations such as `prost` might also be affected?

From my understanding, `prost` is not affected by the same issue: 

- it does not support `unknown` fields ( https://github.com/tokio-rs/prost/issues/2 )
- it performs a recursion depth limit check when parsing unknown fields

tarcieri:
@woodruffw yeah, `related` sounds good

PR #2170, opened 2024-12-18
Add advisory for rage plugin name vulnerability (GHSA-4fg7-vxc8-qx5w)

str4d:


crates/age/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,94 @@
...
+id = "RUSTSEC-0000-0000"
+package = "age"
+date = "2024-12-18"
+url = "https://github.com/str4d/rage/security/advisories/GHSA-4fg7-vxc8-qx5w"
+categories = ["code-execution"]
+aliases = ["GHSA-4fg7-vxc8-qx5w"]
+related = ["GHSA-32gq-x56h-299c"]
+
+[affected]
+functions = {
```
tarcieri, line 12: This is failing lint due to invalid syntax. Perhaps try this instead:

```suggestion
[affected.functions]
```

str4d: That was it. I've updated the examples to use this syntax.

tarcieri (approved):
Hmm, I guess the changes to the templates are OK, though I would've probably preferred them in a separate PR for focused discussion

str4d:
I made the template changes in this PR because I followed the templates to create it, and they did not work 🤷‍♂️

PR #2172, opened 2024-12-19
Report unsoundness and patch in fyrox-core

shinmao:


kornelski (approved):

Shnatsel:
Is there a timeline for a release? 

Since this is not an urgent security issue, I'd prefer not to bother API consumers with non-actionable advisories. So I'd rather merge this once a fixed version is released to crates.io

shinmao:
The maintainer mention that they will include the patch in the version 0.35, which will be released in one month. 👍

shinmao:
Hi, the version 0.36 is just released! I think we can release the advisory now.

PR #2173, opened 2024-12-19
Report unsoundness in Solana/spl-token-swap

shinmao:


kornelski (approved):

kornelski:
There's no patched version, but the project looks unmaintained.

shinmao:
@kornelski yes. The crate doesn't announce to be unmaintained, but it isn't updated over 2 years. Do you think we should release unmaintained advisories?

kornelski:
Before making unmaintained advisories, ask crate maintainers to confirm it. The repository seems to be active.

shinmao:
> Before making unmaintained advisories, ask crate maintainers to confirm it. The repository seems to be active.

I think it is still active since there was a commit two months ago.

PR #2176, opened 2024-12-23
Report UB in glib::VariantStrIter::impl_get in glib < 0.20.0

decathorpe:
reported / fixed here: https://github.com/gtk-rs/gtk-rs-core/pull/1343

This PR was merged as https://github.com/gtk-rs/gtk-rs-core/commit/05dff0e , part of glib `>=0.20.0`, but it was not backported to earlier release branches (so it still affects any project that uses the latest version of the gtk3 bindings, which are [unmaintained](https://github.com/rustsec/advisory-db/pull/2164)).


Shnatsel:
Thank you!

PR #2178, opened 2024-12-23
Report unsoundness in cortex-m

shinmao:
The [issue](https://github.com/rust-embedded/cortex-m/issues/485) has been patched in Nov 2023 but it is still not released in the latest version (0.7.7). We can ask whether the maintainer plan to release it first.

### DIFF
--- a/crates/cortex-m/RUSTSEC-0000-0000.md
+++ b/crates/cortex-m/RUSTSEC-0000-0000.md
@@ -1,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "cortex-m"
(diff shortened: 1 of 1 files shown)


Shnatsel:
Normally users would already get the latest version of the dependency crate. Publishing this advisory as-is would result in false positives even if the latest version did patch it, but right now the advisory seems entirely non-actionable.

Since the issue only manifests given a combination of versions between cortex-m and volatile_register, which we cannot encode in the database format right now, and that it is only unsoundness and not a pressing security issue, I'm inclined not to publish an advisory for it.

shinmao:
Yes. It makes sense. Thanks for pointing it out.

PR #2181, opened 2024-12-28
Report insecure cryptography in magic-crypt

frozolotl:


crates/magic-crypt/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,37 @@
...
+considered secure and it uses them in ways that do not guarantee the integrity
+of the encrypted data.
+
+`MagicCrypt64` uses the insecure DES block cipher in CBC mode without
+authentication. This allows for practical brute force and padding oracle
+attacks and does not protect the integrity of the encrypted data. Key and IV
+are generated from user input using CRC64, which is not at all a key derivation
+function.
+
+`MagicCrypt128` uses MD5 for its key and IV generation, which is also not not
```
Shnatsel, line 26: MD5 is only broken insofar collision resistance is concerned. I am not convinced that this is a problem for key generation. Its use for IV is trickier to assess. But everything else is very clearly a genuine cryptographic issue.

Would you be OK with dropping the mention of MD5 from the advisory? I'd prefer the issues in the advisory to be clearly and unambiguously valid. I'd be happy to merge without this line.

Shnatsel (commented):
Thank you for the report!

Just one nit, otherwise I'd be happy to merge

crates/magic-crypt/RUSTSEC-0000-0000.md
frozolotl: I'm fine with that, yes.

Shnatsel: Nice. Merged. Thank you!

PR #2188, opened 2025-01-09
Add deprecated actix-web-actors crate

sanpii:


kornelski (approved):

robjtede:
`actix-web-actors` is deprecated in order to encourage use of the newer crate `actix-ws`. However, it is not unmaintained, at least for now.

Its source code will likely be dropped for the next semver-major Actix Web release. As it stands though, I'd consider it slightly too early to add this advisory.

PR #2190, opened 2025-01-15
Report incorrect group information in `users`

danth:
This is already documented in https://github.com/ogham/rust-users/issues/44 regarding listing the groups for the current process. However, this also affects listing the supplementary groups for a particular user.

I discovered this while implementing a file server which controls access based on the underlying filesystem permissions. In this case, if the file in question was owned by the `root` group, then it would inadvertently be accessible to everyone, as they are believed to be a member of that group.

There may be other applications out there doing something similar, hence I think it's important to raise this as a security issue and not just a general bug.

The boundary for the affected versions is based on [this commit](https://github.com/ogham/rust-users/tree/954de4b83c744da0fd20073238e2604c8efcd940), where the functionality was first introduced.


crates/users/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
...
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "users"
+date = "2025-01-15"
+url = "https://github.com/ogham/rust-users/issues/44"
+categories = ["privilege-escalation"]
+
+[versions]
+patched = []
+unaffected = ["<= 0.7.0"]
```
riking, line 11: I assume this should be either `< 0.8.0` or `< 0.7.0`, use one of those instead.

crates/users/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
...
+Affected versions append `root` to group listings, unless the correct listing has exactly 1024 groups.
+
+This affects both:
+
+- The supplementary groups of a user
+- The group access list of the current process
+
+If the caller uses this information for access control, this may lead to privilege escalation.
+
+This crate is not currently maintained, so a patched version is not available.
```
riking, line 25: If using an older version is a defense, report that as a workaround

> Versions older than 0.7.0 are not affected, so downgrading to them is a workaround.

crates/users/RUSTSEC-0000-0000.md
danth: Yes, this should be `< 0.8.0`. There were no intermediate versions, so this condition is equivalent.

crates/users/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,27 @@
...
+
+Versions older than 0.8.0 do not contain the affected functions, so downgrading to them is a workaround.
```
djc: I think it's probably helpful to also at least mention the `uzers` crate here, which appears to be an active fork that has purportedly fixed this issue.

danth: Copied the "Recommended alternatives" section from `RUSTSEC-2023-0059`

crates/users/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,34 @@
...
+
+If the caller uses this information for access control, this may lead to
+privilege escalation.
+
+This crate is not currently maintained, so a patched version is not available.
+
+Versions older than 0.8.0 do not contain the affected functions, so downgrading
+to them is a workaround.
+
+## Recommended alternatives
```
djc, line 32: Sorry, I think we have to be a bit careful in our wording here. Suggest to replace "Recommended" with "Potential".

(Also, would be nice to add an empty line after the section title.)

danth: This section was copied word-for-word from the previous report on `users`

djc:
@danth thanks for your work on this!

PR #2191, opened 2025-01-15
Add advisory for segmentation fault in grcov

nyw0102:


nyw0102:
@Shnatsel Thank you for your former advice to split the advisory for "grcov" and "fast-float"! If you do not mind, could you take this issue into account?

Shnatsel:
Thanks for the report, and sorry for the delay!

I investigated the linked bug report and the patch that fixed it, and wrote a more clear description for the issue. I think this is good to go now.

nyw0102:
Thank you for your additional description and merging!

PR #2193, opened 2025-01-18
Advisory for CVE-2025-22620 (chmod 777) in gix-worktree-state

EliahKagan:
This adds a notice for CVE-2025-22620 (GHSA-fqmf-w4xh-33rh) in `gix-worktree-state`.

When a global GHSA is published, I can open another PR to add a reference to that.

I wasn't sure what to do here for affected operating systems. Only Unix-like operating systems are affected, which seems like something that could be valuable to express in the advisory metadata. But if I understand correctly, no value like `"unix"` is recognized for `affected.os`. Currently, I have not specified `affected.os` in the metadata. But would it be better to list all target operating systems that are Unix-like individually? (If so, I could use the technique in https://github.com/rustsec/advisory-db/pull/1911#discussion_r1512022858, which still works, to produce the list.)

cc @Byron

### DIFF
--- a/crates/gix-worktree-state/RUSTSEC-0000-0000.md
+++ b/crates/gix-worktree-state/RUSTSEC-0000-0000.md
@@ -1,0 +1,120 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "gix-worktree-state"
+date = "2025-01-18"
+url = "https://github.com/GitoxideLabs/gitoxide/security/advisories/GHSA-fqmf-w4xh-33rh"
+cvss = "CVSS:3.1/AV:L/AC:L/PR:L/UI:R/S:U/C:N/I:H/A:N"
+keywords = ["permissions"]
+aliases = ["CVE-2025-22620"]
+license = "CC0-1.0"
+
+[affected.functions]
+"gix_worktree_state::checkout" = ["*"]
+
+[versions]
+patched = [">= 0.17.0"]
+```
+
+# gix-worktree-state nonexclusive checkout sets executable files world-writable
+
+### Summary
+
+`gix-worktree-state` specifies 0777 permissions when checking out executable files, intending that the umask will restrict them appropriately. But one of the strategies it uses to set permissions is not subject to the umask. This causes files in a repository to be world-writable in some situations.
+
+### Details
+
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you!

Yeah, unfortunately there is no way to express unix-ness still. Maybe we should add one with `os_family` or some such.

crates/gix-worktree-state/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,120 @@
...
+package = "gix-worktree-state"
+date = "2025-01-18"
+url = "https://github.com/GitoxideLabs/gitoxide/security/advisories/GHSA-fqmf-w4xh-33rh"
+cvss = "CVSS:3.1/AV:L/AC:L/PR:L/UI:R/S:U/C:N/I:H/A:N"
+keywords = ["permissions"]
+aliases = ["CVE-2025-22620"]
+license = "CC0-1.0"
+
+[affected.functions]
+"gix_worktree_state::checkout" = ["*"]
```
Shnatsel, line 13: This should be `patched = ["< 0.17.0"]`, which is admittedly a bit of a wart in the format. I'll fix this field.

EliahKagan:
Thanks!

Should I list the Unix-like targets individually? Or is it better to leave it out?

crates/gix-worktree-state/RUSTSEC-0000-0000.md
EliahKagan: Thanks! I think I've done it this way (with `"*"`) in some other RUSTSEC advisories as well--I vaguely recall there had been some kind of linting error when I had tried to do it differently, but I was probably doing it the wrong way (maybe the relational operator I was using was something other than `<`).

My guess is that `"*"` that I have written in other advisories may already have been fixed. But I'll check right now and, if any remain (and there is a fix, such that a specific upper bound is applicable), I'll open a PR to improve them.

Shnatsel:
I think it's a good idea to list the unix-like targets. Since this PR is already merged, please open a new PR for that change.

The advisory will not go live until we assign it an ID by merging https://github.com/rustsec/advisory-db/pull/2194

EliahKagan:
Sounds good, I'll open a PR to list the OS targets very soon.

crates/gix-worktree-state/RUSTSEC-0000-0000.md
EliahKagan: I've opened #2198 for this. It seem to lint okay.

PR #2196, opened 2025-01-18
List Unix-like target OSes as affected by CVE-2025-22620

EliahKagan:
This builds on #2193 by listing all Unix-like target operating systems as affected (since a category of OSes like "unix" cannot currently be represented in RUSTSEC advisory metadata).

The list was obtained by running the command given in https://github.com/rustsec/advisory-db/pull/1911#discussion_r1512022858.

The vulnerability is specific to Unix-like operating systems because:

- The vulnerable code runs only in the `unix` build configuration.

- 0777 permissions are meaningful on such systems and (due to containing 0002) allow any user account on the system to write.

Therefore, if there are any Unix-like systems where Unix-style filesystem permissions are not used, or that are *truly* single user (i.e. do not use multiple user accounts, not even for running daemons with limited privileges), then this vulnerability would not affect such systems.

In addition, I have not attempted specifically to run the proof of concept for the vulnerability on most of the listed operating systems, nor examined whether `gix-worktree-state` might not be usable on some of them for reasons unrealted to this vulnerability.

Conversely, if new target OSes are added in the future, and they are Unix-like, then they would probably be vulnerable, even though not listed here. (It may not be likely that anyone would be using an affected version of `gix-worktree-state` by that time, though.)

### DIFF
--- a/crates/gix-worktree-state/RUSTSEC-0000-0000.md
+++ b/crates/gix-worktree-state/RUSTSEC-0000-0000.md
@@ -9,6 +9,38 @@ keywords = ["permissions"]
 aliases = ["CVE-2025-22620"]
 license = "CC0-1.0"
 
+[affected]
+# This vulnerability affects Unix-like operating systems.
+os = [
+    "aix",
+    "android",
+    "dragonfly",
+    "emscripten",
+    "espidf",
+    "freebsd",
+    "fuchsia",
+    "haiku",
+    "horizon",
+    "hurd",
+    "illumos",
+    "ios",
+    "l4re",
+    "linux",
+    "macos",
+    "netbsd",
+    "nto",
+    "nuttx",
+    "openbsd",
+    "redox",
+    "rtems",
+    "solaris",
+    "tvos",
+    "visionos",
+    "vita",
+    "vxworks",
+    "watchos",
+]
+
 [affected.functions]
 "gix_worktree_state::checkout" = ["< 0.17.0"]
 


crates/gix-worktree-state/RUSTSEC-0000-0000.md
```diff
@@ -9,6 +9,38 @@ keywords = ["permissions"]
 aliases = ["CVE-2025-22620"]
 license = "CC0-1.0"
 
+[affected]
+# This vulnerability affects Unix-like operating systems.
```
EliahKagan, line 13: If `#` comments are considered okay to have in advisory TOML, then I think this is helpful. But if not then this comment could be removed.

EliahKagan (commented):
This PR is as discussed in https://github.com/rustsec/advisory-db/pull/2193#issuecomment-2600025235. Aside from the caveats mentioned above in the description, there's one thing I wasn't totally sure about, commented below.

EliahKagan:
Thanks!

asomers:
FYI, this change is causing `cargo-audit` to crash.  See https://github.com/rustsec/rustsec/issues/1318 .

Shnatsel:
Right, because older versions don't recognize this platform. This should probably be made a non-fatal error, but in the meantime we'll just remove the offending obscure platforms.

asomers:
And publish a new DB without nuttx?  Sounds good to me.

EliahKagan:
Removing NuttX seems especially okay at least for now, since I'm not clear on whether the vulnerability is exploitable there. (I'm not knowledgeable about what filesystems are used there, and also whether any of the conditional mitigating factors listed in the advisory that usually don't apply might more often, or even always, apply on it.) Another RTOS that's not totally Unix-like and that I wasn't sure about is VxWorks. I can see that's been removed as well, along with various others, some of which I do suspect are affected, such as Hurd. Maybe the list can be revised again, including re-adding some of the operating systems removed in #2201, once tooling permits.

PR #2015, opened 2024-07-18
Advisory for CVE-2024-40644 (program files) in gix-path

EliahKagan:
This adds a notice for the [uncontrolled search path element](https://cwe.mitre.org/data/definitions/427.html) vulnerability CVE-2024-40644 (https://github.com/Byron/gitoxide/security/advisories/GHSA-mgvv-9p9g-3jv4) in `gix-path`.

cc @Byron

PR #2198, opened 2025-01-18
Specify affected functions versions instead of `"*"`

EliahKagan:
When including information about affected functions in advisory metadata, I have used `"*"` in a few advisories when the bounds were no narrower than the bounds for the vulnerability. But specific are still needed or preferred in that situation, as discussed in https://github.com/rustsec/advisory-db/pull/2193#pullrequestreview-2560582694.

This replaces `"*"` in those advisories with more specific bounds.

(For #2193 itself, this has already been done in #2195.)

### DIFF
--- a/crates/gix-path/RUSTSEC-2024-0355.md
+++ b/crates/gix-path/RUSTSEC-2024-0355.md
@@ -15,10 +15,10 @@ license = "CC0-1.0"
 os = ["windows"]
 
 [affected.functions]
-"gix_path::env::exe_invocation" = ["*"]
-"gix_path::env::installation_config" = ["*"]
-"gix_path::env::installation_config_prefix" = ["*"]
-"gix_path::env::system_prefix" = ["*"]
+"gix_path::env::exe_invocation" = [">= 0.10.8, < 0.10.9"]
+"gix_path::env::installation_config" = [">= 0.10.8, < 0.10.9"]
+"gix_path::env::installation_config_prefix" = [">= 0.10.8, < 0.10.9"]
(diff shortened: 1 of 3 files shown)


Shnatsel:
Apparently that works fine with our current tooling, so I guess it's just a matter of format specification. 

We used to stick to explicitly specifying the inverse of affected versions, but now that I think of it, it's probably not a good way to do it and we should just change it to something simpler. Maybe allowing `= "*"` is the way to do it, and explicitly specify that the versions in function specifications can only narrow down the original ranges.

EliahKagan:
Is `"*"` effectively allowed currently, such that it will not cause any problems for tools? (I have no objection to this PR being closed without merging if it is not an improvement.)

Shnatsel:
I'm not aware of any tools using this metadata currently. There has been some experimentation in osv-scanner, but with Rust not even having a stable, machine-readable function specification syntax, I don't think they ever deployed it.

EliahKagan:
In that case, I think I don't really have an opinion on whether this PR should be merged.

By the way, I did figure out what had led me to use `*` in the first place. I had been attempting to use an `==` operator in #2015 for a vulnerability that affected exactly one version of a crate (i.e. it was discovered shortly after the change that introduced it), and that operator wasn't supported (https://github.com/rustsec/advisory-db/pull/2015/commits/45186ec9dfb82db6865286fcc36709c8ff674dca).

EliahKagan:
I'll go ahead and close this, since based on the discussion, it seems like it would not necessarily be an improvement. If it turns out to be useful for me to reopen it or make a new PR along these lines, please let me know.

PR #2200, opened 2025-01-18
Fix bare URL hyperlinks in RUSTSEC-2025-0001

EliahKagan:
In #2193, the body text for [RUSTSEC-2025-0001](https://rustsec.org/advisories/RUSTSEC-2025-0001.html) (CVE-2025-22620) is from GHSA-fqmf-w4xh-33rh, and I forgot to make the two bare URLs linkified in the different Markdown dialect used on rustsec.org. (They link to directly relevant portions of code. Other hyperlinks are on text other than the URL itself, and thus already work.)

This adds `<` `>` around each of them, which I think is sufficient to linkify them as intended in a RUSTSEC advisory.


EliahKagan:
Thanks!

PR #2201, opened 2025-01-18
Remove obscure unix-like platforms from advisory due to tooling limit…

Shnatsel:
…ations

causes crashes on older cargo-audit versions

fixes #1318


asomers:
Thanks for the quick fix!  Note that the correct issue link is actually https://github.com/rustsec/rustsec/issues/1318 .

marvinthepa:
This still fails on cargo-audit 0.17.4:

```code
error: error loading advisory database: parse error: error parsing /data/git/msander/.cargo/advisory-db/crates/gix-worktree-state/RUSTSEC-2025-0001.md: parse error: Unrecognized value 'aix'
```

is that version unsupported? Or could you also remove aix?

tarcieri:
@marvinthepa it's a very old version of `cargo-audit`, can you update?

PR #2209, opened 2025-01-22
Add advisory for segfault in openssl-probe due to environment setters

mmastrac:
We tracked down a crash to an interaction with environment setters in openssl-probe and environment access in another thread. CVE scoring was cloned from a previous environment soundness issue (https://nvd.nist.gov/vuln/detail/CVE-2020-26235).

Filed issue with more details: https://github.com/alexcrichton/openssl-probe/issues/30

This is not really a sensitive issue, so it could probably be delayed a bit to give the crate a chance to update.

## Open Questions

 - [ ] Release 0.2.0 of openssl-probe and mark `< 0.2.0` as vulnerable?

## Major crate status

 - [x] native-tls (117M dl) - completed in https://github.com/sfackler/rust-native-tls/commit/a35127a5cc6d0519c4d6b4dce1fb14ab945ad347, released
 - [x] rustls-native-certs (105M dl)- not vulnerable, version bumped
 - [x] minreq (2M) - completed in https://github.com/neonmoe/minreq/commit/4bc16dba61ae19e3f81de33b80c8a3c0c8a33a0d

### Unfixed or partially fixed

 - [ ] git2-rs (41M) - unfixed
 - [ ] curl-rust (24M) - unfixed
 - [ ] mongodb (5M) - unfixed, but migrated to new, unsafe APIs (still at risk of crash) -- https://github.com/mongodb/mongo-rust-driver/pull/1294#issuecomment-2625713530
 - [ ] pingora-core - unfixed

### DIFF
--- a/crates/openssl-probe/RUSTSEC-0000-0000.md
+++ b/crates/openssl-probe/RUSTSEC-0000-0000.md
@@ -1,0 +1,16 @@
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "openssl-probe"
+date = "2025-01-10"
+url = "https://github.com/alexcrichton/openssl-probe/issues/30"
+references = ["https://www.edgedb.com/blog/c-stdlib-isn-t-threadsafe-and-even-safe-rust-didn-t-save-us"]
+informational = "unsound"
+categories = ["memory-corruption"]
+cvss = "CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:N/I:N/A:H"
+keywords = ["ssl", "openssl", "environment"]
+
+[affected]
+os = ["linux"]
+
+[versions]
+patched = []
--- a/crates/openssl-probe/RUSTSEC-0000-0000.md
+++ b/crates/openssl-probe/RUSTSEC-0000-0000.md
@@ -25,8 +25,15 @@ patched = [">= 0.1.6"]
 
 `openssl-probe` offers non-`unsafe` methods that call `std::env::set_var`, which may be called
 in a multithreaded environment, and potentially clash with environment access on other threads.
+In pure Rust code, concurrent read and write access to the environment is actually safe due to a lock
+taken in the platform implementations of the environment accessors (the documentation does not
+state this, and it's possible it _could_ change in the future). Libraries using other runtimes
+(including Python, those written in pure C and others) do not make use of these internal Rust
+environment locks, however, and instead use their own locks, or unprotected raw access to `libc`'s 
+`getenv`, `setenv`, or even worse, `char** environ`.
 
-When these methods are called while other threads are active and accessing the environment, it
+When these methods in `openssl-probe` (or that matter, any other pure Rust code calling `std::env::set_env`)
+are called while other threads are active and accessing the environment, it
 may cause other threads to access dangling environment pointers in the cases where the underlying
 environment data is moved or resized in response to an additional environment variable being
 added, or a variable's contents being enlarged.
--- a/crates/openssl-probe/RUSTSEC-0000-0000.md
+++ b/crates/openssl-probe/RUSTSEC-0000-0000.md
@@ -48,7 +48,7 @@ environment occur.
 ## Rust's `set_env`
 
 This crate, and all other callers of the Rust `set_env` function (<https://doc.rust-lang.org/std/env/fn.set_var.html>)
-are unsound due to the unfortunate reality of the POSIX standard which defines these enviornment access methods
+are unsound due to the unfortunate reality of the POSIX standard which defines these environment access methods
 without making any sort of thread-safety guarantees.
 
 In Rust's 2024 edition `std::env::set_var` is marked as `unsafe` and the documentation was updated to note
--- a/crates/openssl-probe/RUSTSEC-0000-0000.md
+++ b/crates/openssl-probe/RUSTSEC-0000-0000.md
@@ -16,3 +16,16 @@ os = ["linux"]
 [versions]
 patched = []
 ```
+
+# `openssl-probe` may cause memory corruption in multi-threaded processes
+
+`openssl-probe` offers non-`unsafe` methods that call environment setters, which may be called
+in a multithreaded environment, and potentially clash with environment access on other threads.
+
+When these methods are called while other threads are active and accessing the environment, it
+may cause the other threads to access dangling pointer values in the cases where the underlying
+environment data is moved or resized in response to an additional environment variable being
+added, or a variable's contents being enlarged.
+
+The affected function is `try_init_ssl_cert_env_vars` in 
+<https://github.com/alexcrichton/openssl-probe/blob/master/src/lib.rs#L65>.
--- a/crates/openssl-probe/RUSTSEC-0000-0000.md
+++ b/crates/openssl-probe/RUSTSEC-0000-0000.md
@@ -33,7 +33,21 @@ added, or a variable's contents being enlarged.
 This is shown to occur on Linux, but it will also likely occur on any other platform where `getenv`
 and `setenv` are not thread-safe, though trigger conditions may vary widely.
 
+## Affected Code
+
 The affected function is `try_init_ssl_cert_env_vars` in 
-<https://github.com/alexcrichton/openssl-probe/blob/db67c9e5b333b1b4164467b17f5d99207fad004c/src/lib.rs#L65>.
+<https://github.com/alexcrichton/openssl-probe/blob/db67c9e5b333b1b4164467b17f5d99207fad004c/src/lib.rs#L65>, and
+any other library's function which may call this function directly or indirectly
+<https://github.com/search?q=try_init_ssl_cert_env_vars&type=code>.  `native_tls <= 0.2.12` may
+do so in certain configurations <https://github.com/sfackler/rust-native-tls/blob/2424bc5efd1b8b4bcf60dbda93259a3f29db7f06/Cargo.toml>.
 
 The crate's author released a fix in versions `>=0.1.6` which marks these functions as `unsafe` and `#[deprecated]`.
+
+## Alternative Mitigations
+
+In the case of glibc users, some thread-safety improvements may protect you from `setenv`/`getenv` clashes
+which were introduced in <https://github.com/bminor/glibc/commit/7a61e7f557a97ab597d6fca5e2d1f13f65685c61>,
+however direct `environ` access in multithreaded programs will still risk dangling pointer access.
+
+Users of other `libc` implementations should consult their sourcecode listings for thread-safety guarantees
+around multithreaded environment read/write access, though readers should be prepared to be disappointed.
--- a/crates/openssl-probe/RUSTSEC-0000-0000.md
+++ b/crates/openssl-probe/RUSTSEC-0000-0000.md
@@ -10,11 +10,14 @@ categories = ["memory-corruption"]
 cvss = "CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:N/I:N/A:H"
 keywords = ["ssl", "openssl", "environment"]
 
+[affected.functions]
+"openssl_probe::try_init_ssl_cert_env_vars" = ["< 0.1.6"]
+
 [affected]
 os = ["linux"]
 
 [versions]
-patched = []
+patched = [">= 0.1.6"]
 ```
 
 # `openssl-probe` may cause memory corruption in multi-threaded processes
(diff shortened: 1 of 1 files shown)


crates/openssl-probe/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,39 @@
...
+date = "2025-01-10"
+url = "https://github.com/alexcrichton/openssl-probe/issues/30"
+references = ["https://www.edgedb.com/blog/c-stdlib-isn-t-threadsafe-and-even-safe-rust-didn-t-save-us"]
+informational = "unsound"
+categories = ["memory-corruption"]
+cvss = "CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:N/I:N/A:H"
+keywords = ["ssl", "openssl", "environment"]
+
+[affected.functions]
+"openssl_probe::try_init_ssl_cert_env_vars" = ["< 0.1.6"]
```
alexcrichton, line 14: It might make sense to also include the `init_ssl_cert_env_vars` function here too?

crates/openssl-probe/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,39 @@
...
+environment data is moved or resized in response to an additional environment variable being
+added, or a variable's contents being enlarged.
+
+This is shown to occur on Linux, but it will also likely occur on any other platform where `getenv`
+and `setenv` are not thread-safe, though trigger conditions may vary widely.
+
+The affected function is `try_init_ssl_cert_env_vars` in 
+<https://github.com/alexcrichton/openssl-probe/blob/db67c9e5b333b1b4164467b17f5d99207fad004c/src/lib.rs#L65>.
+
+The crate's author released a fix in versions `>=0.1.6` which marks these functions as `unsafe` and `#[deprecated]`.
```
alexcrichton, line 39: Two possible points of clarification here:

* Perhaps worth linking to libstd's documentation on `set_var`? This advisory is basically "the crate calls `set_var` and it's bad for all the same reasons that `set_var` is being made `unsafe` now"
* This might be a bit pedantic but the previously-safe versions of the functions are`#[deprecated]`, but the now-`unsafe` versions are not `#[deprecated]` they're just `unsafe`-and-documented-as-unsafe.

alexcrichton (commented):
👍

crates/openssl-probe/RUSTSEC-0000-0000.md
mmastrac: I agree -- the text was not entirely clear. I've tried to write this in a blame-free retrospective style; please let me know if it reads OK with that in mind.

crates/openssl-probe/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,68 @@
...
+os = ["linux"]
+
+[versions]
+patched = [">= 0.1.6"]
+```
+
+# `openssl-probe` may cause memory corruption in multi-threaded processes
+
+`openssl-probe` offers non-`unsafe` methods that call `std::env::set_var`, which may be called
+in a multithreaded environment, and potentially clash with environment access on other threads.
```
alexcrichton, line 27: To confirm my own understanding on this, the unsafety here is only possible with non-Rust accesses to env vars, right? I was under the impression that if env vars are exclusively accessed through Rust then `set_var` is indeed safe, but once non-Rust users are added in there's no way to have it be safe.

It might be worth clarifying that here but the whole purpose of this crate is to interact with openssl which is, by definition, a non-Rust component working with env vars so it's also unlikely that a clarification would actually help much.

crates/openssl-probe/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,68 @@
...
+url = "https://github.com/alexcrichton/openssl-probe/issues/30"
+references = ["https://www.edgedb.com/blog/c-stdlib-isn-t-threadsafe-and-even-safe-rust-didn-t-save-us"]
+informational = "unsound"
+categories = ["memory-corruption"]
+cvss = "CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:N/I:N/A:H"
+keywords = ["ssl", "openssl", "environment"]
+
+[affected.functions]
+"openssl_probe::try_init_ssl_cert_env_vars" = ["< 0.1.6"]
+"openssl_probe::init_ssl_cert_env_vars" = ["< 0.1.6"]
```
alexcrichton, line 15: On second thought for the `< 0.1.6` bit here (and the line above) - technically this isn't correct where `init_ssl_cert_env_vars` is still just as unsafe as it was before on 0.1.6+. I'm not sure how that interacts with tooling and this advisory though. Technically though just because someone updates to 0.1.6 doesn't really "fix" things, it's actually so long as these funtions are used then a "fix" is still necessary. In other words the true fix for this advisory isn't actually in `openssl-probe`, it's in all users of `openssl-probe` and fixing is more broadly scoped than just updating the `openssl-probe` crate

mmastrac: That's a good point -- I believe that a deprecation warning _probably_ means that the API consumer is aware and these functions are no longer "affected" but we may need someone else to weigh in.

crates/openssl-probe/RUSTSEC-0000-0000.md
mmastrac: I'll ponder and clarify this a bit -- it's true that we require a non-Rust component in the mix here. 

IIRC the per-platform Rust env implementations take a lock for all reads and writes on platforms other than Windows, but given that the guarantee of safe concurrency within pure Rust code is not documented in `set_env` it might be a small shortcoming in the docs and perhaps an upstream bug is needed there.

crates/openssl-probe/RUSTSEC-0000-0000.md
mmastrac: I took one more pass to clarify the role of the Rust platform locks around environment access.

alexcrichton (approved):

crates/openssl-probe/RUSTSEC-0000-0000.md
eric-seppanen: Imagine I depend on `reqwest`, which transitively depends on `openssl-probe 0.1.5`. I see the advisory, which appears to recommend updating to `0.1.6`. I run `cargo update -p openssl-probe` and am satisfied that the problem is solved.

I will never see this deprecation warning, because it only appears when compiling dependencies, and warnings in dependencies are normally suppressed.

I think don't think `< 0.1.6` should go in the advisory, because it will lead to the wrong outcome for most projects.

tarcieri: > Imagine I depend on `reqwest`, which transitively depends on `openssl-probe`

Sounds like this advisory has large potential to be very noisy, which is something to consider. We've had a lot of backlash in the past from similarly noisy advisories.

eric-seppanen: Perhaps some outreach to the [dependent crates](https://crates.io/crates/openssl-probe/reverse_dependencies) would be a good idea? If those crate maintainers receive some warning before this dependency hits everyone, they may be better prepared (e.g. with a fixed release that resolves the issue... somehow).

eric-seppanen: > I think don't think < 0.1.6 should go in the advisory, because it will lead to the wrong outcome for most projects.

On the other hand, without a "fixed" version, we get an advisory that always triggers, until the dependency is removed entirely. The only other way for tools like `cargo-deny` to prove that the danger is gone would be to update to a new major version of `openssl-probe` with the old function names removed. Which doesn't seem to be the plan?

alexcrichton: Right yeah I wasn't planning on doing a major release of `openssl-probe` as I suspected that would cause even more churn (e.g. warning about duplicate versions, etc).

mmastrac: @alexcrichton If you'd like, I can put together a 0.2.0 patch where we 1) remove the deprecated functions entirely and 2) put the new unsafe functions behind a feature flag with an appropriately noisy name (`__dangerous_single_threaded_environment_setters` or similar). Consumers of the API can make a judgment call on whether their use is safe, and advisories can always be issued down the line if needed.

With that release out, the `native-tls` project can switch from configuring OpenSSL via env vars to one of the other methods that allow providing paths directly to the OpenSSL context (the vars are really just a convenience). Looks like `git2` will also be affected in a similar way.

We can annotate the dangerous methods with appropriate guidance on how OpenSSL can be configured correctly.

mmastrac: After a quick scan, it appears that outside of Alex, only @sfackler (native-tls), @abr-egn (mongodb),  @neonmoe (minreq), and @pimeys (pingora-core) have dependent crates using the affected function.

crates/openssl-probe/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,75 @@
...
+and `setenv` are not thread-safe, though trigger conditions may vary widely.
+
+Note that these function calls are completely safe and sound in purely single-threaded environments,
+or multi-threaded environments where it can be proven that no simultaneous read and writes to the
+environment occur.
+
+## Rust's `set_env`
+
+This crate, and all other callers of the Rust `set_env` function (<https://doc.rust-lang.org/std/env/fn.set_var.html>)
+are unsound due to the unfortunate reality of the POSIX standard which defines these enviornment access methods
```
pvichivanives, line 51:
```suggestion
are unsound due to the unfortunate reality of the POSIX standard which defines these environment access methods
```

crates/openssl-probe/RUSTSEC-0000-0000.md
eric-seppanen: What about dependent crates that *don't* use the affected function? By omission, I assume this includes `native-tls` and `rustls-native-certs`.

These crates are likely to be somewhat impacted, because the automated tools that consume rustsec data will raise the advisory if you're using the affected crate-- which probably leads to complaints directed at that dependent crate, "why are you still using <version>, rustsec says it's unsound!"

If this advisory is going to be approved then they might appreciate some warning.

The path of least resistance is for there to be a released version that all dependent crates can update to that no longer triggers advisories (and is sound).

mmastrac: @eric-seppanen native-tls did use it, it's been fixed. rustls-native-certs wasn't affected but was bumped. All relevant crate owners above were brought into the thread.

mmastrac: (if we end up releasing a 0.2.0 with the affected functions behind a bigger wall, we'll have to do that dance again)

eric-seppanen: In this advisory we're faced with two bad choices:
1. Set the advisory to say "< 0.1.6"; downstream workspaces can bump `openssl-probe` to `0.1.6` in `Cargo.lock` and tooling (`cargo audit` or `cargo deny`) says the problem is gone. But nothing was fixed! The deprecation warning will be hidden if it's in a dependency, so these projects are still using dangerous code. This advisory didn't really have the desired effect.
2. Set the advisory to say "0.1.x"; which is an advisory that can never be cleared (only manually ignored) because that covers all released versions. This advisory had a stronger effect but is annoying because every downstream project that has automated rustsec checks has to take actions beyond updating a dependency. This will also annoy maintainers of dependent crates that are "fixed" (not calling the dangerous functions), because automated rustsec checks can't tell the difference.

Which leads us to a new option:

3. Set the advisory to say "0.1.x" and release 0.2.0 with the safe-but-maybe-unsound function names removed. For the purpose of the advisory, and automated advisory tooling, this is the best outcome: downstream projects can resolve the advisory by bumping dependencies until 0.1.x is gone from `Cargo.lock`. This assumes that the major dependent crates cooperate by replacing 0.1.6 with 0.2.0. If they don't then this is a pointless exercise.

Is it worth it, considering the ecosystem churn caused by 0.1.x and 0.2.0 coexisting for a period? That's harder to answer.

eric-seppanen: Apologies for being long-winded. I was trying to respond to

> native-tls did use it, it's been fixed

Great! Unfortunately, nothing in this advisory says "update to `native-tls 0.2.13`". This advisory tells projects to bump `openssl-probe` to 0.1.6, which for the vast majority of workspaces will not fix anything.

mmastrac:
As a heads-up, I think this particular advisory is going to be too much for me to manage with my limited bandwidth and the conflicting requirements. The process is certainly frustrating as an outsider, so I'm going to have to leave it in your capable hands.

I'm happy that the underlying problems have been fixed and the fixes have circulated, but I'm going to abandon this PR and walk away from the process. Consider the advisory-in-progress as licensed CC0 as needed for whatever has to happen going forward. Appreciate the quick work from everyone involved.

crates/openssl-probe/RUSTSEC-0000-0000.md
tarcieri: I think the advisory definitely needs to be actionable. We published a similar advisory for `chrono` ([RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159.html)) which wasn't actionable and got a considerable amount of blowback, to the point that we ended up withdrawing it.

It would be good to coordinate with major notable downstream dependencies (there are at least 7 with more than a million downloads) to ensure that when we publish the advisory, users of those crates simply need to `cargo update`

PR #2215, opened 2025-01-31
Create advisory for unmaintained `paste` crate

### LINKED ISSUE #2203: paste is unmaintained
https://crates.io/crates/paste is 350k/day downloads is unmaintained as of  Oct 7, 2024 when https://github.com/dtolnay/paste was archived after a note was added https://github.com/dtolnay/paste/commit/6a302522990cbfd9de4e0c61d91854622f7b2999

> _(This project is no longer maintained.)_

AS1100K:
Fixes #2203

### DIFF
--- a/crates/paste/RUSTSEC-0000-0000.md
+++ b/crates/paste/RUSTSEC-0000-0000.md
@@ -1,0 +1,16 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "paste"
+date = "2024-10-07"
+url = "https://github.com/dtolnay/paste"
+informational = "unmaintained"
+
+[versions]
+patched = []
+```
+
+# paste - no longer maintained
+
+The creator of the crate `paste` has stated in the [`README.md`](https://github.com/dtolnay/paste/blob/master/README.md) 
+that this project is not longer maintained as well as archived the repository


crates/paste/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,22 @@
...
+# paste - no longer maintained
+
+The creator of the crate `paste` has stated in the [`README.md`](https://github.com/dtolnay/paste/blob/master/README.md) 
+that this project is not longer maintained as well as archived the repository
+
+Some possible alternatives
+
+- https://crates.io/crates/unique-paste
+- https://crates.io/crates/mini_paste
+- https://crates.io/crates/affix
```
dtolnay, line 22: It is misguided and misleading to include these. None of these crates has been touched in more than a year. One of them was already explicitly abandoned more than 3 years ago. The other two have Issues disabled. None of them run CI. There is no reason to expect that they would be more maintained than the advisory crate.

dtolnay (changes requested):

crates/paste/RUSTSEC-0000-0000.md
AS1100K: Thanks for the review, I copied it directly from #2203 without verifying that those crates were good enough or not. I'm sorry for the mistake. I am also, thinking of creating a new crate i.e. fork of `paste` named [pastey](https://github.com/as1100k/pastey) but I haven't decided the direction I want to go with the direction.

decathorpe:
FWIW I maintain the package for this crate in Fedora Linux, and 150+ of our packages depend on it. As such, I'll likely need to keep it around basically forever anyway, and would also contribute some maintenance if needed. I assume the debian packagers will be in a similar boat. Not sure if that would be enough to revoke this "unmaintained" advisory. (I would have filed an issue on the "paste" repo, but it is already archived so issues are read-only.)

AS1100K:
I have created a crate named [pastey](https://github.com/as1100k/pastey). This crate is the fork of paste and is a direct drop in replacement for paste crate along with some other features and bug fixes that paste was lacking.
Although, this crate is new and I am fully into maintaining it, can we edit this advisory to recommend this crate as an alternative?

tarcieri:
It would be nice to get @dtolnay's blessing on any forks we explicitly name in the advisory

dtolnay:
I don't plan to endorse a particular replacement, but it doesn't make a difference to me. For my own use cases I would use `paste` since it is 'finished' and does everything I need. Perhaps it would be better for it to be vetted by someone whose use case is not already met by `paste`.

decathorpe:
If the project is essentially "done", then I would feel competent to do enough "maintenance" for this advisory to be withdrawn. I am already responsible for the [Fedora Linux](https://src.fedoraproject.org/rpms/rust-paste) package for this crate, so I need to keep it around either way.

To make it explicit, I offer to do this if the "dtolnay/paste" repo is unarchived & transferred to me (or to the commons-rs org, which I have set up for situations like this a while ago).

The crate is also packaged for [Debian](https://packages.debian.org/trixie/librust-paste-dev) and I would reach out to its maintainers if they want to contribute as well.

PR #2219, opened 2025-02-08
add hickory-dns and hickory-proto advisories for incorrect DNSKEY validation

bluejekyll:


alex:
Sigh, I think this is a bug in the validator, those version ranges aren't actually overlapping, as I understand the semver rules?

Shnatsel:
The validator seems to be correct. It's following the [semver precedence rules](https://semver.org/#spec-item-11), under which 0.25.0-alpha.5 comes *before* 0.25.0, so the ranges overlap.

Shnatsel:
If you explain where the issue was introduced and fixed in plain English, I can translate it into the rustsec version ranges for you.

bluejekyll:
Ahhh. Thank you for your explanation. I think I can fix that.

crates/hickory-proto/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,37 @@
...
+url = "https://github.com/hickory-dns/hickory-dns/security/advisories/GHSA-37wc-h8xc-5hc4"
+# Valid categories: "code-execution", "crypto-failure", "denial-of-service", "file-disclosure"
+# "format-injection", "memory-corruption", "memory-exposure", "privilege-escalation"
+categories = ["crypto-failure"]
+keywords = ["dnssec"]
+#aliases = ["CVE-YYYY-NNNN"]
+#cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H"
+
+[versions]
+patched = [">= 0.24.3, < 0.25.0-alpha.5", ">= 0.25.0-alpha.5"]
```
Shnatsel, line 15: This version specification is equivalent to just ">= 0.24.3". Could you clarify what you were trying to express, in plain English? I'll be happy to help translate that into rustsec version ranges.

Shnatsel: If you're looking to express "everything in 0.24 series after 0.24.3, and everything greater than 0.25.0-alpha.5" that would be ["^ 0.24.3", ">= 0.25.0-alpha.5"]

bluejekyll: Thanks, I was unaware of the sigil. Yes, everything in the `0.24` release after `0.24.3` is patched. Everything in the `0.25` release after the `0.25.0-alpha.5` pre-release. I'll update when I have a moment.

bluejekyll: Thank you!

PR #2221, opened 2025-02-10
Report unsoundness in `cve-rs`, `totally-safe-transmute` and `totally-safe`

### LINKED ISSUE #826: totally-safe-transmute is totally unsafe

Nugine:
We are using `cargo audit -D warnings` to avoid problematic dependencies.

To prevent anyone from using these crates in the dependency tree (accidentally?), I think it is meaningful to include them in advisories.

+ https://github.com/meithecatte/fake-static ([RUSTSEC-2020-0013](https://rustsec.org/advisories/RUSTSEC-2020-0013.html))
+ https://github.com/Speykious/cve-rs
+ https://github.com/ben0x539/totally-safe-transmute
+ https://github.com/viktorlott/totally-safe

Reverse dependencies
+ https://crates.io/crates/cve-rs/reverse_dependencies (0 dependents)
+ https://crates.io/crates/totally-safe-transmute/reverse_dependencies (1 dependent)
+ https://crates.io/crates/totally-safe/reverse_dependencies (0 dependents)

resolves #826

### DIFF
--- a/crates/cve-rs/RUSTSEC-0000-0000.md
+++ b/crates/cve-rs/RUSTSEC-0000-0000.md
@@ -1,0 +1,19 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "cve-rs"
+date = "2025-02-10"
+url = "https://github.com/Speykious/cve-rs"
+categories = ["memory-corruption"]
+informational = "unsound"
+
+[versions]
+patched = []
+unaffected = []
+```
+
+# cve-rs introduces memory vulnerabilities in safe Rust
+
+`cve-rs` allows you to introduce common memory vulnerabilities (such as buffer overflows and segfaults) into your Rust program in a memory safe manner.
+
+Internally, this crate does not use unsafe code, it instead exploits a soundness bug in rustc: https://github.com/rust-lang/rust/issues/25860
--- a/crates/cve-rs/RUSTSEC-0000-0000.md
+++ b/crates/cve-rs/RUSTSEC-0000-0000.md
@@ -6,6 +6,7 @@ date = "2025-02-10"
 url = "https://github.com/Speykious/cve-rs"
 categories = ["memory-corruption"]
 informational = "unsound"
+keywords = ["soundness-hole"]
 
 [versions]
 patched = []
--- a/crates/totally-safe-transmute/RUSTSEC-0000-0000.md
+++ b/crates/totally-safe-transmute/RUSTSEC-0000-0000.md
@@ -1,0 +1,19 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "totally-safe-transmute"
+date = "2025-02-10"
+url = "https://github.com/ben0x539/totally-safe-transmute"
+categories = ["memory-corruption"]
+informational = "unsound"
+
+[versions]
+patched = []
+unaffected = []
+```
+
+# totally-safe-transmute allows transmuting any type to any other type in safe Rust
+
+This crate uses a known soundness issue (https://github.com/rust-lang/rust/issues/32670) that will never get fixed. In short, Linux provides a file called `/proc/self/mem` which can be used by a program to modify its own memory. This library modifies an enum variant number by accessing its own memory as a file to effectively transmute a variable.
+
+See also <https://doc.rust-lang.org/std/os/unix/io/index.html#procselfmem-and-similar-os-features>
--- a/crates/totally-safe-transmute/RUSTSEC-0000-0000.md
+++ b/crates/totally-safe-transmute/RUSTSEC-0000-0000.md
@@ -6,6 +6,7 @@ date = "2025-02-10"
 url = "https://github.com/ben0x539/totally-safe-transmute"
 categories = ["memory-corruption"]
 informational = "unsound"
+keywords = ["soundness-hole"]
 
 [versions]
 patched = []
--- a/crates/totally-safe/RUSTSEC-0000-0000.md
+++ b/crates/totally-safe/RUSTSEC-0000-0000.md
@@ -1,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "totally-safe"
+date = "2025-02-10"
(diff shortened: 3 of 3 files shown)


tarcieri:
See previous discussion on #826

Nugine:
The three crates are intentional exploits of rustc soundness bugs, like `fake-static`, which has been flagged in #207 and #270 previously. I personally recommend including them as `informational = "unsound"`.
If RustSec cannot make a decision yet, I'm Ok to leave this PR open and wait.

kornelski:
These crates may be a joke, but the APIs are unsound for real. I expect `cargo audit` is used in projects where ending up with any of them as a dependency isn't funny. I would prefer to have them flagged.

kornelski (approved):

riking:
I think some new common vocabulary here might be helpful - perhaps "[crate] is intentionally unsound" with `keyword=["intentionally-unsound"]` ?

djc:
> I think some new common vocabulary here might be helpful - perhaps "[crate] is intentionally unsound" with `keyword=["intentionally-unsound", "compiler-bug"]` ?

I like this suggestion.

fintelia:
I wonder if something along the lines of "exploit-demo" would be a better tag. The crates are all demonstrations of unsoundness caused by other software (either rustc or the standard library) which work by exposing a clearly unsound API while using no unsafe code themselves.

djc:
> I wonder if something along the lines of "exploit-demo" would be a better tag. The crates are all demonstrations of unsoundness caused by other software (either rustc or the standard library) which work by exposing a clearly unsound API while using no unsafe code themselves.

Maybe `unsound-demo`, then?

fintelia:
These crates do not contain a soundness hole. Rather, they're demonstrations that a soundness hole exists in the compiler/standard library.

djc (approved):

djc:
Thanks!

fintelia:
Having the category be "unsound" and the keyword "soundness-hole" in the advisory makes it sound to me like these are saying that the crates themselves contain soundness holes. Further, the advisory text lacks any discussion that these are demonstrations and not intended for use by other crates. In fact, restating the satirical content of those crate's documentation provides the exact opposite impression. It would be very easy to interpret these advisories as disparaging the crate authors

djc:
> Having the category be "unsound" and the keyword "soundness-hole" in the advisory makes it sound to me like these are saying that the crates themselves contain soundness holes. Further, the advisory text lacks any discussion that these are demonstrations and not intended for use by other crates. In fact, restating the satirical content of those crate's documentation provides the exact opposite impression. It would be very easy to interpret these advisories as disparaging the crate authors

A PR revising the text in a direction you find more appropriate would be welcome.

Nugine:
> Having the category be "unsound" and the keyword "soundness-hole" in the advisory makes it sound to me like these are saying that the crates themselves contain soundness holes. Further, the advisory text lacks any discussion that these are demonstrations and not intended for use by other crates. In fact, restating the satirical content of those crate's documentation provides the exact opposite impression. It would be very easy to interpret these advisories as disparaging the crate authors

If you can trigger UB without using any `unsafe` API from a library, then the library is *unsound*. These crates are *unsound* even though the holes are "re-exported" from rustc.

These demonstrations should not be published to crates.io if they are really not intended for use by other crates. 

I don't think coping the introduction of these crates makes it satirical to the crate authors.

djc:
@Speykious, @ben0x539, @viktorlott any comments/suggestions?

ben0x539:
Hi I'm the totally-safe-transmute author. I appreciate the throughtful discussion in this thread over my silly joke crate. I'm currently not using rust in production so I happily defer to others' expectations that cargo-audit warns them if someone genuinely pulls in totally-safe-transmute.

I would appreciate it if the wording of the advisory included some indication that the crate is doing a bit and as such wasn't meant to be used in real code, and that's why it will not have patched versions. I acknowledge I didn't exactly bother explicitly giving any such indication either, but then again my target audience was rust programmers with extra time on their hands and not consumers of advisories from an automated system who did not sign up for divining intent from context clues.

If you were to put something like "This crate is a toy and should never be used" I think I should be perfectly satisfied, and I think it would also be good information for advisory consumers, since, like, the problem that someone is pulling toy code into your production dependency tree is distinct from the problem that that code is sketchy. I suppose it doesn't actually come up a lot or there would be a tag for it already.

> These demonstrations should not be published to crates.io if they are really not intended for use by other crates.

Yeah I suppose I never had a good reason for posting to crates.io beyond roleplaying as a real crate. I apologize to anyone who's had extra headaches at work because of me having a fun time with my silly code.

djc:
> If you were to put something like "This crate is a toy and should never be used" I think I should be perfectly satisfied, and I think it would also be good information for advisory consumers, since, like, the problem that someone is pulling toy code into your production dependency tree is distinct from the problem that that code is sketchy. I suppose it doesn't actually come up a lot or there would be a tag for it already.

Want to submit a PR along these lines?

crates/cve-rs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,20 @@
...
+keywords = ["soundness-hole"]
+
+[versions]
+patched = []
+unaffected = []
+```
+
+# cve-rs introduces memory vulnerabilities in safe Rust
+
+`cve-rs` allows you to introduce common memory vulnerabilities (such as buffer overflows and segfaults) into your Rust program in a memory safe manner.
```
Speykious, line 18: I saw some discussion above about the utility of restating the project's satirical description. I don't have much of anything to suggest in regards to this PR, but if someone makes a PR in the future and wants to change the descriptions, I propose:
```
`cve-rs` provides demonstrations of common memory vulnerabilities (such as buffer overflows and segfaults) implemented completely within safe Rust.
```
I'm fine either way. That being said, I don't have much experience using Rustsec (I probably should).

djc: For you too: a PR making this change would be great!

PR #2230, opened 2025-02-22
*ring* is maintained again

djc:
Remove unmaintained advisory RUSTSEC-2025-0007.

@ctz and the rustls maintainers now have access to the crates.io entry for ring.

### DIFF
--- a/crates/ring/RUSTSEC-2025-0007.md
+++ b/crates/ring/RUSTSEC-2025-0007.md
@@ -5,6 +5,7 @@ package = "ring"
 date = "2025-02-20"
 informational = "unmaintained"
 url = "https://github.com/briansmith/ring/discussions/2414"
+withdrawn = "2025-02-22"
 
 [versions]
 patched = []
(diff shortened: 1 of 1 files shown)


alex:
I think the correct way to do this is to add `withdrawn = "2025-02-22"`?

djc:
> I think the correct way to do this is to add `withdrawn = "2025-02-22"`?

Ahh, I was looking for something like that.

djc:
> I think the correct way to do this is to add `withdrawn = "2025-02-22"`?

Fixed.

Shnatsel:
Thank you!

It would be nice to add a line or two about rustls maintainers getting access into the advisory text to make the situation more clear.

crates/ring/RUSTSEC-2025-0007.md
```diff
@@ -16,3 +17,9 @@ unaffected = []
 The author has announced an indefinite hiatus in its development, noting that
 any reported security vulnerabilities may go unaddressed for prolonged periods
 of time.
+
+# Update: security maintenance only
```
djc, line 21: @ctz thoughts on this wording?

alex (approved):

PR #2235, opened 2025-03-03
Add advisory for Stack-buffer-overflow in xmas-elf

nyw0102:


alex:
> However, it is still vulnerable because "bucket_count" which stands for the size of hashtable and offset value of the hashtable can be manipulated by users

This doesn't look true? https://github.com/nrc/xmas-elf/blob/master/src/hash.rs#L6-L10 none of these fields are `pub`.

Is there some other way that users can control the values of these fields?

crates/xmas-elf/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,19 @@
...
+package = "xmas-elf"
+date = "2025-03-03"
+url = "https://github.com/nrc/xmas-elf/issues/86"
+categories = ["memory-corruption"]
+
+[affected]
+functions = { "xmas_elf::hash::HashTable::get_bucket" = ["*"] }
+
+[versions]
+patched = []
```
riking, line 13: Needs a release

crates/xmas-elf/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,19 @@
...
+```
+# Stack-buffer-overflow in HashTable Struct
+In this case, the "get_bucket" method within the HashTable struct uses the unsafe keyword to access memory with assert function protection.
```
riking, line 16: Needs major copy editing. Suggestion:

> A maliciously crafted ELF file can present an invalid SectionData HashTable that causes xmas-elf to access memory out of bounds. Absurd values of the `bucket_count` and `chain_count` fields can cause a pointer offset of up to 16GB to be read and used. If the actual ELF file is smaller than this offset, memory will be accessed out of bounds.

riking (changes requested):

crates/xmas-elf/RUSTSEC-0000-0000.md
nyw0102: Thanks! Now I find that this bug is fixed in version '0.10'. Now I will add this information.

crates/xmas-elf/RUSTSEC-0000-0000.md
nyw0102: Thank you for your suggestion. I will add the details based on your suggestion.

PR #2241, opened 2025-03-06
*ring* versions prior to 0.17 are unmaintained.

briansmith:


tarcieri:
Perhaps instead of filing a new advisory, you can update RUSTSEC-2025-0007 and remove the withdrawn attribute? That would avoid having two duplicate advisories for the same thing.

djc:
> Perhaps instead of filing a new advisory, you can update RUSTSEC-2025-0007 and remove the withdrawn attribute? That would avoid having two duplicate advisories for the same thing.

Not sure that would make sense? The intent seems very different to me.

briansmith:
I guess I agree with djc. If somebody already chose to ignore the earlier advisory then they wouldn't see anything with the update, right?

PR #2254, opened 2025-03-17
Import old Wasmtime security advisories 

alexcrichton:
[Wasmtime] recently got a [request] to have our security advisories
published on the RustSec database as well. We've got a few old
advisories on here but we haven't been keeping up-to-date with later
advisories. In lieu of automatic imports from GitHub to RustSec we
figured we'd in the interim manually fill in some fields.

In this PR I'm back-filling security advisories we've had in Wasmtime
into the RustSec database here. The oldest advisory here is 3 years old
and the goal is to have this serve as a template for importing future
advisories that Wasmtime gets. It's not expected for this to cause any
churn or undue warnings but instead is intended to bring RustSec
up-to-date with the advisories we have for this crate.

[Wasmtime]: https://crates.io/crates/wasmtime
[request]: https://github.com/bytecodealliance/wasmtime/issues/10344

### DIFF
--- a/crates/wasmtime/RUSTSEC-0000-0000.md
+++ b/crates/wasmtime/RUSTSEC-0000-0000.md
@@ -1,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "wasmtime"
+date = "2022-02-17"
+url = "https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-88xq-w8cq-xfg7"
+categories = ["memory-corruption"]
+keywords = []
+aliases = ["CVE-2022-23636", "GHSA-88xq-w8cq-xfg7"]
+license = "CC0-1.0"
+
+[versions]
+patched = [">= 0.33.1, < 0.34.0", ">= 0.34.1"]
+```
+
+# Invalid drop of VMExternRef from partially-initialized instances in the pooling instance allocator
+
+This is an entry in the RustSec database for the Wasmtime security advisory
+located at
+https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-88xq-w8cq-xfg7.
+For more information see the GitHub-hosted security advisory.
--- a/crates/wasmtime/RUSTSEC-0000-0000.0.md
+++ b/crates/wasmtime/RUSTSEC-0000-0000.0.md
@@ -1,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "wasmtime"
+date = "2022-03-28"
+url = "https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-gwc9-348x-qwv2"
+categories = ["memory-corruption"]
+keywords = []
+aliases = ["CVE-2022-24791", "GHSA-gwc9-348x-qwv2"]
+license = "CC0-1.0"
+
+[versions]
+patched = [">= 0.35.2", ">= 0.34.2, < 0.35.0"]
+```
+
+#  Use after free with `externref`s and epoch interruption in Wasmtime
+
+This is an entry in the RustSec database for the Wasmtime security advisory
+located at
+https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-gwc9-348x-qwv2.
+For more information see the GitHub-hosted security advisory.
--- a/crates/wasmtime/RUSTSEC-0000-0000.1.md
+++ b/crates/wasmtime/RUSTSEC-0000-0000.1.md
@@ -1,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "wasmtime"
+date = "2022-06-27"
+url = "https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-jqwc-c49r-4w2x"
+categories = []
+keywords = []
+aliases = ["CVE-2022-31104", "GHSA-jqwc-c49r-4w2x"]
+license = "CC0-1.0"
+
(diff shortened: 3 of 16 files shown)


alexcrichton:
For maintainers here, two questions:

* Is there a way to import multiple advisories at once? Or should I make a separate PR-per-advisory for the initial backfill?
* Is it ok to have the body of the advisory here in RustSec be basically empty and point to the "official" source which for us is the GitHub-hosted version?

crates/wasmtime/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,20 @@
...
+[versions]
+patched = [">= 0.33.1, < 0.34.0", ">= 0.34.1"]
+```
+
+# Invalid drop of VMExternRef from partially-initialized instances in the pooling instance allocator
+
+This is an entry in the RustSec database for the Wasmtime security advisory
+located at
+https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-88xq-w8cq-xfg7.
+For more information see the GitHub-hosted security advisory.
```
riking, line 20: @peterhuene, can you release the copyright on the GHSA text here?

peterhuene: Hi @riking. What is the process to do that? If all that is required is a comment stating I release any and all copyright on the text of the advisory, I do so herein.

If there's something more formal required, please let me know.

riking: Yup, that's good. The Rustsec advisory DB is CC-0, so any imports from GHSA need separate copyright releases. @alexcrichton, ideally you can get that release before submitting the PR :)

tarcieri:
> Is there a way to import multiple advisories at once? Or should I make a separate PR-per-advisory for the initial backfill?

@alexcrichton you can submit several advisories in a single PR, and the ID assigner will make a single PR to assign them all incrementing IDs

djc:
> Ok thanks! When adding multiple advisories should I use `RUSTSEC-0000-0000.XX.md` as a naming scheme? It looked like the renamer specifically looked for `"RUSTSEC-0000-0000"` and otherwise discarded the filename.

That sounds about right, let's try it?

> Happy to include the renames, but I'm a bit hesitant to import too much if this will end up pinging many other contributors asking for licensing information. Is that something standard to require? Even for just a link to GHA? Is there something we can do to avoid the licensing back-and-forth?

The [full guidance](https://github.com/rustsec/advisory-db/blob/main/MAINTAINERS_GUIDE.md#check-if-the-advisory-came-from-ghsa):

> * Either use only data published in the associated CVE, since the CVE database is in the public domain
> * Or ask the submitter of the GHSA advisory (not the RustSec pull request author) to release the advisory contents into the public domain

Are wasmtime vulnerabilities assigned a CVE? If so, using just the title + link should be fine assuming that the title is also used in the CVE. Otherwise, it looks like there are just 5 GHSA authors for wasmtime, so getting those releases should be fairly straightforward? (Also presumably if they were employed by a single organization at the time they submitted, it might be enough to get a copy right release from that organization?)

alexcrichton:
It'll take me some time to get around to filling out old advisories. I'm happy to have this merged as-is and I can follow-up with a later batch, but if y'all would prefer this can hold off on merging until I've imported other old advisories.

All Wasmtime vulnerabilities are assigned a CVE, and for example this advisory has a CVE listed in it which you can see on the [GHSA as well](https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-88xq-w8cq-xfg7). The [CVE itself](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-23636) doesn't seem to be a 1:1 copy of what's on GHSA. 

Basically if licensing is going to be a drag on this I think we'll probably skip the step of manually publishing here and wait for the auto-importing tool to work (or something like that). If we can sidestep licensing issues by having the report be a URL that's fine though from our end, the goal here is to just get notifications out to users who exclusively use `cargo audit` that a vulnerability has happened.

tarcieri:
@alexcrichton you can also add a `license = "CC0-1.0"` field to each advisory

alexcrichton:
Ok found some time on my hands and I've uploaded other advisories we've had. I've also added `license = "CC0-1.0"` to each entry

alexcrichton:
Ah alas the `RUSTSEC-0000-0000.XX.md` trick didn't work since then the filename doesn't match the id. In lieu of 16 PRs is it possible to merge without CI passing? Or should I trim this back to one advisory and send more PRs for other advisories?

djc:
We should probably make sure the underlying tools are aligned on their understanding of what constitutes a valid new advisory filename. I'll see if I can have a look at that sometime soon (or maybe you can make a PR to the rustsec repo?).

tarcieri:
To my knowledge it’s files named RUSTSEC-0000-0000.md

djc:
Here's a proposal to improve the tooling to make it possible to have multiple advisories for the same crate in one PR:

https://github.com/rustsec/rustsec/pull/1351

alexcrichton:
I tweaked CI a bit to test out `rustsec-admin` from git and looks like it's green and working, so I'll drop those bits once a new version is published and update to that as well.

djc:
This needs to wait for rustsec-admin 0.8.10, which is blocked on getting a new rustsec release out.

djc:
@alexcrichton sorry for the delay -- the updated rustsec-admin should be integrated into the CI workflow here now, so can you rebase on top of current main?

alexcrichton:
Sure thing,  rebased and updated the commit message/PR title too

djc (approved):

PR #2255, opened 2025-03-22
Add advisory for Stack-buffer-overflow in xmas-elf

nyw0102:


djc:
I think we're going to want to wait to merge this until xmas-elf 0.10 has actually been released. I also don't find the advisory text very clear -- @khuey maybe you can help cleaning this up a bit?

khuey:
I don't have access to the crates.io for the crate so I don't know when a release will be published. I have emailed nrc.

A better description of the issue is that a malformed ELF file could cause an xmas-elf consumer using the `HashTable::get_bucket` or `HashTable::get_chain` functions to do an out-of-bounds read.

djc:
Might be easiest if you're up for it.

nyw0102:
Thank you for your comment! I also think it might be great if you create new PR with better explanation

khuey:
#2260

djc:
Thanks for getting the ball rolling on this!

PR #2268, opened 2025-04-04
Advisory for CVE-2025-31130 (weak SHA-1) in gix-features

EliahKagan:
This adds a notice for CVE-2025-31130 (GHSA-2frx-2596-x5r6) in `gix-features`.

Although `gix-features` is usually used as a "knob" for adjusting features across `gix-*` crates, as well as being used internally:

- The functionality replaced to fix the vulnerability was in `gix-features`.
- All vulnerable crates depend directly or indirectly on `gix-features`.

My understanding is that the usual practice for RUSTSEC advisories in such a situation is to have only an advisory for that one crate, as in https://github.com/rustsec/advisory-db/pull/1705#issuecomment-1575170900. So this adds a RUSTSEC notice only for `gix-features`. (The published repository-local GHSA, and forthcoming global GHSA, are not limited to one affected crate per advisory, and accordingly list more crates.)
 
As in some past advisories, when a global GHSA is published, I can open another PR to add a reference to that.

cc @Byron @emilazy

### DIFF
--- a/crates/gix-features/RUSTSEC-0000-0000.md
+++ b/crates/gix-features/RUSTSEC-0000-0000.md
@@ -10,6 +10,17 @@ keywords = ["hash-collision", "sha-1", "weak-hash"]
 aliases = ["CVE-2025-31130", "GHSA-2frx-2596-x5r6"]
 license = "CC0-1.0"
 
+[affected.functions]
+"gix_features::hash::bytes_of_file" = ["< 0.41.0"]
+"gix_features::hash::bytes" = ["< 0.41.0"]
+"gix_features::hash::bytes_with_hasher" = ["< 0.41.0"]
+"gix_features::hash::hasher" = ["< 0.41.0"]
+"gix_features::hash::Hasher::update" = ["< 0.41.0"]
+"gix_features::hash::Hasher::digest" = ["< 0.41.0"]
+"gix_features::hash::Write::new" = ["< 0.41.0"]
+"gix_features::hash::Write::write" = ["< 0.41.0"]
+"gix_features::hash::Write::flush" = ["< 0.41.0"]
+
 [versions]
 patched = [">= 0.41.0"]
 ```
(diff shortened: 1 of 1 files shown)


crates/gix-features/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,90 @@
...
+date = "2025-04-03"
+url = "https://github.com/GitoxideLabs/gitoxide/security/advisories/GHSA-2frx-2596-x5r6"
+categories = ["crypto-failure"]
+cvss = "CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:C/C:N/I:H/A:N"
+keywords = ["hash-collision", "sha-1", "weak-hash"]
+aliases = ["CVE-2025-31130", "GHSA-2frx-2596-x5r6"]
+license = "CC0-1.0"
+
+[affected.functions]
+"gix_features::hash::bytes_with_hasher" = ["< 0.41.0"]
```
emilazy, line 14: Is there a reason not to list the other I/O hashing functions here? They all call into `bytes_with_hasher`, but then `bytes_with_hasher` calls into the methods on `Hasher` too.

emilazy (commented):
I confirm that I agreed to release the advisory text under CC0 1.0 Universal. Thanks to @EliahKagan and @Byron for making handling this easy and enjoyable!

crates/gix-features/RUSTSEC-0000-0000.md
EliahKagan: Thanks--this is probably just an oversight on my part. I'll look at this shortly and try to improve it.

EliahKagan: I've added `bytes_of_file` and `bytes`. I'm not sure why I had thought those shouldn't be listed in `affected.functions`, but you are correct that they should be: they are public functions in the affected crate that invoke the vulnerable functionality. Thanks again!

djc (approved):

PR #2275, opened 2025-04-08
Add references to RUSTSEC-2025-0021

EliahKagan:
Since it was added in #2268, [RUSTSEC-2025-0021](https://rustsec.org/advisories/RUSTSEC-2025-0021.html) (CVE-2025-31130) has an entry in the GitHub Advisory Database. As planned in #2268, this adds the link to that global GHSA, as well as to the National Vulnerability Database entry for the CVE.

Some additional information may end up being posted in the pull request or issue, in which case one or both of those may make sense to add as references, but I am not sure yet. If so, I can add those separately. But I didn't want to add these references as planned.


djc (approved):

djc:
Thanks!

Issue #2283, opened 2025-04-23
RUSTSEC-2020-0071 and RUSTSEC-2020-0159 seem to have mixed up information


Oakchris1955:
[RUSTSEC-2020-0071](https://rustsec.org/advisories/RUSTSEC-2020-0071.html) (time crate) has info that seems to belong to [RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159.html) (chrono crate), while the last one is almost empty and references a time-rs issue.

Both security advisories are related to CVE-2020-26235, so I can see how this could have happened

djc:
Given that these are pretty old advisories and aren't very relevant to current crate versions, I don't think I'm motivated to go around updating these RFCs at this point, but if you were to submit a PR I would review it.

Oakchris1955:
I will close this once #2285 is merged

PR #2285, opened 2025-04-24
Fix mixed up advisories RUSTSEC-2020-0071 and RUSTSEC-2020-0159

Oakchris1955:
This PR was opened as a solution to #2283

### DIFF
--- a/crates/time/RUSTSEC-2020-0071.md
+++ b/crates/time/RUSTSEC-2020-0071.md
@@ -72,26 +72,8 @@ Users of time 0.1 do not have a patch and should upgrade to an unaffected versio
 
 ### Workarounds
 
-A possible workaround for crates affected through the transitive dependency in `chrono`, is to avoid using the default `oldtime` feature dependency of the `chrono` crate by disabling its `default-features` and manually specifying the required features instead.
+No workarounds are known.
 
-#### Examples:
+### References
 
-`Cargo.toml`:  
-
-```toml
-chrono = { version = "0.4", default-features = false, features = ["serde"] }
-```
-
-```toml
-chrono = { version = "0.4.22", default-features = false, features = ["clock"] }
-```
-
-Commandline:  
-
-```bash
-cargo add chrono --no-default-features -F clock
-```
-
-Sources:  
- - [chronotope/chrono#602 (comment)](https://github.com/chronotope/chrono/issues/602#issuecomment-1242149249)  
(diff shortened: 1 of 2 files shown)


Oakchris1955:
@djc can you merge?

djc:
First, I think your ping is pretty quick after the initial submission of this PR -- I'm volunteer with many other things to do.

Second, it's not obvious to me that this is an improvement. When these issues were discovered I'm pretty sure time 0.3 had already been published for a decent amount of time and so the main usage was via the chrono crate. As such, users who encountered the time 0.1 issue were exceedingly likely to be relying on time 0.1 via chrono. On the other hand, as I recall (as a chrono maintainer at the time), chrono itself contained a different issue that was unrelated to time.

Third, again -- why do you care about this? These advisories are pretty old at this point and unlikely to have much of an audience in the future.

oherrala:
> Third, again -- why do you care about this? These advisories are pretty old at this point and unlikely to have much of an audience in the future.

Future researchers of vulnerabilities probably enjoy if facts are straightened even after a long time has passed. This repository is one of the best collections of information to research Rust vulnerabilities.

djc:
> Future researchers of vulnerabilities probably enjoy if facts are straightened even after a long time has passed. This repository is one of the best collections of information to research Rust vulnerabilities.

It's not like there are factual inaccuracies here, and all of the content is prose, not parsable metadata. Also, my main motivation in volunteering my time to work on this project is to help folks discover vulnerabilities -- not assisting future researchers.

Oakchris1955:
> why do you care about this?

I spent 30 minutes of my life trying to figure out what was going on with these advisories were mixed up, that's why

crates/time/RUSTSEC-2020-0071.md
```diff
@@ -72,26 +72,8 @@ Users of time 0.1 do not have a patch and should upgrade to an unaffected versio
 
 ### Workarounds
 
-A possible workaround for crates affected through the transitive dependency in `chrono`, is to avoid using the default `oldtime` feature dependency of the `chrono` crate by disabling its `default-features` and manually specifying the required features instead.
+No workarounds are known.
```
tarcieri, line 75: Looking at this again, I agree with @djc that the original text does make sense since it's talking about `time` as a `chrono` dependency, and offering workarounds to avoid that, which seems fine to me.

Apologies if I said something else before, I missed this particular bit of context.

PR #2287, opened 2025-04-25
fast-able possible unsound public API

charlesxsh:
Cannot find a way to create PR/issue over the upstream project, so I cannot reference a link to this report.

### DIFF
--- a/crates/fast-able/RUSTSEC-0000-0000.md
+++ b/crates/fast-able/RUSTSEC-0000-0000.md
@@ -1,0 +1,36 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "fast-able"
+date = "2025-04-25"
+categories = ["memory-corruption"]
+```
+
+# Possible unsound public API
+
+At src/vec.rs:
+
+```rust
+impl<V> SyncVec<V>
+where
+    V: Clone,
+{
+    pub fn to_vec(&self) -> Vec<V> {
+        let mut v = Vec::new();
+        for i in self.iter() {
+            v.push(i.clone());
+        }
+        v
+    }
+}
+
+impl<V> SyncVec<V> {
+    ...
+    #[inline]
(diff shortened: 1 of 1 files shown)


fintelia:
I don't think there's actually a soundness bug here.

The reason is a bit silly: the _src/vec.rs_ file isn't actually used when building the crate. Instead, [the relevant `mod` statement uses a `#[path]` attribute to replace the path with _src/vec2.rs_](https://docs.rs/fast-able/1.18.3/src/fast_able/lib.rs.html#12). I can't speak for the soundness of the overall crate, but you can see the real implementation of `SyncVec::get_uncheck` [doesn't have an unsafe block](https://docs.rs/fast-able/latest/src/fast_able/vec2.rs.html#174-181).

charlesxsh:
> I don't think there's actually a soundness bug here.
> 
> The reason is a bit silly: the _src/vec.rs_ file isn't actually used when building the crate. Instead, [the relevant `mod` statement uses a `#[path]` attribute to replace the path with _src/vec2.rs_](https://docs.rs/fast-able/1.18.3/src/fast_able/lib.rs.html#12). I can't speak for the soundness of the overall crate, but you can see the real implementation of `SyncVec::get_uncheck` [doesn't have an unsafe block](https://docs.rs/fast-able/latest/src/fast_able/vec2.rs.html#174-181).

Oh I should add more details. The version of the crate to have this issue is 1.11.7. Link: https://docs.rs/crate/fast-able/1.11.7/source/src/vec.rs. Would you verify does this version of code align with the description? If yes, I will add the detail version info.

fintelia:
Ah, version 1.11.7 does seem to be vulnerable. A quick way to check is to go to the docs.rs page for a specific version and click the "source" link

djc:
@guoyucode we'd like to publish an advisory this -- is that okay with you?

charlesxsh:
@djc should we publish it? no response from @guoyucode for months.

PR #2294, opened 2025-04-28
Add advisory for index error in mp3-metadata

llooFlashooll:


djc:
@GuillaumeGomez r? Any thoughts?

GuillaumeGomez:
Confirmed. I fixed the issue and added a regression test.

djc:
IMO it would be nice to have a bit more context around the actual issue -- how was the panic triggered (in terms of library semantics), ideally even adding affected function calls if possible. I don't think most of the audience will care that it was found via a fuzzer (except that it sort of diminishes the impact, I suppose, if it wasn't found in the real world), and that it was fixed is pretty obvious too.

djc:
Did some further editing. Thanks!

PR #2299, opened 2025-04-29
Install rustsec-admin from git

djc:
As suggested by @Shnatsel in https://github.com/rustsec/rustsec/pull/1356#issuecomment-2838409170.

### DIFF
--- a/.github/workflows/assign-ids.yml
+++ b/.github/workflows/assign-ids.yml
@@ -12,17 +12,15 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Cache cargo bin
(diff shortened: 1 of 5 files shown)


tarcieri:
Hmm, swore it was green when I approved it. The approach seems fine if you can figure out why the build failed

djc:
> Hmm, swore it was green when I approved it. The approach seems fine if you can figure out why the build failed

It was green on the first one, it fails the second time (I reran it to trigger the cached behavior).

djc:
Instead of keying on a version, we can rely on te cache actions hit directly to determine if we need to install.

PR #2309, opened 2025-05-06
RUSTSEC-2025-0030: Indicate 'totally-safe-transmute' is a toy.

ben0x539:
per #2221. Disclaimer: I'm the author of totally-safe-transmute.

People should know that totally-safe-transmute is a toy that's intentionally doing broken things, so that

a) they don't come after me

b) they know not to expect it to get patched

c) they question why someone is pulling toy code into their dependencies


djc (approved):

djc:
Thanks!

PR #2310, opened 2025-05-06
anon-vec unsound APIs and unmaintained

charlesxsh:
@RylanYancey Hi author of this crate, could you please verify that this crate is unmaintained? If the crate is still maintained, I can remove the unmaintance information


djc:
Sent an email directly to the crate's owner.

RylanYancey:
The crate is definitely unmaintained. I built it as a learning project and I believe it could have major vulnerabilities. I would be fine with removing it from crates.io entirely.

PR #2311, opened 2025-05-06
RUSTSEC-2025-0028: Indicate 'cve-rs' is a joke

Speykious:
Per #2221, following the footsteps of #2309.

The description is now slightly better - the sarcastic description is funny, but describing the implementation as "memory safe" would be categorically incorrect (I think).


djc (approved):
Thanks!

Issue #1471, opened 2022-11-25
surf seems to be unmaintained


djc:
The surf HTTP client seems to be unmaintained. I [reached out](https://github.com/http-rs/surf/issues/352) and it doesn't look good. It depends on the latest released version of http-types, which depends on an old version of cookie which in turns depends on the unmaintained stdweb, an old version of aes-gcm which in turns has a number of unmaintained dependencies. There's [an issue](https://github.com/http-rs/http-types/issues/505) in http-types to get a new release out but not much movement there.

surf itself additionally still depends on rustls 0.18, while 0.19 was released two years ago.

pinkforest:
Commented here: https://github.com/http-rs/surf/issues/352#issuecomment-1327778764

The maintainer has indicated willingness to merge any security fixes and per our policy we reserve unmaintained for completely unreachable maintainers or where the maintainer explicitly wishes us to flag it.

An action here might be feasible to flag old versions of rustls and then that will light up anything downstream if feasible ?

pinkforest:
Can't do anything here since the maintainer has said they will fix any security issue and we take the maintainer's word.

Also have offered to mark old versions of rustls unmaintained as the dependencies further up can be used to light up things

Please let us know if you would like us to do that and we can do that to ensure any downstream dependencies complain if they use the old rustls version.

djc:
There's an explicit comment from the maintainer that it should be considered unmaintained:

https://github.com/http-rs/surf/issues/352#issuecomment-1709404352

Is that enough to file this?

decathorpe:
It looks like nothing has changed in the last 1.5 years, so I think this makes the case for "this is really unmaintained" even stronger?

djc:
@decathorpe if you submit a PR I will review.

PR #2316, opened 2025-05-17
add unmaintained advisory for `surf`

decathorpe:
Based on discussion in https://github.com/rustsec/advisory-db/issues/1471#issuecomment-2873417980

I listed `reqwest` and `ureq` as possible alternatives, they seem to be the two most-used HTTP client libraries right now.


djc (approved):
Thanks!

PR #2322, opened 2025-05-22
pingora request smuggling and cache poisoning

Noah-Kennedy:
Pingora has a request smuggling and cache poisoning vulnerability affecting versions 0.5.0 and older, as documented here: https://blog.cloudflare.com/resolving-a-request-smuggling-vulnerability-in-pingora/

### DIFF
--- a/crates/pingora-core/RUSTSEC-0000-0000.md
+++ b/crates/pingora-core/RUSTSEC-0000-0000.md
@@ -4,9 +4,6 @@ id = "RUSTSEC-0000-0000"
 package = "pingora-core"
 date = "2025-05-22"
 url = "https://blog.cloudflare.com/resolving-a-request-smuggling-vulnerability-in-pingora/"
-# Valid categories: "code-execution", "crypto-failure", "denial-of-service", "file-disclosure"
-# "format-injection", "memory-corruption", "memory-exposure", "privilege-escalation"
-categories = []
 keywords = ["request-smuggling", "cache-poisoning"]
 aliases = ["CVE-2025-4366"]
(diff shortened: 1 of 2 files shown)


Noah-Kennedy:
None of the categories appear to quite fit this vuln.

Could I get some input from a maintainer on if leaving that blank is the correct move, or if one of the categories might encompass this vulnerability?

djc:
I think privilege escalation comes closest probably? This does allow an attacker to do things that they normally should not be able to do?

djc:
Also probably good to clean out the comments and give it a correct CVSS?

Noah-Kennedy:
> Also probably good to clean out the comments and give it a correct CVSS?

I can't really calculate a useful score here. It's very dependent on how you are using the framework and what you are doing. Though I suppose this criticism of CVSS scores can be made it 90% of cases.

I can clean out the comments though!

Noah-Kennedy:
> I think privilege escalation comes closest probably? This does allow an attacker to do things that they normally should not be able to do?

Maybe? I can construct scenarios where this could be used to bypass ACLs I suppose.

Noah-Kennedy:
But also, the wording of "privilege escalation" seems too focused for what you can do here.

Noah-Kennedy:
Actually, since there is a CVE now I will link that and use the existing score.

Noah-Kennedy:
Nevermind, I don't know if that one is finalized, removing for now

Noah-Kennedy:
@djc I've added the CVE now that it's live and also added the CVSS

Noah-Kennedy:
@djc I only have a CVSS 4.0, which is unsupported.

I'm going to drop it as it's in the linked CVE.

Noah-Kennedy:
I'm good with merging this now.

PR #2334, opened 2025-06-15
Add unmaintained advisory for `serde`

Mingun:
Unfortunately, after several attempts to get the `serde` maintenance business off the ground, I am forced to turn to more formal channels. I hope that the Rust community will treat the problem with due attention and perhaps offer some alternatives on how to help the project

### DIFF
--- a/crates/serde/RUSTSEC-0000-0000.md
+++ b/crates/serde/RUSTSEC-0000-0000.md
@@ -1,0 +1,52 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "serde"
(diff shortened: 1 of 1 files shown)


tarcieri:
Two different maintainers merged multiple commits last week. This does not meet our criteria of unmaintained.

Mingun:
As I say in description, this is mostly infrastructural changes which not solves real problems of users of the library. I provided several references of PRs abandoned for years, and they are abandoned not by their authors.

djc:
That is an issue to take up with the maintainers. From our perspective, these crates are being maintained.

PR #2336, opened 2025-06-16
DFs in slice-ring-buffer

GeorgeAndrou:
Four new memory safety bugs have been discovered in `slice-ring-buffer` (a fork of the unmaintained `slice-deque` crate). 

All four bugs can lead to double-free violations, when only safe APIs are used. No patches for these vulnerabilities have been developed yet. 

We have contacted the maintainers of the crate (@LiquidityC), and they have granted us permission to request a security advisory for these issues.

### DIFF
--- a/crates/slice-ring-buffer/RUSTSEC-0000-0000.md
+++ b/crates/slice-ring-buffer/RUSTSEC-0000-0000.md
@@ -1,0 +1,27 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "slice-ring-buffer"
+date = "2025-06-16"
+
+url = "https://github.com/LiquidityC/slice_ring_buffer/issues/12"
+categories = ["memory-corruption"]
+keywords = ["memory-safety", "double-free"]
+
+[affected.functions]
+"slice_ring_buffer::SliceRingBuffer::extend_from_slice" = ["<= 0.3.4"]
+"slice_ring_buffer::SliceRingBuffer::shrink_to_fit" = ["<= 0.3.4"]
+"slice_ring_buffer::SliceRingBuffer::insert" = ["<= 0.3.4"]
+"slice_ring_buffer::IntoIter::clone" = ["<= 0.3.4"]
+
+[versions]
+patched = []
+```
+
+# Four Unique Double-Free Vulnerabilities Triggered via Safe APIs
+
+The crate [`slice-ring-buffer`](https://crates.io/crates/slice-ring-buffer) was developed as a fork of [`slice-deque`](https://crates.io/crates/slice-deque) to continue maintenance and provide security patches, since the latter has been officially unmaintained ([RUSTSEC-2020-0158](https://rustsec.org/advisories/RUSTSEC-2020-0158.html)).
+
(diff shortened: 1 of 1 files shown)


djc:
@LiquidityC please confirm that you are okay with an advisory being published for your crate.

Also, are you planning to fix these issues? If that is planned for the short term, we typically like to hold off on publishing an advisory until the fixed version is published.

crates/slice-ring-buffer/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,27 @@
...
+"slice_ring_buffer::SliceRingBuffer::extend_from_slice" = ["<= 0.3.4"]
+"slice_ring_buffer::SliceRingBuffer::shrink_to_fit" = ["<= 0.3.4"]
+"slice_ring_buffer::SliceRingBuffer::insert" = ["<= 0.3.4"]
+"slice_ring_buffer::IntoIter::clone" = ["<= 0.3.4"]
+
+[versions]
+patched = []
+```
+
+# Four Unique Double-Free Vulnerabilities Triggered via Safe APIs
```
djc: Nit: I don't think title-casing is helpful here.

crates/slice-ring-buffer/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,27 @@
...
+
+The crate [`slice-ring-buffer`](https://crates.io/crates/slice-ring-buffer) was developed as a fork of [`slice-deque`](https://crates.io/crates/slice-deque) to continue maintenance and provide security patches, since the latter has been officially unmaintained ([RUSTSEC-2020-0158](https://rustsec.org/advisories/RUSTSEC-2020-0158.html)).
+
+While `slice-ring-buffer` has addressed some previously reported memory safety issues inherited from its fork origin ([RUSTSEC-2021-0047](https://rustsec.org/advisories/RUSTSEC-2021-0047.html)), it still retains multiple unresolved memory corruption vulnerabilities.
+
+Specifically, we have discovered **four new memory safety bugs**, each resulting in **double-free violations** that can occur **when only safe APIs are invoked**. These vulnerabilities correspond to four distinct safe APIs in the crate, each exposing unsound and vulnerable behavior due to incorrect usage of unsafe code internally.
```
djc: Nit: please drop the bold emphasis in this paragraph -- I don't think it helps in this case.

GeorgeAndrou: Done. Thanks!

crates/slice-ring-buffer/RUSTSEC-0000-0000.md
GeorgeAndrou: Done. Thanks!

LiquidityC:
@djc Confirmed that I'm aware of the issues. I currently have no prognosis on when the issues can be fixed. I'm personally busy with work and the only reason I'm a maintainer is due to the original repo and maintainer becoming unreachable so I created this fork to address issues in the original repo.
Hoping that someone in the community might be able to address the issues earlier then I will be able to.

djc:
@GeorgeAndrou can you add something like this to the advisory at the end:

> Unfortunately, the maintainer doesn't have much availability to resolve these issues so there's no concrete timeline for fixes. Community contributions towards fixing these vulnerabilities would be much appreciated.

Then we can publish it -- hopefully that will help in finding someone who can fix these issues.

GeorgeAndrou:
@djc I committed an updated version of the bug report.

PR #2339, opened 2025-06-27
Add advisory for timing variability in curve25519-dalek-ng

fspreiss:
The timing variability in `curve25519-dalek` that was reported via [RUSTSEC-2024-0344](https://rustsec.org/advisories/RUSTSEC-2024-0344) is also applicable to its fork [`curve25519-dalek-ng`](https://crates.io/crates/curve25519-dalek-ng).

This is relevant because crates such as [`ed255519-consensus`](https://crates.io/crates/ed25519-consensus), which are actively used by the community for secret key operations, depend on the vulnerable `curve25519-dalek-ng`.

### DIFF
--- a/crates/curve25519-dalek-ng/RUSTSEC-0000-0000.md
+++ b/crates/curve25519-dalek-ng/RUSTSEC-0000-0000.md
@@ -19,9 +19,12 @@ patched = []
 Timing variability of any kind is problematic when working with  potentially secret values such as elliptic curve scalars, and such issues can potentially leak private keys and other secrets.
 
 To patch the vulnerability, backports of
-* https://github.com/dalek-cryptography/curve25519-dalek/pull/659, and
-* https://github.com/dalek-cryptography/curve25519-dalek/pull/661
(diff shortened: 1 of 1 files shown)


djc:
As a matter of policy, we don't publish advisories without confirmation/agreement from the maintainer, unless the maintainer becomes unresponsive (for 270 days in case of no open issues or 60 days in case of a potential vulnerability).

@hdevalence are you aware of this issue? Do you agree it warrants an advisory?

hdevalence:
I don't think this issue warrants an advisory, but I don't think it's particularly important either way. It should be fixed shortly, just fell through the cracks and there was no follow up.

djc:
> I don't think this issue warrants an advisory,

Why not? If the pre-fork code had one, it seems surprising that this crate wouldn't get one.

fspreiss:
Are there maybe any updates regarding this?

djc:
> Are there maybe any updates regarding this?

It's up to @hdevalence.

PR #2347, opened 2025-07-17
static_cell: report unsoundness in `ConstStaticCell`

ROMemories:
I am not the owner of `static_cell`, but [the owner has manifested their intention not to file an advisory themself](https://github.com/embassy-rs/static-cell/issues/19#issuecomment-3073545476).

### DIFF
--- a/crates/static_cell/RUSTSEC-0000-0000.md
+++ b/crates/static_cell/RUSTSEC-0000-0000.md
@@ -1,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "static_cell"
+date = "2025-07-17"
+url = "https://github.com/embassy-rs/static-cell/issues/19"
+informational = "unsound"
+categories = ["memory-exposure", "memory-corruption"]
+keywords = ["send", "thread-safety"]
+
+[versions]
+patched = [">= 2.1.1"]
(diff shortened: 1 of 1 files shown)


crates/static_cell/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "static_cell"
+date = "2025-07-17"
```
ROMemories, line 5: This is intended to be date of when *the advisory* is made public, not the reported issue, right?

djc:
@Dirbaio I'd like your consent before publishing an advisory for your crate.

Dirbaio:
No concerns from me :+1:

djc (approved):

crates/static_cell/RUSTSEC-0000-0000.md
djc: I think so, I don't know that we have strict guidelines around this.

tarcieri: It's typically been the date the issue was publicly disclosed, e.g. if a vulnerability has been known for several years and retroactively receives an advisory, the year in the advisory and the date field reflect the age of the vulnerability

PR #2353, opened 2025-07-29
Add unmaintained advisory for tsify-next

threema-theo:
See https://github.com/siefkenj/tsify/issues/57#issuecomment-3132840632 .

### DIFF
--- a/crates/tsify-next/RUSTSEC-0000-0000.md
+++ b/crates/tsify-next/RUSTSEC-0000-0000.md
@@ -1,0 +1,15 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tsify-next"
+date = "2025-07-29"
+url = "https://github.com/siefkenj/tsify/pull/56"
+references = ["https://github.com/siefkenj/tsify/issues/57"]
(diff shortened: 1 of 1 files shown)


crates/tsify-next/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,15 @@
...
+url = "https://github.com/siefkenj/tsify/pull/56"
+references = ["https://github.com/siefkenj/tsify/issues/57"]
+informational = "unmaintained"
+[versions]
+patched = []
+```
+
+# tsify-next is unmaintained, use tsify instead
+
+The `tsify-next` crate is not maintained any more; use [`interfaces`](https://crates.io/crates/tsify) instead.
```
djc: I'm confused why you're using `interfaces` as link text for a link to the `tsify` crate instead?

djc (commented):
Thanks!

crates/tsify-next/RUSTSEC-0000-0000.md
threema-theo: my bad :see_no_evil:

PR #2355, opened 2025-08-05
`xcb` `connect_to_fd*` constructors unsound

En-En-Code:
Identifies a flaw in `xcb::Connection::connect_to_fd` and `xcb::Connection::connect_to_fd_with_extensions` which allows file descriptors to be used after close and/or double closed. An [issue was raised back in 2022](https://github.com/rust-x-bindings/rust-xcb/issues/167) to mark these functions as unsafe with no additional context. I discovered this unsoundness causing problems working on a different project and [reported my findings to xcb](https://github.com/rust-x-bindings/rust-xcb/issues/282).
It is unclear as to whether the project is maintained, though previously there was an issue raised as to its maintained status (see #653).
I did not put any categories or keywords. I just did not know what to put for them.

### DIFF
--- a/crates/xcb/RUSTSEC-0000-0000.md
+++ b/crates/xcb/RUSTSEC-0000-0000.md
@@ -4,26 +4,29 @@ id = "RUSTSEC-0000-0000"
 package = "xcb"
 date = "2025-08-05"
 url = "https://github.com/rust-x-bindings/rust-xcb/issues/282"
-references = ["https://github.com/rust-x-bindings/rust-xcb/issues/167"]
+references = [
+  "https://github.com/rust-x-bindings/rust-xcb/issues/167",
+  "https://github.com/rust-x-bindings/rust-xcb/pull/283"
+]
 informational = "unsound"
 
 [versions]
-patched = []
+patched = [">= 1.6.0"]
 
 [affected.functions]
 "xcb::Connection::connect_to_fd" = [">= 1.0.0-beta.0"]
 "xcb::Connection::connect_to_fd_with_extensions" = [">= 1.0.0-beta.0"]
 ```
 
-# Unsoundness in `xcb::Connection::connect_to_fd*` functions
+# `xcb::Connection::connect_to_fd*` functions violate I/O safety
 
 The API of `xcb::Connection` has constructors which allow an arbitrary `RawFd`
 to be used as a socket connection. On either failure of these constructors or
 on the drop of `Connection`, it closes the associated file descriptor. Thus, a
 program which uses an `OwnedFd` (such as a `UnixStream`) as the file descriptor
 can close the file descriptor and continue to attempt using it or close an
-already-closed file descriptor.
+already-closed file descriptor, violating I/O safety.
 
-The functions not being safe was previously documented, although with no
-specifics (https://github.com/rust-x-bindings/rust-xcb/issues/167). No action
-was taken.
+Starting in version 1.6.0, `xcb` provides `Connection::connect_with_fd` and
+`Connection::connect_with_fd_and_extensions` as safe alternatives and
+deprecates the problematic functions.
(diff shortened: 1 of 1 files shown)


djc:
I've emailed @rtbo personally (they seem somewhat active on GitHub at least), let's see if they respond in the next two weeks. If not, we should go ahead and merge the advisory (please help remind me).

rtbo:
Hi,
I've honestly little time to allocate to the maintenance of `rust-xcb`, but I don't want it to die unmaintained either!
I can definitely share a write access to the repo.

fintelia:
One thing to mention is that this isn't a memory safety violation but rather an [I/O safety violation](https://rust-lang.github.io/rfcs/3128-io-safety.html). I think that may warrant tagging it with a dedicated category? It also wasn't clear to me from reading the discussions linked from that RFC whether I/O safety violations can cause undefined behavior, and thus whether this qualifies as "unsound" at all.

I also experimented with the nix crate a bit and discovered that you can trigger the same crash described here simply by opening a file with `File::open`, getting its raw file descriptor via [`as_raw_fd`](https://doc.rust-lang.org/std/os/fd/trait.AsRawFd.html#tymethod.as_raw_fd), and then closing it with both the (safe) [`nix::unistd::close`](https://docs.rs/nix/latest/nix/unistd/fn.close.html) file and via dropping the `File`. If this really constitutes undefined behavior... I feel like a crate as widely used as nix would've been fixed already?

rtbo:
Should be closed by https://github.com/rust-x-bindings/rust-xcb/pull/283

djc:
> One thing to mention is that this isn't a memory safety violation but rather an [I/O safety violation](https://rust-lang.github.io/rfcs/3128-io-safety.html). I think that may warrant tagging it with a dedicated category? It also wasn't clear to me from reading the discussions linked from that RFC whether I/O safety violations can cause undefined behavior, and thus whether this qualifies as "unsound" at all.

That's a good point, this probably doesn't qualify as `unsound`.

En-En-Code:
Sorry for not having a quicker response. I guess I can give my opinions (to be taken with a grain of salt).

As long as someone can respond within a reasonable amount of time to major `xcb` issues, I have no problem calling it maintained. It's clear to me that has been done here.

Maybe I am missing something, but the neither the Reference nor the Nomicon nor the FLS expressly state that calling foreign code in a way which causes undefined behavior in the foreign code (for whatever its definition of UB may be) to be undefined behavior in Rust. UB should cross FFI boundaries, because the opposite is absurd. That was my defense for calling this unsound: the POSIX standard says closing or using a file descriptor after close is UB, safe functions provided by `xcb` enable that to occur.

I/O safety is in this weird place. Its main PR landed 3 years ago, but no one ever defined what constitutes UB when using a `RawFd`. The functions in the standard library which takes a `RawFd` and return some other file-like type, `BorrowedFd::borrow_raw` and `FromRawFd::from_raw_fd`, are marked `unsafe` to enforce I/O safety. But with I/O safety not being "real" (in the sense that it also is not defined in any of the aforementioned places) and `RawFd`s usage being almost completely exclusive to FFI (where misuse should be an unsoundness anyways), I/O safety never got a category. It doesn't even have a keyword somehow.

I believe the `nix` crate to be in the wrong, and the lack of a rigorous I/O safety definition contributing to it sticking around. The equivalent function in `rustix`, [`rustix::io::close`](https://docs.rs/rustix/latest/rustix/io/fn.close.html), is marked unsafe. More broadly, I believe most "safe" functions which allow the user to provide a `RawFd` and do syscalls with it are unlikely to be sound. As example,
```
fn main() {
    let file_fd = File::create_new("hello_world.txt").unwrap();
    let raw_fd = file_fd.as_raw_fd();
    drop(file_fd);
    // This is sound, a RawFd is just a number
    println!("raw_fd: {raw_fd:?}");
    // I believe this should be undefined if it does something undefined in C.
    unsafe { libc::some_syscall(raw_fd) };
}
```

Finally, I've made changes to the advisory now that a fixed version is out. I have still left it with no keyword and `informational = unsound` since there seems to be no consensus on a more apt label.

djc (approved):

djc:
Thanks!

PR #2363, opened 2025-08-14
The API ArrayQueue::push_front is not panic-safe

GeorgeAndrou:
The safe API `array_queue::ArrayQueue::push_front` of the crate `array-queue`, can lead to deallocating uninitialized memory if a panic occurs while invoking the `clone` method on the passed argument.

This issue was reported to the crate's GitHub repository ([issue #3](https://github.com/raviqqe/array-queue/issues/3)) 30 days ago, but there has been no response from the maintainers as of yet.

### DIFF
--- a/crates/array-queue/RUSTSEC-0000-0000.md
+++ b/crates/array-queue/RUSTSEC-0000-0000.md
@@ -1,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "array-queue"
+date = "2025-08-14"
+
+url = "https://github.com/raviqqe/array-queue/issues/3"
+categories = ["memory-corruption"]
+keywords = ["memory-safety", "panic-safety", "uninitialized-memory"]
+
+[affected.functions]
+"array_queue::ArrayQueue::push_front" = [">= 0.3.0, <= 0.3.3"]
+
+[versions]
+patched = []
+unaffected = ["< 0.3.0"]
+```
+
+# ArrayQueue::push_front is not panic-safe
+
+The safe API `array_queue::ArrayQueue::push_front` can lead to deallocating uninitialized memory if a panic occurs while invoking the `clone` method on the passed argument.
+
+Specifically, `push_front` receives an argument that is intended to be cloned and pushed, whose type implements the `Clone` trait. Furthermore, the method updates the queue's `start` index before initializing the slot for the newly pushed element. User-defined implementations of `Clone` may include a `clone` method that can panic. If such a panic occurs during initialization, the structure is left with an advanced `start` index pointing to an uninitialized slot. When `ArrayQueue` is later dropped, its destructor treats that slot as initialized and attempts to drop it, resulting in an attempt to free uninitialized memory.


djc:
Reached out on Bluesky: https://bsky.app/profile/djc.ochtman.nl/post/3lwew7fb7rs2x.

fintelia:
The crate seems to be unsound for a far more basic reason: `ArrayQueue::new` uses `mem::unitialized` and the docs for that method now state...
 > Therefore, it is immediate undefined behavior to call this function on nearly all types, including integer types and arrays of integer types, and even if the result is unused.

This isn't the fault of the crate author since those docs were changed after the code was written. But perhaps it is a good reminder that if you decide to on a crate last updated seven years ago that is used in zero other crates on crates.io, you're doing so at your own risk...

GeorgeAndrou:
Hi @fintelia,
You're right about the use of `mem::uninitialized`, and there is a separate open issue reporting it: [issue #2](https://github.com/raviqqe/array-queue/issues/2).

fintelia:
That issue is only about the `pop_back` method. The more fundamential problem is that [`new` unconditionally calls `mem::uninitialized`](https://github.com/raviqqe/array-queue/blob/32fa10f8f15140fb64a4cf36a2a834f876c91056/src/array_queue.rs#L17) and [`drop` calls it for any non-empty queue](https://github.com/raviqqe/array-queue/blob/32fa10f8f15140fb64a4cf36a2a834f876c91056/src/array_queue.rs#L165).

Like, this is the output from miri:

<pre><font color="#F66151"><b>error</b></font><b>: Undefined Behavior: constructing invalid value at .value.value[0].0.vec.buf.inner.ptr.pointer.pointer: encountered uninitialized memory, but expected a raw pointer</b>
  <font color="#2A7BDE"><b>--&gt; </b></font>/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/array-queue-0.3.3/src/array_queue.rs:17:29
   <font color="#2A7BDE"><b>|</b></font>
<font color="#2A7BDE"><b>17</b></font> <font color="#2A7BDE"><b>|</b></font>             array: unsafe { uninitialized() },
   <font color="#2A7BDE"><b>|</b></font>                             <font color="#F66151"><b>^^^^^^^^^^^^^^^</b></font> <font color="#F66151"><b>constructing invalid value at .value.value[0].0.vec.buf.inner.ptr.pointer.pointer: encountered uninitialized memory, but expected a raw pointer</b></font>
   <font color="#2A7BDE"><b>|</b></font>
   <font color="#2A7BDE"><b>= </b></font><b>help</b>: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
   <font color="#2A7BDE"><b>= </b></font><b>help</b>: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
   <font color="#2A7BDE"><b>= </b></font><b>note</b>: BACKTRACE:
   <font color="#2A7BDE"><b>= </b></font><b>note</b>: inside `array_queue::ArrayQueue::&lt;[StructA; 2]&gt;::new` at /home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/array-queue-0.3.3/src/array_queue.rs:17:29: 17:44
<font color="#33DA7A"><b>note</b></font>: inside `main`
  <font color="#2A7BDE"><b>--&gt; </b></font>src/main.rs:16:21
   <font color="#2A7BDE"><b>|</b></font>
<font color="#2A7BDE"><b>16</b></font> <font color="#2A7BDE"><b>|</b></font>     let mut queue = ArrayQueue::&lt;[StructA; 2]&gt;::new();
   <font color="#2A7BDE"><b>|</b></font>                     <font color="#33DA7A"><b>^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^</b></font>

<font color="#33DA7A"><b>note</b></font><b>: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace</b>
</pre>

GeorgeAndrou:
@fintelia,

Thanks for the feedback regarding Miri's output. You're absolutely right that Miri can detect uses of `mem::uninitialized` and help prevent them.

However, in this issue, I’m specifically reporting a panic-safety bug in the crate, which I believe could also be triggered even if alternatives to `mem::uninitialized` were used.

If you check the second case of the PoC provided in [issue #2](https://github.com/raviqqe/array-queue/issues/2), along with the accompanying explanation, you'll see that it directly matches the issue you're referring to.

fintelia:
Yes, a hypothetical version of the crate that didn't use `mem::unitialized` could have a panic safety bug. But since your proof-of-concept hits unsoundness _before even getting to the panic_, I don't think it is accurate to say that the panic is the problem.

But stepping back, what's the point of all of this? The `array-queue` crate doesn't seem to have anyone using it. The last release was over 7 years ago, so it doesn't seem to be under development anymore either. For most of that time [there has been an advisory posted](https://rustsec.org/advisories/RUSTSEC-2020-0047.html) about an unpatched soundness issue. The bug report for that issue noted in passing that not only was the one method broken, but that any usage of the crate whatsoever would result in unsoundness.

djc:
The author responded on Bluesky: https://bsky.app/profile/raviqqe.bsky.social/post/3lwf56dkw5k2z.

GeorgeAndrou:
@djc I think we can wait for two weeks to see if the author provides a patch. If not, we can revisit the discussion after that period.

GeorgeAndrou:
@djc The two weeks have passed since we notified the crate author, who also agreed to publish the advisory. I think we can close it.

raviqqe:
Hi, sorry for the extra work I incurred to you all... I just published the new version that doesn't `clone` an element on push to the queue. I've also fixed the `pop_back` logic issue.

fintelia:
No, you didn't cause extra work. The blame for that falls on the security researchers searching for bugs in unused software and then pressuring community members to investigate and fix them. Sorry they wasted your time.

djc:
@GeorgeAndrou can you add the fix version?

@raviqqe no worries, thanks for fixing! If you just want to stop maintaining the crate, maybe marking it as unmaintained might be better?

@fintelia I for one appreciate when soundness issues in software published on crates.io are being pointed out and (ideally) getting fixed.

djc (approved):

PR #2366, opened 2025-08-14
IdMap::from_iter may lead to uninitialized memory being freed on drop

GeorgeAndrou:
The constructor `IdMap::from_iter` in the crate `id-map` can create objects where the amount of actually initialized memory is less than what can be accessed through the resulting data structure. This can lead to an attempt to free uninitialized memory during the dropping of such objects, resulting in undefined behavior and potential segmentation faults.

This issue was reported to the crate's GitHub repository ([issue #4](https://github.com/andrewhickman/id-map/issues/4)) 10 days ago, but there has been no response from the maintainers as of yet.

### DIFF
--- a/crates/id-map/RUSTSEC-0000-0000.md
+++ b/crates/id-map/RUSTSEC-0000-0000.md
@@ -22,4 +22,8 @@ Due to a flaw in the constructor `id_map::IdMap::from_iter`, ill-formed objects 
 
 In such cases, when the resulting `IdMap` is dropped, its destructor incorrectly assumes that `values` contains `ids.len() == values.capacity()` initialized elements and attempts to iterate over and drop them. This leads to dereferencing and attempting to free uninitialized memory, resulting in undefined behavior and potential segmentation faults.
 
-The bug was fixed in commit `fab6922`, and all unsafe code was removed from the crate. Note that better-optimized alternatives are [slab](https://crates.io/crates/slab) and [slotmap](https://crates.io/crates/slotmap).
+The bug was fixed in commit `fab6922`, and all unsafe code was removed from the crate.
+
+Note that the maintainer recommends using the following alternatives:
+- [slab](https://crates.io/crates/slab)
+- [slotmap](https://crates.io/crates/slotmap)


djc:
I've pinged the maintainer in the issue -- I've previously had a good experience with this maintainer and they've recently been active on GitHub, so I think we have a decent chance of getting a response from them.

djc:
Can you add the fixed version in the advisory?

GeorgeAndrou:
@djc Can I also include the alternatives at the end of the advisory?

djc:
> @djc Can I also include the alternatives at the end of the advisory?

Yes, that's what I was suggesting.

GeorgeAndrou:
done!

crates/id-map/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
...
+unaffected = ["< 0.1.6"]
+```
+
+# IdMap::from_iter may lead to uninitialized memory being freed on drop
+
+Due to a flaw in the constructor `id_map::IdMap::from_iter`, ill-formed objects may be created in which the amount of actually initialized memory is less than what is expected by the fields of `IdMap`. Specifically, the field `ids` is initialized based on the capacity of the vector `values`, which is constructed from the provided iterator. However, the length of this vector may be smaller than its capacity.
+
+In such cases, when the resulting `IdMap` is dropped, its destructor incorrectly assumes that `values` contains `ids.len() == values.capacity()` initialized elements and attempts to iterate over and drop them. This leads to dereferencing and attempting to free uninitialized memory, resulting in undefined behavior and potential segmentation faults.
+
+The bug was fixed in commit `fab6922`, and all unsafe code was removed from the crate. Note that better-optimized alternatives are [slab](https://crates.io/crates/slab) and [slotmap](https://crates.io/crates/slotmap).
```
djc: The maintainer said:

> I would generally recommend a more widely-used crate like https://crates.io/crates/slotmap or https://crates.io/crates/slab for this purpose.

It's not obvious to me that this was about optimization -- maybe say something more general like, "The maintainer recommends using slab or slotmap instead." (including links)?

PR #2370, opened 2025-08-22
Add `io-safety` keyword to `xcb`

En-En-Code:
Thanks for getting back quickly to my response. Honestly was not expecting the quick turn-around.

Since it being merged makes issues like these within RustSec's purview, I would like to add `io-safety` as a free-form keyword, hopefully to set a precedent that I/O safety violations are problems which should be reported.

### DIFF
--- a/crates/xcb/RUSTSEC-2025-0051.md
+++ b/crates/xcb/RUSTSEC-2025-0051.md
@@ -9,6 +9,7 @@ references = [
   "https://github.com/rust-x-bindings/rust-xcb/pull/283"
 ]
 informational = "unsound"
+keywords = [ "io-safety" ]
 
 [versions]
 patched = [">= 1.6.0"]


djc:
Seems okay, but the branch has conflicts. Want to rebase?

En-En-Code:
Sorry for letting this slide so far down my priority list for how simple it was. (Well, maybe it wasn't that simple since my first rebase made a mess without fixing the conflict, but nothing a force branch reset and force push can't "fix". Hey, it got there in the end.)

djc (approved):

djc:
> Sorry for letting this slide so far down my priority list for how simple it was. (Well, maybe it wasn't that simple since my first rebase made a mess without fixing the conflict, but nothing a force branch reset and force push can't "fix". Hey, it got there in the end.)

Better late than never!

PR #2376, opened 2025-08-30
Remove unaffected from RUSTSEC-2020-0095.md

jayvdb:
There are no releases that are maintained https://crates.io/crates/difference

https://github.com/rustsec/advisory-db/edit/osv/crates/RUSTSEC-2020-0095.json says it is fixed in 2.0.1


djc (approved):

djc:
Thanks!

Issue #2395, opened 2025-09-09
serde_yml is unsound


In-line:
Found in https://github.com/ilya-zlobintsev/LACT/pull/733

It contains UB and author seems to act in malicious intent using AI.

dtolnay:
This bisects to https://github.com/sebastienrousseau/serde_yml/commit/f28516f8c8ff0bd7ba856d5e97cea8817ff98b9f, and within that, to https://github.com/sebastienrousseau/libyml/commit/e9eef0466fd3761bebe8bce1b6d798697a1c7d3a. The UB is not present in serde_yaml / unsafe-libyaml which these crates were forked from.

There is already a different issue https://github.com/rustsec/advisory-db/issues/2212 for a different unsoundness in this crate.

jayvdb:
https://github.com/sebastienrousseau/serde_yml is now archived.
I'll create an advisory for that.

jayvdb:
https://github.com/sebastienrousseau/libyml has also been archived.

bourumir-wyngs:
I have checked the code of unsafe-libyaml and unsafe-libyaml-norway and found not many differences between these. Both contain `(*end).c_offset_from(*start)` that is only defined when both pointers are derived from the same allocation (or one‑past). The problem may persist. I opened a [code review topic](https://users.rust-lang.org/t/review-this-fix-for-unsafe-libyaml-yaml-string-extend/133994) in the Rust community proposing the rewrite.

djc:
Thanks for pushing this forward!

Issue #2212, opened 2025-01-27
Concerns with serde_yaml


evilpie:
@dtolnay raised issues with serde_yaml on twitter/X: https://x.com/davidtolnay/status/1883906113428676938.

cafkafk:
Seems this is specifically the `serde_yml` crate https://github.com/sebastienrousseau/serde_yml, which yea... but I'm not sure this is an actual vulnerability as much as it's just the admittedly frustrating tendency for some projects not to do very thorough vetting of their dependencies.

SkiFire13:
I don't know what can be done about the debatable management of the crate, but at least the `emitter` issue is objectively a soundness issue and can be reported, no?

tarcieri:
Yes, you can file an `informational = "unsound"` advisory for it

jayvdb:
Can be closed due to https://github.com/rustsec/advisory-db/pull/2397

PR #2397, opened 2025-09-11
Add unmaintained libyml and serde_yml

jayvdb:
Closes https://github.com/rustsec/advisory-db/issues/2395 and https://github.com/rustsec/advisory-db/issues/2212

### DIFF
--- a/crates/serde_yml/RUSTSEC-0000-0000.md
+++ b/crates/serde_yml/RUSTSEC-0000-0000.md
@@ -23,7 +23,9 @@ If you rely on this crate, it is highly recommended switching to a maintained al
 - [`serde_norway`](https://crates.io/crates/serde_norway) - Maintained fork of `serde_yaml`, using `unsafe-libyaml-norway`
 - [`serde_yaml_ng`](https://crates.io/crates/serde_yaml_ng) - Maintained fork of `serde_yaml`, using unmaintained `unsafe-libyaml`
 
-## Incomplete alternatives
+## Incomplete pure Rust alternatives
+
+These implementation do not rely on C `libyaml`.
 
 - [`serde_yaml2`](https://crates.io/crates/serde_yaml2)
 - [`yaml-peg`](https://crates.io/crates/yaml-peg)


jayvdb:
@dtolnay , keen to hear if you have any suggestions on how to improve this.

djc:
If these crates actually have vulnerabilities/unsound issues, I guess we should have a stronger advisory for that?

crates/serde_yml/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,26 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "serde_yml"
+date = "2025-09-11"
+url = "https://github.com/rustsec/advisory-db/issues/2395"
+informational = "unmaintained"
+
+[versions]
+patched = []
```
jayvdb: this can be ` = "unsound"` based on https://x.com/davidtolnay/status/1883906113428676938

So we need to add evidence to the advisory?

IMO it should still be `patched = []` , because
- it was AI slop from the moment it was created.  Doesnt make sense to try to identify which of the earlier versions might be ok.
- repo is archived.  its unmaintained status affects all versions

djc: > So we need to add evidence to the advisory?

Yes, I think it would be good to point at specific API that is unsound.

> IMO it should still be `patched = []` , because
> 
> * it was AI slop from the moment it was created.  Doesnt make sense to try to identify which of the earlier versions might be ok.

I'm fine with `patched = []`. I don't want the advisory to be about the fact that the code was AI generated, and instead be more about the low code quality.

> * repo is archived.  its unmaintained status affects all versions

Makes sense, but AIUI not all of the tooling triggers on `informational = "unmaintained"` advisories by default.

crates/libyml/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "libyml"
+date = "2025-09-11"
+url = "https://github.com/rustsec/advisory-db/issues/2395"
+informational = "unmaintained"
```
jayvdb: this can also be "unsound" based on https://github.com/ilya-zlobintsev/LACT/pull/733#issuecomment-3240165183

crates/serde_yml/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
...
+## Recommended alternatives
+
+- [`serde_norway`](https://crates.io/crates/serde_norway) - Maintained fork of `serde_yaml`, using `unsafe-libyaml-norway`
+- [`serde_yaml_ng`](https://crates.io/crates/serde_yaml_ng) - Maintained fork of `serde_yaml`, using unmaintained `unsafe-libyaml`
+
+## Incomplete pure Rust alternatives
+
+These implementation do not rely on C `libyaml`.
+
+- [`serde_yaml2`](https://crates.io/crates/serde_yaml2)
```
jayvdb, line 30: @zim32 , it would be great if you could review https://github.com/zim32/serde_yaml2/pulls to make this more suitable for potential adopters

crates/serde_yml/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
...
+- [`yaml-peg`](https://crates.io/crates/yaml-peg)
```
jayvdb, line 31: fyi @KmolYuan

PR #2410, opened 2025-09-14
mark `martin-mbtiles` as unmaintained due to being renamed to `mbtiles`

CommanderStorm:
I am one of the maintainers of `mbtiles`, so no issue nessesary.


djc (approved):
Thanks!

PR #2412, opened 2025-09-15
fix(fuser): Set correct patched version

FirelightFlagboy:
The patched version was incorrectly set to 1.2.0 where the patch actually landed in 0.16.0


FirelightFlagboy:
Hello @Ltrlg, @djc and @In-line would you review this PR that fix the patched version in the fuser advisory?

djc (approved):

djc:
Thank you for taking care of this!

PR #2421, opened 2025-10-15
Add CVE-2025-62370 `alloy-dyn-abi`

zerosnacks:
As published: https://github.com/alloy-rs/core/security/advisories/GHSA-pgp9-98jm-wwq2

PR #2423, opened 2025-10-15
fix: RUSTSEC-2025-0073 (`alloy-dyn-abi`), update to description and credit

zerosnacks:
If possible I would like update the description to be more accurate. The update to the text was decided after aligning with the reporter.

I have updated the description on: https://github.com/alloy-rs/core/security/advisories/GHSA-pgp9-98jm-wwq2

The previous PR for context: https://github.com/rustsec/advisory-db/pull/2421

I hope this is not too much trouble.


zerosnacks:
Thanks!

PR #2429, opened 2025-10-21
Add advisory for zipora undefined behavior

lewismosciski:
The safe function `fast_prefetch_range` accepts raw pointer without validation, allowing undefined behavior through pointer wraparound.

## Details
- Affects versions < 2.0.1
- **Fixed in v2.0.1** by changing API to accept `&[u8]` instead of raw pointer
- Maintainer confirmed and patched: https://github.com/infinilabs/zipora/issues/10

### DIFF
--- a/crates/zipora/RUSTSEC-0000-0000.md
+++ b/crates/zipora/RUSTSEC-0000-0000.md
@@ -1,0 +1,24 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "zipora"
+date = "2025-10-21"
+url = "https://github.com/infinilabs/zipora/issues/10"
+informational = "unsound"
+categories = ["memory-corruption"]
+keywords = ["undefined-behavior", "soundness"]
+
+[affected.functions]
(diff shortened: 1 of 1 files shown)


djc:
@bindiego are you okay with this advisory being published?

@lewismosciski the RustSec project by policy typically requires maintainers to consent to advisories. If you're going to file lots of these, please bring along the maintainers where possible.

lewismosciski:
Hi @djc, thanks for the heads-up on the policy.

All advisories I filed are for issues that have already been confirmed and patched by the maintainers.

I'm signing off for the day, but I'll contact the maintainers for these pending advisories to get their consent here when I'm back.

One follow-up question: For future reference, what is the policy if a maintainer doesn't respond to a vulnerability report after a reasonable amount of time? Is publishing an advisory still possible then?

djc:
> One follow-up question: For future reference, what is the policy if a maintainer doesn't respond to a vulnerability report after a reasonable amount of time? Is publishing an advisory still possible then?

See https://github.com/rustsec/advisory-db/blob/main/HOWTO_UNMAINTAINED.md.

PR #2431, opened 2025-10-21
Add advisory for remoteprocess soundness issues

lewismosciski:
This PR adds an advisory for soundness issues in remoteprocess.

## Summary
Safe methods in the `ProcessMemory` trait can cause double-free and use-after-free bugs when used with non-Copy types.

## Details
- **Vulnerability**: Methods perform bitwise copies without enforcing Copy trait
- **Impact**: Double-free when copying non-POD types (e.g., String, Vec)
- **Affected functions**: `copy_struct`, `copy_pointer`, `copy_vec`
- **Affected versions**: < 0.5.1
- **Status**: ✅ **Confirmed and fixed by maintainer**
- **Fix**: Added Copy trait bounds to ensure only trivially copyable types can be used
- **Fixed in**: 0.5.1

### DIFF
--- a/crates/remoteprocess/RUSTSEC-0000-0000.md
+++ b/crates/remoteprocess/RUSTSEC-0000-0000.md
@@ -1,0 +1,27 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "remoteprocess"
+date = "2025-10-21"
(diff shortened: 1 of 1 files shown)


lewismosciski:
Hi @benfred,

Could you please confirm you're okay with publishing this advisory for the soundness issue https://github.com/benfred/remoteprocess/issues/105?

Thanks!

PR #2433, opened 2025-10-21
Add advisory for borrowck_sacrifices uninitialized memory

lewismosciski:
This PR adds an advisory for a soundness issue in borrowck_sacrifices.

## Summary
The safe function `any_as_u8_slice` exposes uninitialized memory when used with types containing padding bytes.

## Details
- **Vulnerability**: Creates `&[u8]` including uninitialized padding bytes via `from_raw_parts`
- **Impact**: Violates safety contract of `from_raw_parts`, causing undefined behavior
- **Affected versions**: < 0.2.0
- **Status**: ✅ **Confirmed and fixed by maintainer**
- **Fixed in**: 0.2.0

### DIFF
--- a/crates/borrowck_sacrifices/RUSTSEC-0000-0000.md
+++ b/crates/borrowck_sacrifices/RUSTSEC-0000-0000.md
@@ -1,0 +1,22 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "borrowck_sacrifices"
+date = "2025-10-21"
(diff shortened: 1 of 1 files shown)


lewismosciski:
Hi @alexpyattaev,

Could you please confirm you're okay with publishing this advisory for the soundness issue https://github.com/alexpyattaev/borrowck_sacrifices/issues/1?

Thanks!

alexpyattaev:
Not sure it warrants a RUSTSEC entry but no objections either.

Issue #2414, opened 2025-09-17
rust-unic seems unmaintained


evilpie:
https://github.com/open-i18n/rust-unic contains multiple crates. The last commit has been 5 years ago. Outdated Unicode information causes compatibility issues. (See for example https://github.com/denoland/rust-urlpattern/issues/54)

djc:
@behnam should we mark these crates as unmaintained?

evilpie:
I will prepare a PR.

djc:
I've been able to confirm that these are unmaintained. Are you able to get the ball rolling on a PR?

evilpie:
Actually, unic has _a lot_ different crates, is there some way of having one advisory for multiple crates?

djc:
Nope. But feel free to script something that generates all the mds with almost-identical content.

djc:
(Some discussion in https://github.com/rustsec/advisory-db/pull/2254.)

hsivonen:
Crates whose name are (or start with) `unic-langid` or `unic-locale` come from a [different repo](https://github.com/zbraniecki/unic-locale) that is still maintained (even though the principal author is now working on `icu_locale`), so it probably makes sense not to include those when marking the other crates starting with `unic-` as unmaintained. CC @zbraniecki

evilpie:
We are only talking about crates that come from https://github.com/open-i18n/rust-unic.

psandana:
@evilpie , is there any suggested project to use instead?

djc:
@psandana see the PR linked above. If the advisory does not contain an alternative, we don't know of one.

psandana:
@djc I've seen other RUSTSEC issues that offered the alternatives in the description. That's handy to prevent doing archaeology to find one to move.

djc:
If you'd looked at the PR you'd see that where available, the alternatives are mentioned in the advisory.

As I already said in my previous comment.

psandana:
I see this in the advisory:

<img width="1476" height="1028" alt="Image" src="https://github.com/user-attachments/assets/0dfc6d8a-5443-40fa-b372-bef8dc830653" />

No references to alternatives.

On this example, I see them:
https://rustsec.org/advisories/RUSTSEC-2023-0094.html

<img width="1744" height="1013" alt="Image" src="https://github.com/user-attachments/assets/c652e68b-7091-4136-a949-86ff7fc60a14" />

[...]

PR #2436, opened 2025-10-21
Recommend Rust std alternative to unic-char-range

evilpie:
Part of #2414


djc:
Thanks! I think you have a duplicate "since version" in here.

evilpie:
@djc Sorry about that, I've updated the commit.

Issue #2435, opened 2025-10-21
async-tar and it's forks are vulnerable


lolbinarycat:
see [CVE-2025-62518](https://www.cve.org/CVERecord?id=CVE-2025-62518), [GHSA-j5gw-2vrg-8fgx](https://github.com/astral-sh/tokio-tar/security/advisories/GHSA-j5gw-2vrg-8fgx), and [this announcement blogpost](https://edera.dev/stories/tarmageddon).

This should be added to the rustsec database so library maintainers will be informed about it when running `cargo audit`.

djc:
Want to submit an advisory? cc @woodruffw

woodruffw:
Yeah, it'd be great to have RUSTSEC advisories for these. For context, the CVE is **only** for `astral-tokio-tar` (which is patched), the other forks of `tokio-tar` are potentially not patched or have divergent versions (I'm not familiar with them).

woodruffw:
@lolbinarycat if you're interested in submitting these entries, I'd be happy to review them. Otherwise, I can probably find some time tomorrow to submit these.

lolbinarycat:
I can submit advisories for these, though it will probably take until at least tomorrow.

Should I also submit separate "unmaintained" advisories, or is one advisory per crate enough?

djc:
One advisory per crate is enough, just state explicitly that it is unmaintained and no fixes are expected to be forthcoming (or some such).

Thanks!

jayvdb:
so we have https://github.com/rustsec/advisory-db/pull/2442 and https://github.com/rustsec/advisory-db/pull/2443 merged - what else is needed here?

woodruffw:
I believe krata-tokio-tar also needs an advisory, at least based on Edera's post.

PR #696, opened 2021-01-27
beef: beef::Cow lacks a Sync bound on its Send trait allowing for data races

JOE1994:
Original issue report: https://github.com/maciejhirsz/beef/issues/37

PR #2442, opened 2025-10-23
Add advisory for tokio-tar PAX mis-parsing (CVE-2025-62518)

kpcyrd:
Related to #2435 (doesn't fully fix).

I'm not sure if "format-injection" is the correct category, but there aren't many to pick from for logic bugs.

The advisory can be used for the other crates too (with minor changes), I'm not sure if adding multiple advisories to the same pull request is the correct workflow though.

cc @woodruffw

### DIFF
--- a/crates/tokio-tar/RUSTSEC-0000-0000.md
+++ b/crates/tokio-tar/RUSTSEC-0000-0000.md
@@ -1,0 +1,34 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tokio-tar"
+date = "2025-10-21"
+url = "https://edera.dev/stories/tarmageddon"
+categories = ["format-injection"]
+keywords = ["unsound", "parsing", "smuggling", "file-smuggling", "unmaintained"]
+related = ["CVE-2025-62518", "GHSA-j5gw-2vrg-8fgx"]
+
+[affected.functions]
+"tokio_tar::Archive::new" = ["<= 0.3.1"]
+"tokio_tar::ArchiveBuilder::new" = ["<= 0.3.1"]
+
+[versions]
+patched = []
+```
+
+# `tokio-tar` parses PAX extended headers incorrectly, allows file smuggling
+
+The archive reader incorrectly handles PAX extended headers, when the ustar
+header incorrectly specifies zero size (`size=000000000000`), while a PAX
+header specifies a non-zero size, `tokio-tar::Archive` is going to read the
+file content as tar entry header.
+
+This can be used by a tar file to present different content to `tokio-tar`
+compared to other tar reader implementations.
+
+This bug is also known as `CVE-2025-62518` and `GHSA-j5gw-2vrg-8fgx`, as those
+crates share a common ancestor codebase.
+
+The `tokio-tar` crate is archived and no longer maintained, we recommend you
+switch to an alternative crate such as:
+- [`astral-tokio-tar`](https://crates.io/crates/astral-tokio-tar)


crates/tokio-tar/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tokio-tar"
+date = "2025-10-21"
+url = "https://edera.dev/stories/tarmageddon"
+categories = ["format-injection"]
```
woodruffw, line 7: Yeah, this seems a bit off (but it also seems like the closest of the existing categories)...

If we had a new one, I think `parser-differential` or even `parser-confusion` or something like that would be appropriate.

crates/tokio-tar/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
...
+keywords = ["unsound", "parsing", "smuggling", "file-smuggling", "unmaintained"]
+aliases = ["CVE-2025-62518", "GHSA-j5gw-2vrg-8fgx"]
```
woodruffw: Someone who knows RUSTSEC's practices better than me can confirm, but does it make sense to alias these? From what I can tell this advisory will be for `tokio-tar`, but the aliases are for `astral-tokio-tar`.

djc: I don't think we should have "aliases" for advisories that cover a different package, even if it is a fork.

tarcieri: If it's for a similar but not identical advisory, use `related`

crates/tokio-tar/RUSTSEC-0000-0000.md
kpcyrd: I've changed the categories to `logic-bug` and `parser-confusion`, since those are probably good categories to have in general.

There was one pull request that introduced a new category (`thread-safety`), based on the comments it seems this requires changes in `rustsec-admin`: https://github.com/rustsec/advisory-db/pull/696

crates/tokio-tar/RUSTSEC-0000-0000.md
kpcyrd: I've removed `aliases =`, and instead moved it to the description text (as I think this is still useful information, but let me know if you'd rather change/remove this).

crates/tokio-tar/RUSTSEC-0000-0000.md
kpcyrd: Looking into the `rustsec-admin` codebase, `format-injection` is defined as:
```rust
/// Mishandled escaping allowing an attacker to execute code or perform
/// otherwise unexpected operations, e.g. shell escaping, SQL injection, XSS.
FormatInjection,
```

tarcieri: I would suggest proposing new categories separate from trying to use them in advisories. They aren't freeform and will block merging the advisory (see failed validation).

crates/tokio-tar/RUSTSEC-0000-0000.md
kpcyrd: I've added them back in with `related =`, thanks for the pointer

crates/tokio-tar/RUSTSEC-0000-0000.md
tarcieri: FWIW I don't think format injection fits here. It's more of an input validation bug IMO

kpcyrd: I've opened a pull request to introduce additional categories: https://github.com/rustsec/rustsec/pull/1448

djc:
Would be nice to stick with existing categories for this PR, so we can get it merged. @woodruffw can you do one for Astral's fork?

woodruffw:
Sure, I'll have one up for astral-tokio-tar in a bit 🙂

woodruffw:
Opened https://github.com/rustsec/advisory-db/pull/2443 for the astral-tokio-tar advisory.

kpcyrd:
I've changed it back to `format-injection` since that's also what #2443 is using

djc:
Thanks!

PR #2443, opened 2025-10-24
Add advisory for astral-tokio-tar (CVE-2025-62518)

woodruffw:
This records CVE-2025-62518 as a RUSTSEC advisory as well.

See also #2442.

PR #2444, opened 2025-10-24
ci: ratchet down permissions, pin all actions

woodruffw:
[zizmor](https://docs.zizmor.sh) flagged these while I was doing #2443 🙂 

This fixes a handful of mostly minor security issues in the workflows/Dependabot configurations used in this repo. Namely:

* All actions are now fully hash-pinned, to make runs more hermetic/make it impossible for an external tag overwrite to modify CI behavior here.
* All `actions/checkout` usage now either explicitly disables credential persistence *or* explicitly enables it if the job previously implicitly required it (e.g. for `git push`).
* All workflows now drop their default permissions to the empty set (`permissions: {}`), and only give each internal job the specific permissions its needs.
* Dependabot now includes a `cooldown` on its updates (of 7 days), making it harder for an opportunistic compromise of an action dependency to impact this repo (since it'd need to survive in the wild for longer).

Hopefully these changes are welcome! If not, please feel free to close this.

### DIFF
--- a/.github/workflows/assign-ids.yml
+++ b/.github/workflows/assign-ids.yml
@@ -4,16 +4,22 @@ on:
   push:
     branches: main
 
+permissions: {}
+
 jobs:
   assign-ids:
     name: Assign IDs
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write # needed to create pull requests
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
 
       - name: Cache cargo bin
         id: admin-cache
-        uses: actions/cache@v4
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
           path: ~/.cargo/bin
           key: rustsec-admin-c7f56c474e01619b78b9c39bdb626d982f3bee90
@@ -35,7 +41,7 @@ jobs:
           ls -R ./crates/ ./rust/ | sha256sum >> .duplicate-id-guard
 
       - name: Create pull request
-        uses: peter-evans/create-pull-request@v7
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           commit-message: ${{ steps.assign.outputs.commit_message }}
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,5 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 10
+    cooldown:
+      default-days: 7
--- a/.github/workflows/export-osv.yml
+++ b/.github/workflows/export-osv.yml
@@ -4,15 +4,20 @@ on:
   push:
     branches: main
 
+permissions: {}
+
 jobs:
   publish-web:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write # needed for pushing back to the repo
     steps:
-      - uses: actions/checkout@v5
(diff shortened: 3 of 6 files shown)


tarcieri:
@woodruffw these changes are welcome but https://github.com/rustsec/rustsec is probably the more important repo as it contains the actual source code, whereas this repo is largely a bunch of TOML files

woodruffw:
> @woodruffw these changes are welcome but [rustsec/rustsec](https://github.com/rustsec/rustsec?rgh-link-date=2025-10-24T16%3A06%3A08Z) is probably the more important repo as it contains the actual source code, whereas this repo is largely a bunch of TOML files

Cool, I can follow up there as well!

.github/workflows/assign-ids.yml
```diff
@@ -4,16 +4,22 @@ on:
...
+
 jobs:
   assign-ids:
     name: Assign IDs
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write # needed to create pull requests
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```
djc, line 16: I played with zizmor in https://github.com/quinn-rs/quinn/pull/2285 and I'm not sure about these Actions hashes. Generally in the non-attack case it seems better to fairly regularly upgrade, and this setup seems to make that substantially harder.

oherrala: @djc Dependabot IMO can help update those hashes when it includes the version in comment as above.

djc: > @djc Dependabot IMO can help update those hashes when it includes the version in comment as above.

Okay, if that works it seems okay. But IIRC zizmor suggests not using that because of the supply chain risks...

woodruffw: > Okay, if that works it seems okay. But IIRC zizmor suggests not using that because of the supply chain risks...

Unless someone has modified the docs without telling me, zizmor doesn't recommend *against* Dependabot (or any other automatic updating bot) 🙂 

There are some audits that detect common misconfigurations of Dependabot, but I would definitely recommend using it *with* cooldown to keep actions and other dependencies updated.

woodruffw: Also note: updating these kinds of pinned actions is easier than ever locally too, thanks to tools like `pinact` and `gha-update`! Some recommendations on those here:

https://docs.zizmor.sh/audits/#remediation_25

djc: I saw those, but didn't find any of them very appealing. Anyway, I think relying on Dependabot is fine for now.

djc:
It looks like this has broken the ID assigner:

```
Pushing pull request branch to 'origin/assign-ids'
  /usr/bin/git push --force-with-lease origin assign-ids:refs/heads/assign-ids
  remote: Permission to rustsec/advisory-db.git denied to github-actions[bot].
  fatal: unable to access 'https://github.com/rustsec/advisory-db/': The requested URL returned error: 403
  Error: The process '/usr/bin/git' failed with exit code 128
```

woodruffw:
> It looks like this has broken the ID assigner:
> 
> 
> 
> ```
> 
> Pushing pull request branch to 'origin/assign-ids'
> 
>   /usr/bin/git push --force-with-lease origin assign-ids:refs/heads/assign-ids
> 
>   remote: Permission to rustsec/advisory-db.git denied to github-actions[bot].
> 
>   fatal: unable to access 'https://github.com/rustsec/advisory-db/': The requested URL returned error: 403
> 
>   Error: The process '/usr/bin/git' failed with exit code 128
> 
> ```

Hmm, it probably needs a combination of `persist-credentials: true` and the `contents: write` permission within that job. Sorry about that, GHA makes it hard to see these kinds of permission dependencies until they actually fail.

I can send a patch for that in a moment.

woodruffw:
https://github.com/rustsec/advisory-db/pull/2448 has the fix.

PR #2448, opened 2025-10-26
ci: fix create-pull-request permissions

woodruffw:
This fixes two jobs that use the `create-pull-request` action. This action requires both `contents: write` and `pull-requests: write`, since it creates and pushes a branch before opening the PR. 

Documented here: https://github.com/marketplace/actions/create-pull-request#token


djc (approved):
Thanks!

PR #2449, opened 2025-10-26
ci: fix typo in permissions key

djc:
This went wrong in

- #2448 

but apparently GitHub doesn't complain if the workflow isn't triggered or something?

cc @woodruffw


alex (approved):

woodruffw:
Yeah, that's bad behavior on GitHub's part: invalid workflows are silently skipped instead of causing a loud error. Sorry for the typo there...

PR #2457, opened 2025-11-13
wasmtime: Unsound API access to a WebAssembly shared linear memory

alexcrichton:
Republishing https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-hc7m-r6v8-hg9q here


djc (approved):
(FWIW, we should have support for CVSS 4 now.)

Issue #2214, opened 2025-01-29
Unmaintained and unsound crate: lockfree


VorpalBlade:
https://crates.io/crates/lockfree is unmaintained and also unsound (using std::mem::uninitialized). I have tried reaching out to the author of the crate on their gitlab and on github with no success.

It has been much more than the required 90 days. I would like to start the process for issuing a rustsec advisory. 

There is a fork from 2023, that I can in no way wouch for: https://crates.io/crates/lock_freedom

References:

* Unsound: https://gitlab.com/bzim/lockfree/-/issues/4
* Unmaintained:
   * https://gitlab.com/bzim/lockfree/-/issues/6
  * https://github.com/brunoczim/lockfree/issues/4

djc:
Would you be able to submit a PR?

PR #2461, opened 2025-11-16
Add lockfree unsound & unmaintained

jayvdb:
Closes https://github.com/rustsec/advisory-db/issues/2214

### DIFF
--- a/crates/lockfree/RUSTSEC-0000-0000.md
+++ b/crates/lockfree/RUSTSEC-0000-0000.md
@@ -1,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "lockfree"
+date = "2025-11-17"
+url = "https://gitlab.com/bzim/lockfree/-/issues/4"
+categories = ["memory-corruption"]
+keywords = ["memory-safety", "std::mem::uninitialized"]
+informational = "unsound"
+
+[versions]
+patched = []
+unaffected = ["<= 0.4.0"]
+```
+
+# lockfree crate is unsound and unmaintained
+
+Affected versions of this crate called `mem::uninitialized()` to construct `Removable<T>`.
+This is unsound.
+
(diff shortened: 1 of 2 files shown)


crates/lockfree/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
...
+# lockfree crate is unsound and unmaintained
+
+Affected versions of this crate called `mem::uninitialized()` to construct `Removable<T>`.
+This is unsound.
+
+In addition, the `lockfree` crate is no longer actively maintained.  If you rely on this crate, consider switching to a recommended alternative.
+
+## Recommended alternatives
+
+- [`lock_freedom`](https://crates.io/crates/lock_freedom)
```
djc, line 25: I don't think this is active/trustworthy enough to merit a recommendation.

jayvdb: so should I look for others, or just not recommend any?

djc: Just don't recommend any -- users can look for themselves.

jayvdb: Maintainer of lock_freedom is giving it a bit of love.

Also I have found https://github.com/DiscordJim/lfqueue and https://github.com/emilHof/sling and https://github.com/emilHof/unlink as other narrower alternatives that users may find useful if they dont need the full suite of features of `lockfree`.

djc: None of these are in widespread usage or by maintainers that have other popular crates.

jayvdb: lock_freedom now has 4 dependents at https://github.com/wyatt-herkamp/lock_freedom/network/dependents , and one at https://crates.io/crates/lock_freedom/reverse_dependencies

djc: I still don't think it's enough. Just remove the recommendations, then we can merge this.

Issue #2211, opened 2025-01-25
`rand_os` was deprecated 5.5 years ago - crates.io still reports many daily downloads


paolobarbolini:
As explained by the [readme, `rand_os`](https://crates.io/crates/rand_os) was deprecated 5.5 years ago. The high number of downloads on crates.io shows many private projects (reverse dependencies show a very low number of downloads at https://lib.rs/crates/rand_os/rev) may still be using it. Would it make sense to have an advisory about this?

tarcieri:
Yes that sounds like a good idea cc @dhardy

dhardy:
The crate was deprecated because functionality was rolled into `rand_core` in v0.6. The fact that people are still using this indicates that they are using an old version of `rand_core` (and presumably `rand`).

According to the download stats, the vast majority are for `0.1.3` which isn't even the latest version of `rand_os` (`0.2.2` using `rand_core` v0.4). Why exactly this is I don't know but the top threereverse dependencies (by downloads) appear to be related to the same project, which (according to `cardano-serialization-lib` source) is still using v0.1 of `rand_os`. It may therefore be a single high-profile project responsible for the vast majority of these downloads.

Whether or not you decide to issue an advisory because of high usage of an old crate version I'll leave up to you.

CC @newpavlov @josephlr

newpavlov:
I think we can release an empty `rand_os` v0.3 with a deprecation notification pointing users to `rand_core::OsRng`. `rand_core` v0.2 then would be shown as an outdated dependency (e.g. during `cargo update`), so direct users would have a chance to learn about the crate deprecation.

I don't think `rand_os` has any security issues per se.

jayvdb:
Created https://github.com/rustsec/advisory-db/pull/2462

PR #2462, opened 2025-11-16
Mark rand_os as unmaintained

jayvdb:
Closes https://github.com/rustsec/advisory-db/issues/2211

### DIFF
--- a/crates/rand_os/RUSTSEC-0000-0000.md
+++ b/crates/rand_os/RUSTSEC-0000-0000.md
@@ -15,4 +15,4 @@ patched = []
 The `rand_os` crate is deprecated and no longer actively maintained, as `OsRng`
 is now part of `rand_core`.
 
-If you are using this crate, consider upgrade to `rand_core` 0.5.1 or higher.
+If you are using this crate, consider upgrading to `rand_core` 0.5.1 or higher.


dhardy:
Can advisories target specific versions?

There's a chance we re-launch `rand_os` for `rand_core` v0.10.

paolobarbolini:
> Can advisories target specific versions?
> 
> There's a chance we re-launch `rand_os` for `rand_core` v0.10.

In that case it wouldn't make sense to have an _unmaintained_ advisory.

jayvdb:
Advisories can be rescinded.

jayvdb:
IMO it is all the more important to get this advisory out, so current users of `rand_os` upgrade before a new `rand_os` appears.

Or will the new `rand_os` have the same API as the old one, this providing a viable upgrade path for users of the old `rand_os`?

dhardy:
> _unmaintained_ advisory

Take a look at the [download stats](https://crates.io/crates/rand_os); the high usage doesn't even apply to the latest version (0.2.x). So whether or not we release a new version is entirely irrelevant to the high download rate of 0.1.x.

> Or will the new `rand_os` have the same API as the old one, this providing a viable upgrade path for users of the old `rand_os`?

It would require the latest `rand_core` so be incompatible.

jayvdb:
Right, an advisory will assist in reducing usage of 0.1.x as well as 0.2.x.

crates/rand_os/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,18 @@
...
+[versions]
+patched = []
+```
+
+# and_os crate is unmaintained
+
+The `rand_os` crate is deprecated and no longer actively maintained, as `OsRng`
+is now part of `rand_core`.
+
+If you are using this crate, consider upgrade to `rand_core` 0.5.1 or higher.
```
djc: Nit: "consider upgrading".

Issue #1935, opened 2024-04-10
cargo-asm might be unmaintained


oherrala:
* Last release on crates.io published 2018 (https://crates.io/crates/cargo-asm)
 * Last commit on repository on 2019 (https://github.com/gnzlbg/cargo-asm)
 * Maintenance status unanswered since 2022 (https://github.com/gnzlbg/cargo-asm/issues/244)
 * Maintainer has been @gnzlbg and he also has `slice_deque` crate marked as unmaintained (https://rustsec.org/advisories/RUSTSEC-2020-0158.html)
 
There seems to be maintained alternative: https://crates.io/crates/cargo-show-asm

Shnatsel:
It is a binary crate, not used as a dependency by any other crates, so if we file an advisory if will not be surfaced through most tools. Although perhaps lib.rs will show it?

PR #2467, opened 2025-11-17
Mark cargo-asm unmaintained

jayvdb:
Closes https://github.com/rustsec/advisory-db/issues/1935


djc:
@Shnatsel in https://github.com/rustsec/advisory-db/issues/1935#issuecomment-2048272069:

> It is a binary crate, not used as a dependency by any other crates, so if we file an advisory if will not be surfaced through most tools. Although perhaps lib.rs will show it?

Not sure if this is worth it.

jayvdb:
It is definitely worth it.

osv-scanner finds stuff like this on docker images.

PR #2473, opened 2025-11-18
Add maintained fork as alternative for daemonize

decathorpe:
I have forked the original `daemonize` crate as `daemonix` and merged changes that fix the UB issue that was reported against the original crate. It can be used as a "safer" / "maintained" drop-in replacement for daemonize.


djc:
I'm not convinced we want to recommend brand new forks as official alternatives. Ideally they should be easy to find when searching crates.io for relevant keywords (including potentially the old crate name), though.

decathorpe:
I understand the hesitation. Though that's kind of a chicken-and-egg problem ... I'm fine with this PR being left unmerged for now, until there's either a better alternative or more projects have adopted the new crate.

PR #2479, opened 2025-11-22
Use CVSS 3.1 for RUSTSEC-2025-0125

jayvdb:
c.f. https://github.com/EmbarkStudios/cargo-deny/issues/804

Fixes RUSTSEC raised in https://github.com/rustsec/advisory-db/pull/2476 by crate maintainer, and assigned by GHA in https://github.com/rustsec/advisory-db/pull/2477

### DIFF
--- a/crates/thread-amount/RUSTSEC-2025-0125.md
+++ b/crates/thread-amount/RUSTSEC-2025-0125.md
@@ -7,7 +7,7 @@ url = "https://github.com/jzeuzs/thread-amount/pull/29"
 categories = ["denial-of-service"]
 keywords = ["memory-leak", "handle-leak", "resource-exhaustion", "windows", "macos"]
 aliases = ["GHSA-jf9p-2fv9-2jp2", "CVE-2025-65947"]
-cvss = "CVSS:4.0/AV:N/AC:L/AT:N/PR:N/UI:N/VC:N/VI:N/VA:H/SC:N/SI:N/SA:N/E:P/AU:Y/R:U/V:D/RE:L/U:Green"
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
 
 [affected]
 os = ["windows", "macos", "ios"]


jayvdb:
ping @djc

djc:
@jzeuzs does this look okay to you?

Skgland:
It looks like[^1] there are two[^2][^3] other advisories that would like to use CVSS:4.0 though they currently have it commented out.
Would it make sense to do the same here instead of converting it to CVSS:3.1?
Also given that it looks like this just requires a dependency update in cargo-deny[^4] and there is already a PR EmbarkStudios/cargo-deny#805 for that, would it make sense to not do anything here and just have everyone update cargo-deny once that lands?


[^1]: https://github.com/search?q=repo%3Arustsec%2Fadvisory-db+%22CVSS%3A4.0%22&type=code
[^2]: https://github.com/rustsec/advisory-db/blob/594ee9d61e0adc739bf65133e83b79c4c577361a/crates/pingora-core/RUSTSEC-2025-0070.md?plain=1#L11
[^3]: https://github.com/rustsec/advisory-db/blob/594ee9d61e0adc739bf65133e83b79c4c577361a/crates/pingora-core/RUSTSEC-2025-0037.md?plain=1#L9
[^4]: https://github.com/EmbarkStudios/cargo-deny/issues/804#issuecomment-3566434046

djc:
> Also given that it looks like this just requires a dependency update in cargo-deny4 and there is already a PR EmbarkStudios/cargo-deny#805 for that, would it make sense to not do anything here and just have everyone update cargo-deny once that lands?

Potentially, but failing cargo-deny runs for everyone is a pretty bad failure mode, so if we can make this a temporary workaround for a bit that seems like an improvement.

jayvdb:
Why do we need someones opinion about this PR?

It is a technical conversion only.

jzeuzs:
> @jzeuzs does this look okay to you?

Hi! Yes, it looks good to me. Thanks.

PR #2483, opened 2025-11-25
Add vulnerability advisory for nftnl

cuiwenhao123:
A heap-buffer-overflow vulnerability exists in the Rust wrapper for libnftnl, triggered via the nftnl::Batch::with_page_size constructor. When a small or malformed page size is provided, the underlying C code allocates an insufficient buffer, leading to out-of-bounds writes during batch initialization. 

Issue is [here](https://github.com/mullvad/nftnl-rs/issues/76). 
The fix information can be found [here](https://github.com/mullvad/nftnl-rs/pull/82).


djc:
Thanks!

PR #2488, opened 2025-11-26
Update RUSTSEC-2025-0126 with a patched version / mitigation suggestion

MarkusPettersson98:
Add patched version number to RUSTSEC-2025-0126 - a fix for the CVE was released in `nftnl 0.9.0`: https://github.com/mullvad/nftnl-rs/blob/v0.9.0/CHANGELOG.md#fixed.

### DIFF
--- a/crates/nftnl/RUSTSEC-2025-0126.md
+++ b/crates/nftnl/RUSTSEC-2025-0126.md
@@ -8,7 +8,7 @@ categories = ["memory-corruption"]
 aliases = ["GHSA-2fjw-whxm-9v4q"]
 
 [versions]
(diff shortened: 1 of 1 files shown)


djc:
@faern r?

faern:
My understanding is that the vulnerability referred to in [RUSTSEC-2025-0126](https://rustsec.org/advisories/RUSTSEC-2025-0126.html) is indeed fixed only in `0.9.0`, not in `0.8.0` as the advisory metadata originally suggested. So this update seems correct!

faern (approved):

amousset:
Thanks!

PR #2491, opened 2025-11-28
Add maxminddb mmap unsoundness advisory

paolobarbolini:
Adds the advisory for https://github.com/oschwald/maxminddb-rust/issues/86.
Cc @oschwald


oschwald:
Thanks for asking. That is fine.

While looking into this, I did notice that many of the most popular dependents of `memmap2` also have this same concern or more significant concerns where code actually writes to the map. That said, the claim that it is unsound as an external process could modify the file is valid.

Issue #2500, opened 2025-12-05
Category for malicious information stealing?


carols10cents:
I'm working on filing advisories for crates we've recently removed from crates.io. These crates were intentionally malicious and attempting to exfiltrate information, including cryptocurrency private keys and API tokens. Should there be a category (or categories) for this?

djc:
Seems okay to me. Not sure whether it should be more specific (`exfiltration`) or wider (`malicious`)?

jlizen:
I would appreciate a malicious takedown category, personally, even if there is an even more granular category for exfiltration. It would also be great to have an advisory severity set even if there is no cve is opened.

My use case is that, as an private crates.io mirror administrator, I want to immediately yank crates that have malicious takedowns in an event driven workflow based on the rustsec feed. 

For severity, I see many consumers snoozing informational RUSTSEC advisories due to not wanting to fail CI on eg unmaintained warnings. I imagine most folks would want to bubble up malicious takedowns to require CI interventions. (Yes, they could also filter on justification even if informational/unknown severity, but that requires tooling support, nice to have defense in depth).

GrassedgeT:
Support, recently there have been more reports on the removal of malicious software packages. I hope this malicious tag can be implemented soon, otherwise it will be difficult for software supply chain security researchers to filter out reports on malicious software packages.

djc:
It's still not obvious to me what the right name or granularity for this is.

In the meantime, seems like using the `expect-deleted` field could help with this signalling?

GrassedgeT:
I think `malicious`is sufficient enough, as it primarily reflects its malicious nature, rather than just stating that it should be deleted

tarcieri:
`malicious` or `malware` seem fine to me

jlizen:
`malicious` is fine for me

I would possibly consume an `expect-deleted` field as well, but it is less likely to be piped through managed CVE feed services by default unlike the justification field

LawnGnome:
`malicious` has been added as a category, and I've backfilled the relevant advisories with that category.

For those consuming OSV data: this will be available in `.affected[].database_specific.categories[]` as `malicious`.

We may want to consider distinguishing malware takedowns more obviously on the advisory list page on rustsec.org, but this should address this specific need.

Issue #2499, opened 2025-12-05
Descriptions of categories?


carols10cents:
I'm trying to write up some advisories and I'm not sure what the categories mean 😕 It'd be nice to have a description somewhere, along with some examples of what would fit in each category and what wouldn't?

For example, a crate that's maliciously attempting to steal cryptocurrency, does that go in the category "crypto-failure"? Or is that category crypto as in cryptography?

tarcieri:
I think cases like this could probably use a new category like `malware`. Failing that, `code-execution` probably covers it.

Regarding a list and description of available categories, the rustdoc is the main source of documentation for now:

https://docs.rs/rustsec/latest/rustsec/advisory/enum.Category.html

PR #2501, opened 2025-12-05
Advisories for evm-units and uniswap-utils

carols10cents:
My understanding is this is going to fail the linter because we've already taken down these crates from crates.io: https://github.com/rustsec/rustsec/issues/1479

And I didn't add categories because of https://github.com/rustsec/advisory-db/issues/2499 and  https://github.com/rustsec/advisory-db/issues/2500

### DIFF
--- a/crates/uniswap-utils/RUSTSEC-0000-0000.md
+++ b/crates/uniswap-utils/RUSTSEC-0000-0000.md
@@ -1,7 +1,7 @@
 ```toml
 [advisory]
 id = "RUSTSEC-0000-0000"
-package = "evm-units"
+package = "uniswap-utils"
 date = "2025-12-03"
 url = "https://blog.rust-lang.org/2025/12/03/crates.io-malicious-crates-evm-units-and-uniswap-utils/"
 references = ["https://socket.dev/blog/malicious-rust-crate-evm-units-serves-cross-platform-payloads"]


tarcieri:
They probably need at least a placeholder version. `>= 0` should work.

carols10cents:
Looks like a bunch of crates have this:

```
[versions]
patched = []
```

crates/uniswap-utils/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,15 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "evm-units"
```
tarcieri:
```suggestion
package = "uniswap-utils"
```

carols10cents:
Ok _now_ i think this is at the "crate doesn't exist in crates.io" lint failure?

```
error: error linting advisory DB .: crates.io index error: crates.io package name does not match package name in advisory for uniswap-utils in RUSTSEC-0000-0000
```

tarcieri:
Yep, looks like we need https://github.com/rustsec/rustsec/issues/1479 to move forward now

djc:
I think we're ready for this now, see

- https://github.com/rustsec/advisory-db/pull/2508

carols10cents:
@djc Whoops, didn't realize yinz were waiting on me here! I've rebased and added the expect-deleted flag, so I think this is ready to merge.

PR #2504, opened 2025-12-05
Add advisories for finch-rust and sha-rust

carols10cents:


djc:
I think we're ready for this now, see

- https://github.com/rustsec/advisory-db/pull/2508

carols10cents:
@djc Whoops, didn't realize yinz were waiting on me here! I've rebased and added the expect-deleted flag, so I think this is ready to merge.

djc (approved):

djc:
Thanks!

PR #2508, opened 2025-12-08
Set expect-deleted flag for rustdecimal and vec-const

djc:
See

- https://github.com/rustsec/rustsec/pull/1480

Are there any caveats/risks to adding new metadata fields?


djc:
@tarcieri ping?

tarcieri (approved):

PR #2525, opened 2025-12-19
fix: comment out CVSS with unsupported version for RUSTSEC-2024-0445

carodewig:
CVSS v4 isn't supported yet, so this causes parsing errors when running `cargo deny`. I commented out the line for parity with other advisories ([examples](https://github.com/search?q=repo%3Arustsec%2Fadvisory-db%20%22CVSS%3A4%22&type=code)).

Seems to be the same issue as #2477


carodewig:
Updating to cargo-deny 0.18.9 fixed this

PR #2527, opened 2025-12-20
obfstr: add information about CVE-2024-58253

alexanderkjall:
Information taken from: https://nvd.nist.gov/vuln/detail/CVE-2024-58253

@CasualX Is it ok if this is added to rustsec?

### DIFF
--- a/crates/obfstr/RUSTSEC-0000-0000.md
+++ b/crates/obfstr/RUSTSEC-0000-0000.md
@@ -1,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "obfstr"
+date = "2024-10-04"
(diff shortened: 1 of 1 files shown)


CasualX:
Sure? If you think it has value.

Let's be clear here: With this bug you can exploit your own source code to trigger UB (invalid conversion of [u8] to str). Debug assertions perform a check to catch bugs. You cannot use this macro on runtime/user data, it must be a compile time constant.

Not exactly going to hack the pentagon with this one...

djc:
Please drop the CVE URL and use the first-party issue as the primary URL.

alexanderkjall:
@djc done, sorry that it has taken time

djc:
I think this should be `informational = "unsound"` to reduce noise for most downstream users.

PR #2529, opened 2025-12-20
coreos-installer: add information about CVE-2021-20319

alexanderkjall:
information taken from: https://bugzilla.redhat.com/show_bug.cgi?id=2011862

@jlebon ok if this is published in rustsec?


jlebon:
Yes.

djc:
FWIW @alexanderkjall the CVEs mentioned as an alias get a link, so I think adding them as a reference as well is superfluous.

PR #2530, opened 2025-12-21
aes-gcm: Add information about CVE-2023-42811

alexanderkjall:
Information taken from: https://github.com/RustCrypto/AEADs/security/advisories/GHSA-423w-p2w9-r7vq

Note that this have a slightly different version range than the github advisory, and marks version 0.10.0 as vulnerable also.

Based on running the PoC against 0.10.0 and also reading the code here:

https://docs.rs/aes-gcm/0.10.0/src/aes_gcm/lib.rs.html#247
https://docs.rs/aes-gcm/0.10.2/src/aes_gcm/lib.rs.html#287-311
https://docs.rs/aes-gcm/0.10.3/src/aes_gcm/lib.rs.html#287-311

I think that 0.10.0 is also vulnerable.

@tarcieri is it ok if this gets published in rustsec?

### DIFF
--- a/crates/aes-gcm/RUSTSEC-0000-0000.md
+++ b/crates/aes-gcm/RUSTSEC-0000-0000.md
@@ -1,0 +1,82 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "aes-gcm"
+date = "2023-11-22"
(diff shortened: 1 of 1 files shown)


tarcieri:
Yes that's fine. I'm a little surprised one for this didn't actually get filed already.

PR #2532, opened 2025-12-21
starship: Add information about CVE-2024-41815

alexanderkjall:
information taken from: https://github.com/starship/starship/security/advisories/GHSA-vx24-x4mv-vwr5

@davidkna Ok if this gets added to rustsec?


davidkna:
I don't mind adding this here, though starship no longer has any active library users that I know of.

djc:
@alexanderkjall please drop the superfluous CVE reference.

alexanderkjall:
removed, thanks for the review :)

PR #2538, opened 2025-12-22
Report unsoundness for ruint

shinmao:
The safe API provided in the crate can trigger out-of-bounds, which is undefined behavior.

### DIFF
--- a/crates/ruint/RUSTSEC-0000-0000.md
+++ b/crates/ruint/RUSTSEC-0000-0000.md
@@ -1,7 +1,7 @@
 ```toml
 [advisory]
 id = "RUSTSEC-0000-0000"
-package = "uint"
+package = "ruint"
 date = "2025-12-22"
 url = "https://github.com/recmo/uint/issues/550"
 categories = ["memory-corruption"]


djc:
@recmo @prestwich @gakonst are you okay with having an advisory published for this?

prestwich:
`reciprocal_mg10` is explicitly not considered part of the stable API of the crate, so it seems a bit unnecessary. Is there an existing policy for this?

djc:
> `reciprocal_mg10` is explicitly not considered part of the stable API of the crate, so it seems a bit unnecessary. Is there an existing policy for this?

I can see the warning. Without having used this crate before, I don't consider it a very strong case against an advisory, because:

- I would interpret it not being "part of the stable API" to mean that there could be semver-incompatible changes in otherwise semver-compatible version bumps (that is, with the emphasis on **stable**), rather than a warning that the function is unsound (in the sense of requiring the caller to uphold particular invariants that are not enforced by the API).
- The warning gives no indication why/what part of the function makes it dangerous or unstable.

If there are additional invariants to uphold, the function should ideally just be marked `unsafe`. Alternatively, I think you'd have a stronger case if it were marked as `#[doc(hidden)]` or otherwise explicitly documented the potential issues from using it.

prestwich:
sounds good. I have no issue with an advisory being published, thanks

djc:
> sounds good. I have no issue with an advisory being published, thanks

Do you think a fix/mitigation will be published soon? If so, we'd like to wait for that.

prestwich:
I consider this very low risk, and I'd like to review the code for other instances before publishing a patch. I don't think this is a high priority, so I'm unlikely to publish a new version in the next few days

kayabaNerve:
While a patch isn't currently available, a mitigation would presumably be to extend the workspace's `Cargo.toml` with:

```toml
[profile.dev.package.ruint]
debug-assertions = true
[profile.release.package.ruint]
debug-assertions = true
[profile.test.package.ruint]
debug-assertions = true
[profile.bench.package.ruint]
debug-assertions = true
```

though I'm unsure how well this plays with CLI flags explicitly disabling `debug-assertions` for a build, when this mitigation would be expected to be _always on_. I'd expect the CLI to override it, making this incomplete.

gakonst:
ACK - defer to James here

prestwich:
This was addressed by publishing v1.17.1. Any action I should take to update the advisory docs?

shinmao:
@prestwich thanks for your updates. @djc do I need to re-send PR to update the patched version, or you can directly update it?

djc:
Would be great if someone can send a PR that updated `versions.patched` in the existing advisory.

shinmao:
> Would be great if someone can send a PR that updated `versions.patched` in the existing advisory.

Ok. I can do it today.

PR #2539, opened 2025-12-22
Report unsoundness for lmdb-rs

shinmao:
The safe API, which has been reported with undefined behavior, can trigger UB again. It requires a more systematic fix on satisfying safety invariants.

### DIFF
--- a/crates/lmdb-rs/RUSTSEC-0000-0000.md
+++ b/crates/lmdb-rs/RUSTSEC-0000-0000.md
@@ -1,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "lmdb-rs"
+date = "2025-12-22"
+informational = "unsound"
+url = "https://github.com/vhbit/lmdb-rs/issues/69"
+keywords = ["unsound"]
+references = ["https://github.com/vhbit/lmdb-rs/issues/67"]
+
+[affected.functions]
+"lmdb_rs::FromMdbValue::from_mdb_value" = [">= 0.0.0"]
+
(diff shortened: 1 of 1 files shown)


djc:
@vhbit are you okay with publishing an advisory for this?

@shinmao if you're going to file more advisories, please seek approval from maintainers.

shinmao:
Hi @djc , thanks for the reminder. As I previously reported, the grace period would be 2 weeks after reporting issues in repo. If still no response, then we can go ahead to send the advisory. Isn't this the policy? Just for confirmation.

djc:
> Hi @djc , thanks for the reminder. As I previously reported, the grace period would be 2 weeks after reporting issues in repo. If still no response, then we can go ahead to send the advisory. Isn't this the policy? Just for confirmation.

It's more of a guideline than a policy, and it's not completely obvious that just pinging on GitHub is sufficient.

shinmao:
> It's more of a guideline than a policy, and it's not completely obvious that just pinging on GitHub is sufficient.

Understood. I think I should also ask the approval for advisories rather than just publising the issue. Thanks!

tarcieri:
@shinmao we have no 2 week grace periods. The industry has generally settled on a 90 day window, but that's typically for private vulnerability disclosures

shinmao:
@tarcieri gotcha! I will wait for the response from the maintainers.

djc:
The closest we get is this:

> Unresponsiveness by the author over a period of 270 days (or 60 days of unresponsiveness after being notified of a vulnerability) is the minimum before a crate will be considered unmaintained.

PR #2540, opened 2025-12-24
pgp: Add information about CVE-2024-53856

alexanderkjall:
information taken from: https://github.com/rpgp/rpgp/security/advisories/GHSA-9rmp-2568-59rv

@dignifiedquire / @hko-s Ok if this gets published in rustsec?


hko-s:
Sounds good, thanks for documenting this in rustsec!

PR #2543, opened 2025-12-24
Add advisory for capnp unsoundness

shinmao:
The unsoundness issue has been verified and patched in the latest version of crate: https://github.com/capnproto/capnproto-rust/issues/605

I would like to confirm for the maintainer that whether the advisory will break any normal use cases. 

From my understanding, this will only alarm in the `cargo audit` so it might not be bothering. Also, the users of the crate should still be notified the potential UB can be triggered with the safe function in affected versions.

### DIFF
--- a/crates/capnp/RUSTSEC-0000-0000.md
+++ b/crates/capnp/RUSTSEC-0000-0000.md
@@ -1,0 +1,42 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "capnp"
+date = "2025-12-24"
+url = "https://github.com/capnproto/capnproto-rust/issues/605"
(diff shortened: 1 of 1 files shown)


djc:
Does this need to be updated to point to 0.25.0 as the fixing version?

dwrensha:
0.24.0 fixes the issue, but at the cost of adding `unsafe` markers in generated code that gets included in downstream projects.

0.25.0 removes the need for that `unsafe`. I expect that skipping version 0.24.0 will be the smoothest upgrade path for most projects.

shinmao:
Thanks @dwrensha for patch this issue. I think we can move forward to close this PR.

PR #2544, opened 2025-12-25
process_lock: add information about CVE-2025-48751

alexanderkjall:
information taken from: https://github.com/tickbh/ProcessLock/issues/1

@tickbh ok if this gets added to rustsec?

### DIFF
--- a/crates/process_lock/RUSTSEC-0000-0000.md
+++ b/crates/process_lock/RUSTSEC-0000-0000.md
@@ -1,0 +1,61 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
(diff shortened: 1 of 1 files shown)


alexanderkjall:
I tried running the linter locally, but I got:

```
$ /home/capitol/.cargo/bin/rustsec-admin lint --skip-namecheck rustdecimal,vec-const
error: error loading advisory DB repo from .: crates.io index error: unable to locate remote HEAD
```

Is there some trick to running it locally?

djc:
Do you have the same git rev of the rustsec-admin tool? I think the --skip-namecheck argument is gone.

PR #2551, opened 2025-12-30
Add advisory for gix-date vulnerability

shinmao:


djc:
@Byron are you okay with having this published as an advisory?

Byron:
If we do that, downstream CI would light up without recurse, something I'd like to avoid if possible. The ultimate decision I happily leave to you, but if possible, this PR could wait for https://github.com/GitoxideLabs/gitoxide/pull/2306 and a new release with the fix.

I should get to it tomorrow.

djc:
Oh sure, happy to wait until tomorrow.

Byron:
With https://github.com/GitoxideLabs/gitoxide/pull/2306 about to merge, it's clear that (unfortunately) the fix is a breaking change (that will probably affect no-one).
Thus, it should suffice to consider `gix-date` >= v0.12 as fixed, which I might just make available today as an extra release of `gitoxide`.

There are a couple more PRs to look at and merge before that happens though, so I think it's fair to either wait till tomorrow, or wait for my message here later today (assuming I don't forget which is a possibility).

Byron:
Oh, and I remember that it's possible to subscribe to [this discussion](https://github.com/GitoxideLabs/gitoxide/discussions/1693) which triggers a notification once the top-level `gitoxide` crate was released.

djc:
@shinmao can you update this with the patched version 0.12? (And squash your commits.)

shinmao:
@djc done. Thanks!

PR #2554, opened 2026-01-05
Add RUSTSEC advisory for rkyv issue #644

shinmao:
Have confirmed with the maintainer and fixed the bug in latest release


djc:
Thanks!

PR #2560, opened 2026-01-05
Add rustsec for virt1

cuiwenhao123:
The `Connect::open_read_only` method in the `Connect` implementation is causing a panic due to a NulError. The NulError indicates that the `CString` conversion failed due to the presence of a null byte (`\0`) in the input string.
The issue is [here](https://gitlab.com/libvirt/libvirt-rust/-/issues/31).


djc:
Is there a difference between this and #2558?

cuiwenhao123:
@djc #2558 and #2560 are issues caused by different APIs of the same library. However, both bugs are due to the same underlying cause.

djc:
I don't think it makes sense to have two advisories for this. Suggest to merge these. Please give the PR a useful title (not just the crappy branch name). Would be useful to have an `affected.functions` section that mentions both methods.

PR #2561, opened 2026-01-05
jj-cli: add information about CVE-2024-51990

alexanderkjall:
Information taken from: https://github.com/jj-vcs/jj/security/advisories/GHSA-88h5-6w7m-5w56

@yuja is it ok if this gets added to rustsec?


yuja:
Sounds good to me. @martinvonz, any thoughts?

PR #2564, opened 2026-01-06
NulError Panic in virt::Connect::open and virt::Connect::open_read_only

cuiwenhao123:
The Connect::open and Connect::open_read_only methods in the Connect implementation are causing a panic due to a NulError. The NulError indicates that the CString conversion failed due to the presence of a null byte (\0) in the input string.

The issue is [here](https://gitlab.com/libvirt/libvirt-rust/-/issues/31).

### DIFF
--- a/crates/virt/RUSTSEC-0000-0000.md
+++ b/crates/virt/RUSTSEC-0000-0000.md
@@ -2,8 +2,12 @@
 [advisory]
 id = "RUSTSEC-0000-0000"
 package = "virt"
-date = "2025-9-25"
+date = "2025-09-25"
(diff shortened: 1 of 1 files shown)


cuiwenhao123:
@alex For this issue, the crate author has acknowledged the problem and promised to fix it. Can it be assigned a RustSec ID?

djc:
Ultimately I think this doesn't warrant an advisory. Storing a null byte in a `&str` seems like something that would be extremely unlikely to get wrong by accident, and a panic in the case that happens doesn't seem all that noteworthy.

PR #2565, opened 2026-01-06
crates/rkyv/RUSTSEC-2026-0001.md:  Add v0.7 version with issue patched

mcronce:
Patched in https://github.com/rkyv/rkyv/pull/645; released with https://github.com/rkyv/rkyv/commit/2e47d8d9e027d7584f16c476bb698b0a35bc0b82

### DIFF
--- a/crates/rkyv/RUSTSEC-2026-0001.md
+++ b/crates/rkyv/RUSTSEC-2026-0001.md
@@ -8,7 +8,7 @@ categories = ["memory-corruption"]
 keywords = ["oom", "undefined-behavior"]
 
 [versions]
-patched = [">= 0.7.46", ">= 0.8.13"]
+patched = [">= 0.7.46, < 0.8.0", ">= 0.8.13"]
 ```
 
 # Potential Undefined Behaviors in `Arc<T>`/`Rc<T>` impls of `from_value` on OOM


crates/rkyv/RUSTSEC-2026-0001.md
```diff
@@ -8,7 +8,7 @@ categories = ["memory-corruption"]
 keywords = ["oom", "undefined-behavior"]
 
 [versions]
-patched = [">= 0.8.13"]
+patched = [">= 0.7.46", ">= 0.8.13"]
```
djc: I think this is wrong because `>= 0.7.46` implies that 0.8.0 is also patched, which it isn't. Suggest having a look at how some other advisories do it.

notonamap: apparently this is how it's done:

```suggestion
patched = [">= 0.7.46, < 0.8.0", ">= 0.8.13"]
```

for reference: https://github.com/rustsec/advisory-db/blob/main/crates/tokio/RUSTSEC-2025-0023.md?plain=1

mcronce: Thanks - I don't really use these

djc (approved):

PR #2569, opened 2026-01-06
Integer overflow in pcre2

cuiwenhao123:
The `CaptureLocations::get` method in `rust-pcre2/src/bytes.rs` may panic due to an integer overflow when calculating `i * 2`. This issue occurs when the input `i` is large enough to cause the multiplication to exceed the maximum value of `usize`.

```Rust
pub fn get(&self, i: usize) -> Option<(usize, usize)> {
        let ovec = self.data.ovector();
        let s = match ovec.get(i * 2) {
            None => return None,
            Some(&s) if s == PCRE2_UNSET => return None,
            Some(&s) => s,
        };
        let e = match ovec.get(i * 2 + 1) {
            None => return None,
            Some(&e) if e == PCRE2_UNSET => return None,
            Some(&e) => e,
        };
        Some((s, e))
    }
```
The issue is [here](https://github.com/BurntSushi/rust-pcre2/issues/50).
The fix is [here](https://github.com/BurntSushi/rust-pcre2/pull/51).

### DIFF
--- a/crates/pcre2/RUSTSEC-0000-0000.md
+++ b/crates/pcre2/RUSTSEC-0000-0000.md
@@ -1,0 +1,32 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "pcre2"
+date = "2025-09-24"
+url = "https://github.com/BurntSushi/rust-pcre2/issues/50"
+
+
+[versions]
+patched = [">= 0.2.11"]
+```
+
+# Integer overflow in CaptureLocations::get method
+
+The CaptureLocations::get method in rust-pcre2/src/bytes.rs:815:32 may panic due to an integer overflow when calculating i * 2. This issue occurs when the input i is large enough to cause the multiplication to exceed the maximum value of usize.
+
+```Rust
+pub fn get(&self, i: usize) -> Option<(usize, usize)> {
+        let ovec = self.data.ovector();
+        let s = match ovec.get(i * 2) {
+            None => return None,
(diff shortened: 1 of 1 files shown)


cuiwenhao123:
@alex Can I get a RustSec ID for this?

djc:
@BurntSushi do you think this qualifies as worthy of an advisory?

BurntSushi:
For a panic? No, I wouldn't assume that would be worthy of an advisory. I've certainly never considered filing for advisories for other bugs that could result in panics, and I feel like I don't see that as established practice in the crate ecosystem. But I'm not terribly familiar with the advisory guidelines. Is [this](https://github.com/RustSec/advisory-db/blob/main/CONTRIBUTING.md#criteria) what I should be looking at? I guess that does have:

> Panics in code advertised as "panic-free" (particularly if useful for network DoS attacks)

I guess this routine wasn't explicitly advertised as "panic-free," but that was the intent. On the other hand, this was an integer overflow, so unless overflow checks were explicitly enabled, you wouldn't get a panic in a release profile.

I guess speaking for me personally, I'm not especially inclined to think an advisory is worth it here. But I admit I'm making a subjective judgment based on what I think the actual impact of this is beyond what the guidelines say.

djc:
Additional recent context/discussion here:

- #2572

BurntSushi:
Ah yeah reading that discussion, I'd say this does not qualify. In particular, the _input_ to this function is rarely from end users. I mean, it could be, but it's not common in my experience. Especially over a network boundary.

Issue #2572, opened 2026-01-06
Discussion: signal/noise ratio too low?


djc:
From the rkyv maintainer:

> I think the advisory database has a signal-to-noise ratio that is too low. While there are many legitimate advisories, there are also plenty of advisories that are just pedestrian bugs without a specific impact on security. If codebases have an "attention budget" for preventing vulnerability exploitation, advisories act like a heuristic for where to allocate that attention. The bar for moving that attention to specific issues should be relatively high because we already have good general-purpose heuristics to follow. Using this issue as an example:
> 
> * The threat model is not well-defined. An attacker would have to be able to OOM a process as this specific subroutine is being executed. OOM-ing before would most likely cause the process to safely panic, and rkyv itself does not provide any kind of mechanism for causing an OOM.
> 
> * The actual outcome is undefined behavior. UB can lead to a vulnerability alone, but it's a very broad class of errors and many of them do not have that kind of potential. Here, a null pointer gets turned into an out pointer and written to. In most situations, that will become a segfault. rkyv doesn't guarantee panic-freedom, so just halting isn't a security issue. Obviously there's no guarantee that the compiler will actually produce code that does this, but...
> 
> * There is no proof of concept. This would be a very difficult bug to trigger even under controlled circumstances, and if you can't cause this bug even with everything going your way then I don't think it's worth bringing a lot of attention to.
> 
> Overall, I'd consider this the equivalent of a ["bogus CVE"](https://lwn.net/Articles/944209/) for RUSTSEC (as RUSTSEC is not a CNA and does not issue CVEs). The incentives leading to bogus CVEs are systemic, and I won't waste time fighting them when it doesn't make a difference. [I take security issues seriously](https://github.com/rkyv/rkyv?tab=security-ov-file#readme) and [I do file CVEs](https://nvd.nist.gov/vuln/detail/CVE-2021-31919) to notify users of serious issues when warranted so they can protect themselves. But it's not my job to keep other people's vulnerability databases clean - if RUSTSEC wants people to use and rely on their advisories then the onus is on them.

There have been a bunch of high-volume lower-quality submitters active and I've been facilitating that a little bit (mainly because reviewing advisories is volunteer work and so I'm trying to limit the amount of time spent on it).

Should we establish/document some clearer guidelines to increase the bar, especially for higher volume advisory submitters?

cc @alex @tarcieri @Shnatsel

alex:
[...]

PR #2576, opened 2026-01-07
lru: add advisory for Stacked Borrows violation

paolobarbolini:
Adds the advisory for https://github.com/jeromefroe/lru-rs/pull/224.

@jeromefroe let us know if you are ok with this.

### DIFF
--- a/crates/lru/RUSTSEC-0000-0000.md
+++ b/crates/lru/RUSTSEC-0000-0000.md
@@ -1,0 +1,24 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "lru"
+date = "2026-01-07"
+url = "https://github.com/jeromefroe/lru-rs/pull/224"
+informational = "unsound"
+categories = ["memory-corruption"]
+keywords = ["stacked-borrows"]
+
+[versions]
+patched = [">= 0.16.3"]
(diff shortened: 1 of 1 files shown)


alamb:
Can someone please update the advisory to explain more about what the impact of the issue to help downstream users  evaluate if they are affected by the issue? Specifically, the PR https://github.com/jeromefroe/lru-rs/pull/224 says

> This invalidates the pointer held within KeyWrapper by the HashMap, but the HashMap still holds and accesses it on subsequent reads or writes to the LRU, which is unsound.

What possible exploit / corruption is possible due to this unsoundness?  The PR points that MIRI calls out a potential issue, but the implications are hard to understand. 

Is the idea that anything using `IterMut` on an LRU cache to iterate and mutate the contents can cause memory corruption?

paolobarbolini:
AFAIK (not a Stacked Borrows expert) on **current compilers** this **should not** lead to UB, which is why I marked it as `informational = "unsound"`.

alamb:
> AFAIK (not a Stacked Borrows expert) on **current compilers** this **should not** lead to UB, which is why I marked it as `informational = "unsound"`.

I see -- if that is the case,  then I think this is another example of contributing to a low signal to noise ration in the advisory database:

PR #2584, opened 2026-01-10
trailer: add information about CVE-2025-47737

alexanderkjall:
Information taken from: https://github.com/Geal/trailer/issues/2

@Geal any opposition to adding this to rustsec?

### DIFF
--- a/crates/trailer/RUSTSEC-0000-0000.md
+++ b/crates/trailer/RUSTSEC-0000-0000.md
@@ -1,0 +1,52 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "trailer"
+date = "2025-05-04"
+url = "https://github.com/Geal/trailer/issues/2"
+aliases = ["CVE-2025-47737"]
+cvss = "CVSS:3.1/AV:L/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:L"
+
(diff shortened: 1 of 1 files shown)


alexanderkjall:
I guess we won't get any answers from the maintainer on this one :(

djc:
I've reached out to the maintainer by email.

djc:
@Geal another ping for this one as well.

Geal:
You can add this advisory. I will soon retire that crate anyway, it was used only in one project that I don't work on anymore

crates/trailer/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,49 @@
...
+     unsafe { 
+         let trailer = Trailer::allocate(capacity); 
+         let ptr = trailer.ptr as *mut T; 
+         ptr.write(T::default()); 
+         trailer 
+     } 
+ } 
+```
+
+The constructor does check the T is not a ZST in
```
djc: I think this should be "does not check"?

Please also add a note that this crate is not being maintained and will not be fixed.

alexanderkjall: Good catch, thanks.

Note added about trailer being unmaintained.

djc (approved):

PR #2585, opened 2026-01-10
below: add information about CVE-2025-27591

alexanderkjall:
information taken from: https://www.openwall.com/lists/oss-security/2025/03/12/1

@lnyng any objections to adding this to rustsec?

### DIFF
--- a/crates/below/RUSTSEC-0000-0000.md
+++ b/crates/below/RUSTSEC-0000-0000.md
@@ -19,7 +19,7 @@ patched = [">= 0.9.0"]
 Below is a tool for recording and displaying system data like
 hardware utilization and cgroup information on Linux.
 
-## 2) Symlink Attack in `/var/log/below/error_root.log`
+## Symlink Attack in `/var/log/below/error_root.log`
 
 Below's systemd service runs with full `root` privileges. It attempts to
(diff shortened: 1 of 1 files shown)


lnyng:
Looks good to me

djc:
Please drop all the narrative and timeline stuff and only keep a short explanation of the actual issue. People can click on the links if they need more details.

alexanderkjall:
@djc updated

crates/below/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,101 @@
...
+[versions]
+patched = [">= 0.9.0"]
+```
+
+# World Writable Directory in /var/log/below Allows Local Privilege Escalation
+
+Below is a tool for recording and displaying system data like
+hardware utilization and cgroup information on Linux.
+
+## 2) Symlink Attack in `/var/log/below/error_root.log`
```
djc: You have `2)` (and `3)`, below), but no 1). Maybe just get rid of the header numbers?

alexanderkjall: good catch, thanks :)

They are removed now

PR #2586, opened 2026-01-11
Update RUSTSEC-2025-0141.md: Add Speedy as an alternative

asomers:
Add the Speedy crate as an alternative to Bincode.  I find it to be a closer replacement than rkyv or bitcode.


djc:
> I find it to be a closer replacement than rkyv or bitcode.

Why?

asomers:
> > I find it to be a closer replacement than rkyv or bitcode.
> 
> Why?

* Bitcode does not attempt to maintain a stable format across versions.
* rkyv's philosophy leads to a very different on-disk format.  For example, its use of relative pointers.

Also, I didn't realize when I opened this PR, but there's an even closer alternative that has recently popped up: bincode-next.  It's a direct fork of bincode.

djc:
I feel like these concerns are adequately covered by wincode, which is already in there.

asomers:
Ok, @djc .

PR #2587, opened 2026-01-11
lucet-runtime: add information about CVE-2021-43790

alexanderkjall:
information taken from: https://github.com/bytecodealliance/lucet/security/advisories/GHSA-hf79-8hjp-rrvq

@boxdot any objections to adding this to rustsec?

### DIFF
--- a/crates/lucet-runtime/RUSTSEC-0000-0000.md
+++ b/crates/lucet-runtime/RUSTSEC-0000-0000.md
@@ -1,0 +1,65 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "lucet-runtime"
+date = "2021-11-30"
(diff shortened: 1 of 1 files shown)


boxdot:
Any particular reason why I was tagged on this PR?

djc:
> Any particular reason why I was tagged on this PR?

Seems like that was a mistake, sorry about that! @pchickey maybe?

pchickey:
Lucet was archived in March 2022 https://github.com/bytecodealliance/lucet so I don't expect having this database entry in 2026 is actually useful. But it doesn't have factual problems

djc:
@alexanderkjall can you look into the CI lint failure here?

alexanderkjall:
@djc done, I spent quite a bit of time to get the linter to run locally without success. I think it was some sort of problem in gix, and it seems that tame-index and rustsec is moving towards sparse-indexes over https instead? So maybe not worth the time to get to the bottom of.

PR #2588, opened 2026-01-15
cmov: add advisory for non-constant-time code generation

tarcieri:
This vulnerability is has also been reported as CVE-2026-23519 and GHSA-2gqc-6j2q-83qp


tarcieri:
Temporarily converting this to draft: I'm looking for ways to get actual guarantees on ARM32 and if I can get a release out with that quickly I'd like to update the advisory with information about it

PR #2590, opened 2026-01-16
Fix list formatting for alternatives to bincode (RUSTSEC-2025-0141.md)

badboy:
Due to the indentation this was rendered as code, not as a list with proper links


djc (approved):
Ahh, thanks!

PR #2593, opened 2026-01-17
Update maintenance status and alternatives in RUSTSEC-2024-0436.md

esrauch:
Add additional alternative of paste-complete which is another fork which does not intend to add additional features

### DIFF
--- a/crates/paste/RUSTSEC-2024-0436.md
+++ b/crates/paste/RUSTSEC-2024-0436.md
@@ -17,4 +17,5 @@ that this project is not longer maintained as well as archived the repository
 
 ## Possible Alternative(s)
 
+- [paste-complete](https://crates.io/crates/paste-complete), a fork of paste which is maintained but intends to only make changes in the event of security-sensitive issues.
 - [pastey](https://crates.io/crates/pastey), a fork of paste and is aimed to be a drop-in replacement with additional features for paste crate


esrauch:
Please see the (hopefully not too inflammatory) text on https://crates.io/crates/paste-complete for the motivation here.

As a concrete motivating explanation, as Google's official Protobuf crate, `paste` seems like the correct choice of dep and we wouldn't ant to take a dep onto `pastey`. But, if Protobuf publishes with a dep on `paste` will cause alarm and concern because of this rustsec.

As near term workaround I have published paste-complete is published and I can subscribe to that it is maintained. I assume Rustsec thinks these notices are WAI and so probably the right long term here is for `paste` crate to be taken away from current maintainer and moved to someone else who would be considered an active maitnainer an then the rustsec can be dropped.

djc:
I'm not interested in advertising this crate with its zero downloads. People can always find it themselves when they need it.

esrauch:
Yes, I did create it just to try to force some resolution on this topic.

I'd still like to ask to have the discussion here even in that case; listing Pastey in the notice seems problematic by comparison by driving people from a "good" package to something that is much higher risk to get bad patches.

I think looking at the prior issues discussing this rustsec in other projects, very few projects view Pastey as a viable alternative to the archived Paste project from a security perspective (I didn't actually find any case that did when I searched for GitHub issues discussing how to resolve the problem of this rustsec showing up as a problem).

Maybe it's appropriate to remove Pastey or include more disclaimers that the notice is that Paste is archived but has no known security concerns?

esrauch:
Actually, is there a better process to discuss if this rustsec should be dropped/retracted instead? Maybe better topic on a github issue instead of in this PR comments?

Unmaintained notices being in rustsec make a lot of sense for certain libraries; if (for example) Tonic or Prost or whatever performed a similar action, where those foreseeably should expect to need patches for security/correctness, it is a good idea to warn transitive users that they might be using something that will foreseeably fall into being a security concern in the future.

Archiving of paste by contrast seems exactly premised on 'this will never need a security patch' and it seems the community is generally in agreement with that (where the response from ~all projects is just to denylist this rustsec rather than take action). Then (unlike other cases) having the rustsec published for this specific case seems to really be aligned with the community understanding of the situation.

djc:
I'm open to a discussion on whether the paste advisory should be withdrawn. Suggest opening a PR to do that with a bunch of context and discussion in the PR description.

tarcieri:
@esrauch there have been various issues opened effectively complaining or questioning the usefulness of tracking unmaintained crates like `paste`, and perhaps we should put some canned answers in `HOWTO_UNMAINTAINED.md`.

If you feel a crate is truly "done" and will never receive updates, one option is to vendor the code somewhere, or as you have done make a fork you control which allows you to reuse it across multiple other crates. At least in other package ecosystems, unmaintained packages are frequent targets of supply chain attacks, and taking either of these steps will protect you from them. Now I sure hope that doesn't happen to the particular author of this crate as they have unusually widespread reach in the Rust ecosystem, but that's one general motivation and the steps you can take to address it. [Being able to do security response when issues are discovered is another](https://edera.dev/stories/tarmageddon).

However to address some of your other concerns:

I too have been a bit concerned about `pastey` given its widespread usage, relatively unknown author, and currently failing CI. On the other hand I don't want to play gatekeeper. Perhaps we can give a clearer disclaimer in our recommendations.

Something we've discussed in the past that could potentially be implemented (not sure if there's an open issue offhand) is leveraging the dependency graph analysis in `cargo-lock` to locate the toplevel packages in a workspace and then only apply (at least by default) informational advisory checks to immediate, rather than transitive, dependencies. There seems to be a bit of a thundering hurd problem when we file unmaintained advisories, tooling opens issues on downstream projects (which is also something we should try to fix), and then downstream users all go to complain to upstream maintainers.

esrauch:
> At least in other package ecosystems, unmaintained packages are frequent targets of supply chain attacks, and taking either of these steps will protect you from them

I don't personally have an opinion one way or the other but I think the root cause problem is effectively an ecosystem misalignment regarding this sentence.

Rust ecosystem also likes the idea of 'complete', paste is something of a 'missing part of the std' crate. If it was true that this being abandoned is a supply chain concern then probably crates.io should move ownership of the crate to a new (credible/known/respected) maintainer. But they don't because they don't see the situation as a supply chain concern (no notices on the 'security' tab on crates.io)

So now the situation is kind of badly stuck in a bad local minimum; if we release a major new crate that uses paste it will trigger a bunch of noisy security notices on our users. But, since the ecosystem otherwise _doesn't_ agree that this is actually a security concern the crate is not going to 'become maintained' again (either by the paste crate being moved to a maintainer or a successor crate established).

tarcieri:
That doesn't address my point. Crates that are "done": not receiving updates, neglected by their maintainers, but widely used in the ecosystem, remain juicy targets for supply chain attacks.

> if we release a major new crate that uses paste it will trigger a bunch of noisy security notices on our users.

IMO this is the problem I think we really need to fix. [See the last paragraph of my previous response](https://github.com/rustsec/advisory-db/pull/2593#issuecomment-3764324834)

esrauch:
> That doesn't address my point. Crates that are "done": not receiving updates, neglected by their maintainers, but widely used in the ecosystem, remain juicy targets for supply chain attacks.

I think this is still what I was trying to get at in my previous reply but maybe I wasn't clear.

Tooling identifying it as bad (even for only depth=1 rdeps) means that it identified it as _some_ form of bad state that  warrants "maybe you should reconsider if you really want to do this" to the users of it. You are saying why users of paste should be informed and nudged to reconsider if they really want to use paste; I understand that argument and I'm not trying to argue for or against that.

The problem is that the paste crate is not getting a new maintainer nor a established-replacement created due to the consensus seems to be _not_ accepted as a bad state, even to a degree of "depth=1 rdeps should be opt-in info-notified about this". I think if there was even an info(soundness) notice on paste the situation would actually get resolved in some way or another (either by paste getting a patch or a clean successor crate being established).

Basically then the only 'problem' state is that info(unmaintained) is both being flagged as at all (under a reasonable argument that you are saying), but not otherwise consensus accepted as even the most milquetoast/mild possible security topic (since I believe if it was accepted that there was even the most extremely mild possible security concern that there would get _some_ resolution in the case of paste crate specifically).

https://github.com/rustsec/advisory-db/issues/2572 is very pertinent discussion but IMO it seems even more extreme in this case compared to the info(soundness) notices.

tarcieri:
There are actions anyone adding an unmaintained crate to their project should consider to prevent supply chain attacks. At the very least, they should probably use a `=` (or failing that, `<=`) pin on the version requirement, so if an unexpected new version is published it won't be picked up automatically.

We could probably benefit by documenting both the rationale for why we collect and report this information, as well as what steps users should take if they do choose to depend on an unmaintained crate.

I think community sentiment is split on the topic of unmaintained crates. In the wake of Tarmageddon there was quite a bit of recent concern about it. On the other hand there are people like yourself who simply don't care that crates are unmaintained and consider the tracking around them a burden.

esrauch:
>  On the other hand there are people like yourself who simply don't care that crates are unmaintained and consider the tracking around them a burden.

This is my personal github account a little bit poking the bear in 'free time' compared to speaking from my Google corp github role, but I think from the point of view of Google Protobuf TL I wouldn't be in a position to tell other people they shouldn't care. For context, we really couldn't accept a dep with a CVE pubsilhed even if that CVE was 100% inapplicable (or false/refuted), because just the confusion and concern problem is enough even if we knew for sure there was no security concern at all. And it seems rustsec informational advisories are viewed the same as low-impact CVEs, the entry ends up being a defacto ban from us using those crates.

If the real 'ground truth' is really that even ~20% of the Rust community believes that Paste is a security issue, then I probably do need to advocate for a replacement crate even if the Google-staffed Rust library owners believe there is no security concern. It's unfortunately a very awkward state if we have to advocate to people who think that the state is 'correct' that if rustsec is taking the decision that even causes a perception of a security concern, that it is irrelevant whether it is securty concern (or even if rustsec itself disclaims it is a security issue or is just providing information).

In reality/pragmatically, we could always do something like publish `google-paste` exactly only to bypass the rustsec here for all Google publislhed libraries, but it feels like a bad solution for the ecosystem compared to if we could either get consensus that `paste` is either ok or sufficiently not ok that `paste` crate name would be transferred to someone else even over the original authors objection (the fact that he is realistically in a position to object would imply that its not actually abandoned).

8573:
> crate name would be transferred to someone else even over the original authors objection

Has the Crates.io team ever done this with a crate that was not a stub with no useful code? My impression is that they consistently have had the stance that they will not exercise such a level of governance unless compelled by law, and that non-stub crates can be managed however their owners like as long as they are not outright malicious.

8573:
> There are actions anyone adding an unmaintained crate to their project should consider to prevent supply chain attacks. At the very least, they should probably use a `=` (or failing that, `<=`) pin on the version requirement, so if an unexpected new version is published it won't be picked up automatically.

What kinds of supply chain attack does dependency pinning not prevent, if published crates are immutable? (assuming the pin is to a published crate on Crates.io, not a Git repo with no hash)

tarcieri:
@8573 I recommended dependency pinning specifically to mitigate the supply chain attack.

My next recommended steps would be the [ones I suggested earlier](https://github.com/rustsec/advisory-db/pull/2593#issuecomment-3764324834) to either vendor the code or maintain your own fork, but if both of those are unacceptable, pinning the version is sufficient to mitigate the potential attack.

esrauch:
Vendoring the code inside the published crate has some licensing annoyances (the overall crate has to establish the corresponding individual licenses, which don't cover the total crate, which is always possible but in other cases we try hard to avoid that kind of licensing situation on our published libs).

With pinning, it will still end up flagged to our users (modulo tooling changes that are being discussed on the other GH issue).

Forking does work pragmatically here, and wearing my Google hat it makes sense to just do that and call it a day.  Wearing my 'just some guy' hat that resolution just seems 'smelly' to me. It would basically be that the lifecycle of a crates could be "in development, then complete, now everyone should start vendoring/forking it starting now".

A lot of other libraries being abandoned it seems like that would be completely sensible for the understanding to be "yeah you're gonna want to fork this if you want to keep using it from now on, since this is a random library and the maintainer has abdicated", but for Paste the "yeah the expectation is that everyone should be vendoring this from now on" feels like it would be a pretty big ecosystem wart.

tarcieri:
> for Paste the "yeah the expectation is that everyone should be vendoring this from now on" feels like it would be a pretty big ecosystem wart.

`concat_idents!` is ultimately a missing compiler feature.

Edit: to expound upon that, [in the past I've tried to drum up interest in getting these sort of compiler-gap crates integrated into the language as first-class features](https://internals.rust-lang.org/t/calculating-which-3rd-party-crates-are-good-candidates-for-std-inclusion-via-left-pad-index/11129), which helped recruit effort to get the `matches` crate integrated as the `matches!` macro.

Perhaps the situation here is enough to drum up interest in actually finally getting ident concatenation integrated as a first-class language feature?

djc:
> Maybe it's appropriate to remove Pastey or include more disclaimers that the notice is that Paste is archived but has no known security concerns?

That seems okay to me.

> Something we've discussed in the past that could potentially be implemented (not sure if there's an open issue offhand) is leveraging the dependency graph analysis in `cargo-lock` to locate the toplevel packages in a workspace and then only apply (at least by default) informational advisory checks to immediate, rather than transitive, dependencies. There seems to be a bit of a thundering hurd problem when we file unmaintained advisories, tooling opens issues on downstream projects (which is also something we should try to fix), and then downstream users all go to complain to upstream maintainers.

I don't see that as a viable solution. Ultimately the people deploying the software (that is, generating/maintaining the lockfile) are responsible for what they ship, and if the maintainer of an intermediate package becomes inactive this would result in this problem never getting solved.

> Archiving of paste by contrast seems exactly premised on 'this will never need a security patch' and it seems the community is generally in agreement with that (where the response from ~all projects is just to denylist this rustsec rather than take action). Then (unlike other cases) having the rustsec published for this specific case seems to really be aligned with the community understanding of the situation.

I mostly agree with your argument that the risk for paste specfifically is very low. To be clear, I think you forking the crate is perfectly okay -- I just don't want to recommend it to everyone when it has basically no validation (but that could change over time). It might be cool if you can move the repo in github.com/google, even if issues are still disabled or whatever as a sign it is done.

> If it was true that this being abandoned is a supply chain concern then probably crates.io should move ownership of the crate to a new (credible/known/respected) maintainer. But they don't because they don't see the situation as a supply chain concern (no notices on the 'security' tab on crates.io)

crates.io as a matter of policy never moves ownership to a new maintainer, as I understand it, which seems like a decent choice in terms of mitigating supply chain concerns.

> On the other hand there are people like yourself who simply don't care that crates are unmaintained and consider the tracking around them a burden.

That seems like a substantial misrepresentation of the OP's arguments so far, which doesn't seem helpful.

> Forking does work pragmatically here, and wearing my Google hat it makes sense to just do that and call it a day. Wearing my 'just some guy' hat that resolution just seems 'smelly' to me. It would basically be that the lifecycle of a crates could be "in development, then complete, now everyone should start vendoring/forking it starting now".

I think there's a difference between "complete" and "all maintainers are abdicating responsibility", and the latter could still cause risks even if they might be limited in practice. Practically, I think forking paste for this case is the right thing.

(Are you really sure you need paste and can't address the needs for identifiers with something like syn directly? That's what I've done in the past.)

> A lot of other libraries being abandoned it seems like that would be completely sensible for the understanding to be "yeah you're gonna want to fork this if you want to keep using it from now on, since this is a random library and the maintainer has abdicated", but for Paste the "yeah the expectation is that everyone should be vendoring this from now on" feels like it would be a pretty big ecosystem wart.

I have worked with a lot of maintainers (see some mentions [here](https://dirkjan.ochtman.nl/writing/2026/01/09/reviewing-2025.html)) to continue maintenance of crates they didn't want to maintain anymore, and I think this often a viable approach. Since I don't have a need for paste, and AFAIK none of my sponsors do, I haven't asked the original maintainer if they'd be interested that -- and they might not be anyway.

> @8573 I recommended dependency pinning specifically to mitigate the supply chain attack.

Pinning doesn't help with downstreams, because they'll still get alerted, I guess? It might be interesting if old pinned versions wouldn't cause alerts, although that would be somewhat tricky to implement in cargo-audit (and cargo-deny).

PR #2596, opened 2026-01-21
Add rust-sec for triton-vm

jan-ferdinand:
For some versions, the verifier of the Triton VM proof system failed to complete all steps necessary to guarantee that a given proof was, in fact, a proof.


djc:
Thanks!

PR #2600, opened 2026-01-25
Add advisory on use-after-free in `oneshot` library

faern:
See https://github.com/faern/oneshot/issues/73 for details. I discovered this race just a few days ago. I got the patch working yesterday, and managed to release a patched version (0.1.12) today.

This race exists in all versions of `oneshot`, except by accident not in the initial `0.1.1` release. In the first release this would just cause a panic instead, which would at least not be UB.


djc:
Thanks!

PR #2604, opened 2026-01-27
Add CVE-2026-24116 for Wasmtime.

cfallin:
This is a database entry for our recent advisory here: https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-vc8c-j3xm-xj73

### DIFF
--- a/crates/wasmtime/RUSTSEC-0000-0000.md
+++ b/crates/wasmtime/RUSTSEC-0000-0000.md
@@ -1,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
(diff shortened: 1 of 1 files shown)


djc:
Thanks! What's the reason for the gaps in the version ranges?

cfallin:
> What's the reason for the gaps in the version ranges?

We patched our v41, v40, and v36 releases -- v41 and v40 because we support the last two ordinary releases, and v36 because it is our current Long-Term Support (LTS) release.

djc:
Makes sense -- sorry, I got confused for a bit and thought this was a sparse representation (and thus faulty) representation of affected versions.

PR #2614, opened 2026-01-31
Add unmaintained advisory for filum

llooFlashooll:


djc:
Please get approval from maintainers for each of your advisories. If there is an actual security issue (not just panicking on bad input), it would be better to file an advisory for that instead.

llooFlashooll:
Your point is right! However, I created the issue half a year ago, and haven't received replies from the maintainers. This is the main reason I think unmaintained advisory is a better option.

PR #2618, opened 2026-01-31
Add unmaintained advisory for tinycdb-rs

llooFlashooll:


s14051sa-png:
Reset withdrawal

CodesInChaos:
> as the last commit was 11 years ago.

I think it would be clearer to put the year number in the description, since "x years" will become increasingly misleading as time passes.

djc:
It would be better to make this an advisory about the unsoundness issue, with a note that it's unlikely to get fixed.

Nevertheless, we should try to contact the maintainer. @andrew-d any reasons we should not publish an advisory like this?

andrew-d:
> It would be better to make this an advisory about the unsoundness issue, with a note that it's unlikely to get fixed.
> 
> Nevertheless, we should try to contact the maintainer. @andrew-d any reasons we should not publish an advisory like this?

Hah, I hadn't thought about this crate in a while. @llooFlashooll - my apologies, I think I missed your notification on the repo. I will do my best to update and do a release at some point in the next few weeks; if I don't get to this or decide that it's too much work for the time I have available, then I'll respond here and we can mark it as unmaintained.

djc:
Much appreciated!

PR #2620, opened 2026-02-02
Add RUSTSEC advisory for git2

BoLu1225:
Have confirmed with the maintainer and fixed the bug in latest release

### DIFF
--- a/crates/git2/RUSTSEC-0000-0000.md
+++ b/crates/git2/RUSTSEC-0000-0000.md
@@ -4,6 +4,7 @@ id = "RUSTSEC-0000-0000"
 package = "git2"
 date = "2026-02-02"
 url = "https://github.com/rust-lang/git2-rs/pull/1213"
+informational = "unsound"
 categories = ["memory-corruption"]
 keywords = ["undefined-behavior"]
 


djc:
@ehuss do you agree that publishing an advisory for this is useful? Should it be [`informational = "unsound"`](https://github.com/rustsec/advisory-db/blob/main/MAINTAINERS_GUIDE.md#check-that-the-metadata-is-correct)?

crates/git2/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,16 @@
...
+url = "https://github.com/rust-lang/git2-rs/pull/1213"
+categories = ["memory-corruption"]
+keywords = ["undefined-behavior"]
+
+[versions]
+patched = [">=0.20.4"]
+```
+# Potential undefined behavior when dereferencing Buf struct 
+
+if we dereference the Buf struct right after calling new() on Buf struct, it passes Null Pointer to the unsafe function slice::from_raw_parts. Based on the safety section documentation of function,
```
cuviper:
```suggestion
if we dereference the Buf struct right after calling new() or default() on Buf struct, it passes Null Pointer to the unsafe function slice::from_raw_parts. Based on the safety section documentation of function,
```

cuviper (commented):
FYI we are discussing in the security response team whether this warrants a full CVE, but it didn't come to our attention until after the fact. I personally think _at least_ `informational = "unsound"` is reasonable though, and I suppose we can update rustsec if we decide to go further.

djc:
> FYI we are discussing in the security response team whether this warrants a full CVE, but it didn't come to our attention until after the fact. I personally think _at least_ `informational = "unsound"` is reasonable though, and I suppose we can update rustsec if we decide to go further.

Thanks. @BoLu1225 can you add the `informational` metadata?

PR #2626, opened 2026-02-05
Add advisory for `time`

jhpratt:


djc:
Would be nice to add a little prose here.

(I'll see if I can fix that the linting error has no details.)

jhpratt:
I was just doing that — copying over the advisory text specifically

jhpratt:
Fixed an obvious syntax error in the toml bit, so now I'm not sure what's failing the linter :slightly_smiling_face:

tnull:
So this fixed one patch release after bumping MSRV on the prior patch release, leaving downstream users that can't arbitrarily upgrade `rustc` vulnerable? Cool cool.

jhpratt:
@tnull Be respectful. I patched this on short notice while dealing with a less-than-great event. This is absolutely not the place to discuss MSRV policy. You've already opened an issue in the time-rs/time repo and had it addressed.

If you want to harass me, you will be blocked and reported to GitHub.

crates/time/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,42 @@
...
+[versions]
+patched = [">= 0.3.47"]
+unaffected = ["< 0.3.6"]
+```
+
+### Impact
+
+When user-provided input is provided to any type that parses with the RFC 2822 format, a denial of
+service attack via stack exhaustion is possible. The attack relies on formally deprecated and
+rarely-used features that are part of the RFC 2822 format used in a malicious manner. Ordinary,
```
tnull, line 31: It would be good to clarify exactly which features these are, so users can check whether they are affected or not.

jhpratt: If someone is handling user input without further verification (which is expected), they are affected. It's not an optional feature — it's a core part of the specification. There truly is zero chance that someone would encounter this by accident; it would have to be malicious.

While I'm being vague, that's because it hasn't been exploited to my knowledge. Obviously it can be inferred from the recent patch, but I at least want it to take nonzero effort to exploit. I was provided with a proof of concept that was trivially verified.

djc:
> Fixed an obvious syntax error in the toml bit, so now I'm not sure what's failing the linter 🙂

It's "Expected # header after TOML front matter". Fixing the linter output in

- https://github.com/rustsec/rustsec/pull/1528

jhpratt:
Ah, it didn't like the `###` I used. I've changed the headers to all be H1, in addition to adding the now-issued CVE ID.

crates/time/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,42 @@
...
+"time::PrimitiveDateTime::parse" = [">= 0.3.6, < 0.3.47"]
+"time::OffsetDateTime::parse" = [">= 0.3.6, < 0.3.47"]
+"time::UtcDateTime::parse" = [">= 0.3.38, < 0.3.47"] # type not present until 0.3.38
+
+[versions]
+patched = [">= 0.3.47"]
+unaffected = ["< 0.3.6"]
+```
+
+# Impact
```
alexander-kjall-hafslundeco, line 27: Sorry for the late feedback, but this header is used to describe the vulnerability here: https://rustsec.org/packages/time.html

Maybe it can be improved to something more descriptive?

djc: Would be nice -- the fastest way to get that done would be if you can submit a PR.

jhpratt: Ah, go for it. I went for "Stack exhaustion denial of service attack" as the advisory title on GitHub.

alexander-kjall-hafslundeco: https://github.com/rustsec/advisory-db/pull/2629

PR #2629, opened 2026-02-06
improve header slightly for RUSTSEC-2026-0009

alexander-kjall-hafslundeco:


jhpratt:
Should probably change the existing headers to h2, adding a new h1 at the top.

alexander-kjall-hafslundeco:
fixed

PR #2636, opened 2026-02-10
Add patched versions to RUSTSEC-2025-0142

dlon:
See https://github.com/mullvad/mnl-rs/issues/15 and https://github.com/mullvad/mnl-rs/pull/21 (the fix).


djc:
Thanks!

Issue #2646, opened 2026-02-16
Security advisoriy PR #2637 closed without explanation


nadimkobeissi:
I'm writing to seek to understand why our proposed advisory in https://github.com/rustsec/advisory-db/pull/2637 was closed.

For context: as part of our research paper ["Verification Theatre: False Assurance in Formally Verified Cryptographic Libraries,"](https://eprint.iacr.org/2026/192) we discovered five vulnerabilities in the cryptographic libraries hpke-rs and libcrux-psq. We submitted fixes upstream, which the maintainers (Cryspen) eventually applied by copy-pasting our code from rejected PRs, but they did not issue security advisories. We then filed PR #2637 to the RustSec advisory database covering the three most critical issues:

1. **hpke-rs nonce reuse via sequence number overflow** — the library stores HPKE sequence numbers as `u32` instead of checking against the RFC 9180 mandated limit of 2^96-1, with the overflow guard being ineffective in release builds, enabling nonce reuse in AES-GCM and allowing plaintext recovery and authentication key compromise.
2. **hpke-rs missing X25519 zero-check validation** — RFC 9180 Section 7.1.4 mandates checking whether the DH shared secret is all-zero, but hpke-rs skips this in both the RustCrypto and libcrux backends, allowing attackers to supply low-order points and force predictable session keys.
3. **libcrux-psq AES-GCM decryption panic** — `decrypt_out` calls `.unwrap()` on AES-GCM decryption results, causing process crashes on malformed ciphertexts across nine of eighteen supported ciphersuites.

@djc initially asked us to wait two weeks for maintainer involvement before proceeding. During that period, Cryspen published their own GitHub Security Advisories, but these had significant quality issues. In particular, GHSA-g433-pq76-6cmf classified a nonce-reuse vulnerability enabling full plaintext recovery as merely "moderate" severity, lacked details and impact sections, and bundled multiple distinct vulnerabilities into a single advisory without adequate documentation of each. When I raised these concerns, the Cryspen maintainer (@franziskuskiefer) blocked me on GitHub rather than addressing the technical feedback.

Despite all of this, and despite @pinkforest correctly pointing out that RustSec has in fact merged advisories without maintainer buy-in in the past, @djc closed the PR on February 16 without providing any explanation.

I'd like to understand:

1. **Why was the PR closed?** No reason was given, and my question asking for one has gone unanswered.
[...]

PR #2637, opened 2026-02-10
Push libcrux-psq and hpke-rs advisory proposals

nadimkobeissi:
These are proposals for the more impactful vulnerabilities discovered as part of this research work: 

[_Verification Theatre: False Assurance in Formally Verified Cryptographic Libraries_](https://eprint.iacr.org/2026/192)

Cryspen has not issued any security advisory for any of these bugs despite patching them four days after I initially submitted my fixes.

In November 2025, another serious security issue was discovered in the `libcrux-intrinsics` crate, and Cryspen *also* did not issue a security advisory; other people had to take it upon themselves to [submit one](https://rustsec.org/advisories/RUSTSEC-2025-0133.html) to this database, and hope that a direct announcement from Cryspen would follow (it never did).

Hence, I am doing the same thing here.

The full disclosure timeline, including Cryspen's response, is documented in the paper.

Happy to accept any feedback on the wording of any of the submitted issues. Thanks.

PR #2647, opened 2026-02-16
RustSec Advisories for hpke-rs and libcrux-psq

nadimkobeissi:
This PR proposes three RustSec advisories for cryptographic vulnerabilities discovered during research (["Verification Theatre: False Assurance in Formally Verified Cryptographic Libraries"](https://eprint.iacr.org/2026/192)). The upstream maintainers (Cryspen) have patched these issues but have not filed RustSec advisories, which means `cargo audit` does not currently flag affected versions.

Cryspen has published GitHub Security Advisories (GHSAs) for some of these issues. However, we believe RustSec advisories are still warranted for the reasons outlined below.

### Proposed Advisories

**1. hpke-rs: Nonce reuse via sequence number overflow**

The library stores HPKE encryption context sequence numbers as `u32` (max 2^32 - 1) rather than enforcing the RFC 9180 limit of 2^96 - 1. The overflow guard is ineffective in release builds due to Rust's default wrapping behavior for integer overflow, allowing silent counter wrap-around. This enables nonce reuse in AES-GCM, which permits plaintext recovery and authentication key compromise, a critical vulnerability.

**2. hpke-rs: Missing X25519 zero-check validation**

RFC 9180, Section 7.1.4 requires that implementations check whether the Diffie-Hellman shared secret is the all-zero value. hpke-rs omits this check in both its RustCrypto and libcrux backends. An attacker supplying low-order points can force a zero shared secret, enabling session key prediction and message decryption.

**3. libcrux-psq: AES-GCM decryption panic on malformed ciphertext**

The `decrypt_out` method calls `.unwrap()` on AES-GCM 128 decryption results, causing a process crash when presented with malformed ciphertexts. Nine of eighteen supported ciphersuites are affected. An unauthenticated attacker can trigger denial of service by sending corrupted data, rendering the entire implementation not IND-CCA secure, a critical vulnerability.

### Why RustSec advisories are needed alongside the existing GHSAs

Cryspen published GHSA-g433-pq76-6cmf for hpke-rs and GHSA-435g-fcv3-8j26 for libcrux-psq. While we appreciate that an advisory was issued, there are several gaps that leave downstream users without the information they need:

- **Severity classification**: The hpke-rs GHSA rates the nonce-reuse vulnerability as "moderate." Nonce reuse in AES-GCM enables full plaintext recovery and authentication key compromise, which is generally considered a critical or high-severity issue in cryptographic contexts. Similarly, the libcrux-psq GHSA rates the IND-CCA-breaking bug as "moderate" only.
- **Missing impact analysis**: The GHSAs does not include a details or impact section explaining what an attacker can achieve by exploiting these vulnerabilities, making it difficult for downstream consumers to assess their risk.
- **Bundled vulnerabilities**: Multiple distinct vulnerabilities with different root causes and impacts are grouped into a single advisory, which makes it harder for users to evaluate which issues affect their use case.
- **No RustSec coverage**: The GHSAs are not mirrored in RustSec, so `cargo audit` — the standard tool Rust developers rely on to check for known vulnerabilities — does not surface these issues.

### Prior discussion

These advisories were previously proposed in PR #2637, which was closed. We sought an explanation in #2646. We are resubmitting with the hope of a productive path forward. We are happy to incorporate feedback on advisory wording, severity ratings, or any other aspect of these filings, and to work with both RustSec reviewers and the upstream maintainers to ensure the advisories are accurate and fair.

Thank you for your time and for maintaining this important resource for the Rust ecosystem.

### DIFF
--- a/crates/hpke-rs/RUSTSEC-0000-0000.1.md
+++ b/crates/hpke-rs/RUSTSEC-0000-0000.1.md
@@ -10,7 +10,7 @@ keywords = ["x25519", "hpke", "validation"]
 references = ["https://datatracker.ietf.org/doc/rfc9180/"]
 
 [versions]
-patched = [">0.6.0"]
+patched = [">=0.6.0"]
 ```
 
 # Missing mandatory X25519 all-zero output validation in HPKE
--- a/crates/hpke-rs/RUSTSEC-0000-0000.md
+++ b/crates/hpke-rs/RUSTSEC-0000-0000.md
@@ -10,7 +10,7 @@ keywords = ["nonce-reuse", "integer-overflow", "hpke"]
 references = ["https://datatracker.ietf.org/doc/rfc9180/"]
 
 [versions]
-patched = [">0.6.0"]
+patched = [">=0.6.0"]
 ```
 
 # Nonce reuse via HPKE sequence number overflow
--- a/crates/libcrux-psq/RUSTSEC-0000-0000.md
+++ b/crates/libcrux-psq/RUSTSEC-0000-0000.md
@@ -10,7 +10,7 @@ keywords = ["aes-gcm", "panic", "unwrap"]
 references = ["https://github.com/cryspen/libcrux"]
 
 [versions]
-patched = [">0.0.7"]
+patched = [">=0.0.7"]
 ```
 
 # Denial of service via AES-GCM decryption panic in PSQ transport layer


crates/hpke-rs/RUSTSEC-0000-0000.1.md
```diff
@@ -0,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
```
pinkforest, line 3: You can only submit one advisory per crate in a PR with id `RUSTSEC-0000-0000`

The lint check fails because it doesn't expect 000-000**1**.md

crates/hpke-rs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
...
+id = "RUSTSEC-0000-0000"
+package = "hpke-rs"
+date = "2026-02-10"
+url = "https://github.com/cryspen/hpke-rs/pull/118"
+categories = ["crypto-failure"]
+keywords = ["nonce-reuse", "integer-overflow", "hpke"]
+references = ["https://datatracker.ietf.org/doc/rfc9180/"]
+
+[versions]
+patched = []
```
pinkforest: I thought it is fixed in 0.6.0 ?
```suggestion
patched = [">0.6.0"]
```

crates/hpke-rs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
...
+
+The hpke-rs library stores the HPKE encryption context sequence number as a `u32`, which has a maximum value of 2^32 - 1. RFC 9180 requires that the sequence number be checked against 2^(8*Nn) - 1, where Nn is the nonce length (12 bytes for all standard AEAD algorithms), yielding a maximum of 2^96 - 1.
+
+The overflow guard in `increment_seq()` compares the `u32` sequence number (cast to `u128`) against 2^96 - 1. Since a `u32` can never reach this value, the comparison is always false and the check is dead code.
+
+In debug builds, Rust's default overflow checking causes a panic when the counter wraps past 2^32 - 1. In release builds, the counter silently wraps to zero via `+=`, causing nonce reuse.
+
+Nonce reuse in AES-GCM allows plaintext recovery and leaks the GHASH authentication key, enabling universal forgeries. Nonce reuse in ChaCha20-Poly1305 enables plaintext recovery via XOR of ciphertexts encrypted under the same nonce.
+
+A fix was submitted via pull request but was not merged by the maintainer. The maintainer later copied the same fix and merged it independently, without providing credit or issuing a security advisory.
```
pinkforest: AFAIK they provided attribution and not sure re: relevancy.
```suggestion
Fix submitted was adopted in 0.6.0.
```

crates/hpke-rs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,25 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
```
pinkforest, line 3:
```suggestion
id = "RUSTSEC-0000-0000"
aliases = ["GHSA-g433-pq76-6cmf"]
```

crates/hpke-rs/RUSTSEC-0000-0000.1.md
Skgland: Multiple advisories for one crate in a single PR are possible.
The filename needs to be `RUSTSEC-0000-0000.<N>.md` where `<N>` is replaced with a number see https://github.com/rustsec/advisory-db/pull/2254/changes for example

pinkforest: Oh yes correct my bad

nadimkobeissi: @Skgland Thanks for this, I had no idea. I'll push a fix now.

crates/hpke-rs/RUSTSEC-0000-0000.md
nadimkobeissi: I'm sincerely lost on whether this is the right move, given the several existing problems with the GHSA (incorrect severity, lack of description of the vulnerability or discussion of impact/mitigation). I welcome advice.

crates/hpke-rs/RUSTSEC-0000-0000.1.md
```diff
@@ -0,0 +1,23 @@
...
+package = "hpke-rs"
+date = "2026-02-10"
+url = "https://github.com/cryspen/hpke-rs/pull/117"
+categories = ["crypto-failure"]
+keywords = ["x25519", "hpke", "validation"]
+references = ["https://datatracker.ietf.org/doc/rfc9180/"]
+
+[versions]
+patched = []
```
teor2345: This empty "patched" entry is inconsistent with the text below saying the library has been patched.

crates/hpke-rs/RUSTSEC-0000-0000.1.md
```diff
@@ -0,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "hpke-rs"
+date = "2026-02-10"
+url = "https://github.com/cryspen/hpke-rs/pull/117"
```
teor2345, line 7: Usually this would point to the fix that was actually merged, see below for a suggestion referencing the original disclosure.

crates/hpke-rs/RUSTSEC-0000-0000.1.md
```diff
@@ -0,0 +1,23 @@
...
+
+# Missing mandatory X25519 all-zero output validation in HPKE
+
+RFC 9180, Section 7.1.4 mandates that HPKE implementations "MUST check whether the Diffie-Hellman shared secret is the all-zero value and abort if so." This check is required because the X25519 function, when given certain low-order points as input (the identity element, points of small order), produces an all-zero output.
+
+The hpke-rs library's Diffie-Hellman implementation, in both the RustCrypto and libcrux provider backends, returns the raw output of the X25519 computation without performing this validation. The result flows directly into `extract_and_expand()`, which derives HPKE session keys from the shared secret.
+
+An attacker who can supply a malicious public key containing a low-order point can force the shared secret to zero, making the HPKE key schedule deterministic and predictable. The attacker can then compute the same session keys as the sender and decrypt all messages.
+
+A fix was submitted via pull request but was not merged by the maintainer. The maintainer later merged their own fix without issuing a security advisory.
```
teor2345: This text says the library was patched, which is inconsistent with an empty "patched" entry.

It also might not be necessary to go into this level of detail. Instead, you could say something more neutral:

```suggestion
This vulnerability was originally disclosed as https://github.com/cryspen/hpke-rs/pull/117
```

crates/hpke-rs/RUSTSEC-0000-0000.1.md
```diff
@@ -0,0 +1,23 @@
...
+date = "2026-02-10"
+url = "https://github.com/cryspen/hpke-rs/pull/117"
+categories = ["crypto-failure"]
+keywords = ["x25519", "hpke", "validation"]
+references = ["https://datatracker.ietf.org/doc/rfc9180/"]
+
+[versions]
+patched = []
+```
+
```
teor2345, line 15: I understand [their SECURITY.md](https://github.com/cryspen/hpke-rs/blob/f215716614ee823b3df082ed3016ddb87753f771/SECURITY.md?plain=1#L7) is very new, but it says:
> hpke-rs is still pre-release and does not support any specific versions yet.

Does this impact the severity of this issue, or justify wording something like:



```suggestion

hpke-rs is currently pre-release, and [not supported for use in production code](https://github.com/cryspen/libcrux/blob/b7187c5b115f64bf98830b7ac7980d68cbe295d6/SECURITY.md?plain=1#L5).
```

teor2345 (commented):
Some suggestions for consistency, that apply similarly to all 3 advisories.

crates/hpke-rs/RUSTSEC-0000-0000.1.md
pinkforest: If it's published in crates.io so it's covered. Severity isn't affected by things like this but it can be a mention in the advisory itself to ensure tools like cargo audit etc. traps advising that hey you should not use this in production etc. if that type of statement is still left there where people may have missed it when using the library.

Also it's important to keep in mind that when ever people started using the library they might or might not have read the said disclaimer so it should be mentioned in the advisory probably to ensure people understand the status of the library. If this disclaimer would have been from the beginning that would be another but since this is new I would include this info to make sure people understand the status.

crates/hpke-rs/RUSTSEC-0000-0000.1.md
pinkforest: Original issues / fixes get typically linked in the advisory `field = [".."]` types instead of mentioning them in the vulnerability description as it is machine readable reference.

crates/hpke-rs/RUSTSEC-0000-0000.1.md
pinkforest: Yes I think he fixed already one of them, I also should have said `">= 0.6.0"` instead of `"> 0.6.0"`

crates/hpke-rs/RUSTSEC-0000-0000.md
pinkforest: Alternatively could use related given it's semantically more correct but not really the same advisory.

```suggestion
id = "RUSTSEC-0000-0000"
related = ["GHSA-g433-pq76-6cmf"]
```

crates/hpke-rs/RUSTSEC-0000-0000.1.md
nadimkobeissi: Cryspen has actively lobbied for their “pre-release” software to be used in production in Signal, Google and OpenMLS, where it is today, and I think that speaks way more than a 0.0.x version number.

teor2345: You can see the Signal and OpenMLS dependencies here, I assume the Google ones are private (or transitive):
https://crates.io/crates/hpke-rs/reverse_dependencies

pinkforest: signal does not publish their crates in crates.io - the one signal there is a vendored-in fork who decided to publish it. for advisory-db it hasn't mattered whether there is dependency visible in reverse but only that it is published in crates.io

crates/hpke-rs/RUSTSEC-0000-0000.1.md
```diff
@@ -0,0 +1,24 @@
...
+related = ["GHSA-g433-pq76-6cmf"]
+package = "hpke-rs"
+date = "2026-02-10"
+url = "https://github.com/cryspen/hpke-rs/pull/117"
+categories = ["crypto-failure"]
+keywords = ["x25519", "hpke", "validation"]
+references = ["https://datatracker.ietf.org/doc/rfc9180/"]
+
+[versions]
+patched = [">0.6.0"]
```
CodesInChaos: I think using `>=` for `patched` is clearer

CodesInChaos:
> An attacker who can supply a malicious public key containing a low-order point can force the shared secret to zero, making the HPKE key schedule deterministic and predictable. The attacker can then compute the same session keys as the sender and decrypt all messages.

Can't the attacker compute the session key just as easily when they supply a normal public key for which they know the corresponding private key? So it's not clear to me this has a practical security impact, or if it's just non compliance with the specification.

>  An unauthenticated attacker can trigger denial of service by sending corrupted data, rendering the entire implementation not IND-CCA secure, a critical vulnerability.

The claim that an unauthenticated crash vulnerability (which is definitely bad) renders the application not IND-CCA secure because it can't respond to further queries is a rather creative interpretation of IND-CCA.

nadimkobeissi:
> The claim that an unauthenticated crash vulnerability (which is definitely bad) renders the application not IND-CCA secure because it can't respond to further queries is a rather creative interpretation of IND-CCA.

I don't understand. It's a classic distinguisher.

crates/hpke-rs/RUSTSEC-0000-0000.1.md
nadimkobeissi: Fixed for all advisories.

nadimkobeissi:
Advisories have been merged in #2667 for some of the issues here, but notably not for the hpke-rs issues which include catastrophic nonce reuse.

nadimkobeissi:
Given that this PR was closed without explanation or feedback, I'm assuming it's because some of its advisories were already submitted in #2667. I'll open another PR with just the missing advisories.

PR #2661, opened 2026-02-22
Add advisory for xml-rs: unbounded entity expansion (Billion Laughs)

BrianMcWilliams:
Advisory for xml-rs: Unbounded Entity Expansion (Billion Laughs / XML Bomb, CWE-776)

xml-rs expands DTD entity references recursively with no depth or size limit. A sub-1KB payload can cause exponential memory consumption (~3M amplification ratio).

No patch exists. Alternative: quick-xml (does not expand entities by default).

### DIFF
--- a/crates/xml-rs/RUSTSEC-0000-0000.md
+++ b/crates/xml-rs/RUSTSEC-0000-0000.md
@@ -1,0 +1,47 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "xml-rs"
+date = "2026-02-22"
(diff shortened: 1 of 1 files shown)


djc:
@kornelski any thoughts on this advisory?

fintelia:
The linked issue is titled "Profile-Guided Optimization (PGO) results" and doesn't seem to have anything to do with this advisory

kornelski:
xml-rs has protection against this attack.

The report by @BrianMcWilliams is an LLM hallucinated spam.

PR #2663, opened 2026-02-22
Add advisory for reqwest: SSRF via default redirect policy

BrianMcWilliams:
Advisory for reqwest: SSRF via default redirect policy (CWE-918)

The default redirect policy follows redirects to private/internal IPs including cloud metadata endpoints (169.254.169.254). Includes mitigation code example.

### DIFF
--- a/crates/reqwest/RUSTSEC-0000-0000.md
+++ b/crates/reqwest/RUSTSEC-0000-0000.md
@@ -1,0 +1,45 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "reqwest"
+date = "2026-02-22"
+url = "https://github.com/seanmonstar/reqwest/issues/2260"
+categories = ["denial-of-service"]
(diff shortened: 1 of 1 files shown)


djc:
@seanmonstar are you okay with this getting published? Would be nice to include a patched version.

seanmonstar:
Definitely not.

This will just alert people with nothing to upgrade. Many won't need a change at all, leading to alert fatigue. And it will cause many to raise issues asking me "why it isn't fixed yet".

This isn't the right place to file, it should have been an issue on the reqwest repo.

Even then, I do not see this as a vulnerability in reqwest. The default policy is perfectly fine in many cases. When using user input in a server setting, it is always wise to add proper protections. It's a common redirect policy that is not a vulnerability in many other languages and clients.

CodesInChaos:
Attempting to prevent SSRF via a Redirect policy doesn't make sense. Since:

1. The SSRF could already happen on the initial request, no redirect required
2. You cannot reliably tell if a url will resolve to a private IP. So the policy can't reliably reject these.

PR #2665, opened 2026-02-23
Add advisory: reqwest SSRF via default redirect policy

BrianMcWilliams:
reqwest's default redirect policy follows redirects to private/internal IP ranges (RFC 1918, link-local, loopback, cloud metadata endpoints), enabling SSRF in server-side applications that fetch user-controlled URLs.


alex:
Is there a change in the status since https://github.com/rustsec/advisory-db/pull/2663?

PR #2683, opened 2026-02-27
RustSec Advisories for hpke-rs

nadimkobeissi:
This PR proposes three RustSec advisories for some of the cryptographic vulnerabilities discovered during research (["Verification Theatre: False Assurance in Formally Verified Cryptographic Libraries"](https://eprint.iacr.org/2026/192)), namely the ones for the hpke-rs crate, which included the disclosure of a critical nonce reuse vulnerability. The upstream maintainers (Cryspen) have patched these issues but have not filed RustSec advisories, which means `cargo audit` does not currently flag affected versions.

Cryspen has published GitHub Security Advisories (GHSAs) for some of these issues. However, we believe RustSec advisories are still warranted for the reasons outlined below.

### Proposed Advisories

**1. hpke-rs: Nonce reuse via sequence number overflow**

The library stores HPKE encryption context sequence numbers as `u32` (max 2^32 - 1) rather than enforcing the RFC 9180 limit of 2^96 - 1. The overflow guard is ineffective in release builds due to Rust's default wrapping behavior for integer overflow, allowing silent counter wrap-around. This enables nonce reuse in AES-GCM, which permits plaintext recovery and authentication key compromise, a critical vulnerability.

**2. hpke-rs: Missing X25519 zero-check validation**

RFC 9180, Section 7.1.4 requires that implementations check whether the Diffie-Hellman shared secret is the all-zero value. hpke-rs omits this check in both its RustCrypto and libcrux backends. An attacker supplying low-order points can force a zero shared secret, enabling session key prediction and message decryption.

### Why RustSec advisories are needed alongside the existing GHSAs

Cryspen published GHSA-g433-pq76-6cmf for hpke-rs. While we appreciate that an advisory was issued, there are several gaps that leave downstream users without the information they need:

- **Severity classification**: The hpke-rs GHSA rates the nonce-reuse vulnerability as "moderate." Nonce reuse in AES-GCM enables full plaintext recovery and authentication key compromise, which is generally considered a critical or high-severity issue in cryptographic contexts.
- **Missing impact analysis**: The GHSA does not include a details or impact section explaining what an attacker can achieve by exploiting these vulnerabilities, making it difficult for downstream consumers to assess their risk.
- **No RustSec coverage**: The GHSA is not mirrored in RustSec, so `cargo audit` — the standard tool Rust developers rely on to check for known vulnerabilities — does not surface these issues.

### Prior discussion

These advisories were previously proposed in PR #2637, which was closed. We sought an explanation in #2646. Another PR, #2647, was also closed after Cryspen submitted their own RustSec advisories for libcrux-psq, but the advisories for hpke-rs remain missing, hence this PR.

Thank you for your time and for maintaining this important resource for the Rust ecosystem.

### DIFF
--- a/crates/hpke-rs/RUSTSEC-0000-0000.md
+++ b/crates/hpke-rs/RUSTSEC-0000-0000.md
@@ -24,4 +24,4 @@ In debug builds, Rust's default overflow checking causes a panic when the counte
 
 Nonce reuse in AES-GCM allows plaintext recovery and leaks the GHASH authentication key, enabling universal forgeries. Nonce reuse in ChaCha20-Poly1305 enables plaintext recovery via XOR of ciphertexts encrypted under the same nonce.
 
-A fix was submitted via pull request but was not merged by the maintainer. The maintainer later copied the same fix and merged it independently, without providing credit or issuing a security advisory.
+This vulnerability was originally disclosed as https://github.com/cryspen/hpke-rs/pull/TODO PR NUMBER HERE


crates/hpke-rs/RUSTSEC-0000-0000.1.md
```diff
@@ -0,0 +1,25 @@
+```toml
```
teor2345, line 1: > This PR proposes three RustSec advisories

I can only see two advisories in the PR diff, but the PR description says three, is this a copy-paste oversight?

crates/hpke-rs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,27 @@
...
+
+The hpke-rs library stores the HPKE encryption context sequence number as a `u32`, which has a maximum value of 2^32 - 1. RFC 9180 requires that the sequence number be checked against 2^(8*Nn) - 1, where Nn is the nonce length (12 bytes for all standard AEAD algorithms), yielding a maximum of 2^96 - 1.
+
+The overflow guard in `increment_seq()` compares the `u32` sequence number (cast to `u128`) against 2^96 - 1. Since a `u32` can never reach this value, the comparison is always false and the check is dead code.
+
+In debug builds, Rust's default overflow checking causes a panic when the counter wraps past 2^32 - 1. In release builds, the counter silently wraps to zero via `+=`, causing nonce reuse.
+
+Nonce reuse in AES-GCM allows plaintext recovery and leaks the GHASH authentication key, enabling universal forgeries. Nonce reuse in ChaCha20-Poly1305 enables plaintext recovery via XOR of ciphertexts encrypted under the same nonce.
+
+A fix was submitted via pull request but was not merged by the maintainer. The maintainer later copied the same fix and merged it independently, without providing credit or issuing a security advisory.
```
teor2345: I'd encourage you to use the same wording as the entry above - just a neutral reference to the original disclosure.


```suggestion
This vulnerability was originally disclosed as https://github.com/cryspen/hpke-rs/pull/TODO PR NUMBER HERE
```

crates/hpke-rs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,27 @@
...
+patched = [">=0.6.0"]
+```
+
+# Nonce reuse via HPKE sequence number overflow
+
+The hpke-rs library stores the HPKE encryption context sequence number as a `u32`, which has a maximum value of 2^32 - 1. RFC 9180 requires that the sequence number be checked against 2^(8*Nn) - 1, where Nn is the nonce length (12 bytes for all standard AEAD algorithms), yielding a maximum of 2^96 - 1.
+
+The overflow guard in `increment_seq()` compares the `u32` sequence number (cast to `u128`) against 2^96 - 1. Since a `u32` can never reach this value, the comparison is always false and the check is dead code.
+
+In debug builds, Rust's default overflow checking causes a panic when the counter wraps past 2^32 - 1. In release builds, the counter silently wraps to zero via `+=`, causing nonce reuse.
```
teor2345: Technically it's just the default release config that disables overflow checks. Some projects' release workflows could still turn them on, but that's unlikely in cryptographic projects due to performance.

I'm not sure if that's worth mentioning in detail here, maybe just a tweak like this:
```suggestion
In debug builds, Rust's default overflow checking causes a panic when the counter wraps past 2^32 - 1. In default release builds, the counter silently wraps to zero via `+=`, causing nonce reuse.
```

teor2345 (commented):
Some suggestions / questions to clarify these advisories

djc:
Closing this since the OP has been blocked.

daira:
> Closing this since the OP has been blocked.

This is an incredible abdication of responsibility to users of software that depends on these libraries. 😕

No attempt to argue that any of what @nadimkobeissi has said is inaccurate. Just "we don't like him", basically. The Rust community sees what is happening here.

djc:
> This is an incredible abdication of responsibility to users of software that depends on these libraries. 😕

I have asked the maintainers multiple times to submit advisories explaining the issues here. I don't think I am the one who has been abdicating responsibility.

The maintainers have finally submitted a PR with advisories for these issues:

- https://github.com/rustsec/advisory-db/pull/2742

PR #2684, opened 2026-02-27
report vulnerabilities in aho-corasick crate

YuxiLing:
Add two cves: CVE-2025-70350 and CVE-2025-70351. CVE IDs are assigned, but details are not updated online.

### DIFF
--- a/crates/aho-corasick/RUSTSEC-0000-0000.md
+++ b/crates/aho-corasick/RUSTSEC-0000-0000.md
@@ -1,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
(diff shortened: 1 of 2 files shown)


djc:
Those CVE IDs seem to be for very different things; I cannot find any public record of the vulnerabilities these advisories describe.

cc @BurntSushi

BurntSushi:
This is the first I'm hearing of these bugs (to the best of my recollection).

But these definitely aren't vulnerabilities or CVEs. These are likely best addressed by documenting panicking preconditions. Notice that in both cases the panics are provoked by deliberately providing incorrect offsets.

djc:
Thanks for the quick response!

PR #2685, opened 2026-02-27
Advisory for nonce reuse in a Sigma protocol in `vrf_fun`

rot256:
Verified by the maintainer (@LLFourn) of https://github.com/LLFourn/secp256kfun


djc:
Looks like there's a patched version now in 0.12.1. @LLFourn can you confirm your okay with us publishing this advisory?

PR #2690, opened 2026-03-03
`time_calibrators`: this time, it's plural

LawnGnome:


djc:
Since I prefer to rebase commits, would nice to keep commit messages a little more factual.

PR #2697, opened 2026-03-05
pingora-cache: add CVE-2026-2836

drcaramelsyrup:
Also published at https://github.com/cloudflare/pingora/security/advisories/GHSA-f93w-pcj3-rggc.


djc:
Thanks!

Suggest also adding the GHSA to aliases.

PR #2705, opened 2026-03-09
Add advisory for `sd-notify`'s `remove_var` in safe function

tbu-:


djc:
Asked for maintainer permission in the linked upstream issue.

LawnGnome:
@djc Do you have any concerns here? (Technically, the maintainer more shrugged than approved, but I think we might as well merge this.)

djc:
I was hoping the submitter would add the patched version before we merge this, but could definitely have been more explicit about that!

lnicola:
> Technically, the maintainer more shrugged than approved

I'm not really familiar with the RUSTSEC workflow, can I really prevent the advisory from being filed? If not, then it's a ping, not a question.

crates/sd-notify/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
...
+
+[versions]
+patched = []
+```
+
+# Unsafe removal of environment variables
+
+Affected versions of this crate expose unsetting environment variables in safe functions. Since unsetting environment variables is thread-unsafe on Linux (see [documentation of `std::env::remove_var`](https://doc.rust-lang.org/std/env/fn.remove_var.html)), this is unsound.
+
+There's a discussion of this issue at [sd-notify#9](https://github.com/lnicola/sd-notify/issues/9) and a proposed fix at [sd-notify#13](https://github.com/lnicola/sd-notify/pull/13).
```
lnicola:
```suggestion
There's a discussion of this issue at [sd-notify#9](https://github.com/lnicola/sd-notify/issues/9) and a merged patch introducing new APIs at [sd-notify#13](https://github.com/lnicola/sd-notify/pull/13).
```

crates/sd-notify/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,31 @@
...
+os = ["linux"]
+
+[affected.functions]
+"sd_notify::listen_fds" = ["< 0.5.0"]
+"sd_notify::listen_fds_with_names" = ["< 0.5.0"]
+"sd_notify::notify" = ["< 0.5.0"]
+"sd_notify::watchdog_enabled" = ["< 0.5.0"]
+
+[versions]
+patched = []
```
lnicola: Should this say `>= 0.5.0` or something?

djc:
> I'm not really familiar with the RUSTSEC workflow, can I really prevent the advisory from being filed? If not, then it's a ping, not a question.

You can make technical arguments that an advisory is inappropriate for this issue and that could lead to us not publishing the advisory. So yes, I do consider it an actual question where we're inviting your thoughts on whether an advisory is a good fit for the issue in your crate.

lnicola:
I see, thanks.

I accept that using `set_var` in multithreaded programs is unsound. However, I'm not convinced it's too problematic for `sd-notify`, especially since pinging systemd is something you can do:

 - once on start-up, which you'll do either before starting any threads (then it's sound), or after waiting for them to be ready; background threads can still change the environment, but I feel it's less common after start-up
 - periodically (e.g. for the watchdog), but then you can't use the problematic `unset_env` parameter

In addition, while Rust had a couple of CVEs assigned to the standard library, there wasn't one for `std::env::{set_var,remove_var}`. I think `sd-notify` is similar to `dotenv` / `dotenvy` (something you do at start-up). `dotenvy::dotenv` is not unsafe, and there's no RUSTSEC advisory for it, so the precedent seems to be not to file advisories. But I'm not certain about it (maybe there are 1000 advisories for `set_var` usage in other crates), which is why I said "do as you see fit".

PS: the only ones I can spot are `RUSTSEC-2020-0071` and `RUSTSEC-2020-0159` (`time` and `chrono`). But you're more likely to use those crates from multiple threads compared to `sd-notify`. They're also more likely to be used from a library, while `sd-notify` is something only binaries will normally use.

tbu-:
> * once on start-up, which you'll do either before starting any threads (then it's sound), or after waiting for them to be ready; background threads can still change the environment, but I feel it's less common after start-up

The trouble is that it's unsafe to do in a function marked with `#[tokio::main]`, because that is after there are multiple threads. Having a function marked with `#[tokio::main]` isn't uncommon in async Rust programs.

lnicola:
> The trouble is that it's unsafe to do in a function marked with `#[tokio::main]`

Only if the other threads are accessing the environment at the same time. And if the main thread isn't starting any tasks, nothing is running on those other threads.

crates/sd-notify/RUSTSEC-0000-0000.md
tbu-: Done.

crates/sd-notify/RUSTSEC-0000-0000.md
tbu-: Done.

tbu-:
> Only if the other threads are accessing the environment at the same time.

I don't think tokio provides a guarantee for that, and if they don't then it's going to be hard to audit and continuously verify.

But it's an interesting idea. One could ask the tokio folks if they want to provide such a guarantee. I think it'd be useful for the async Rust ecosystem.

lnicola:
As far as I can tell, there's no environment access done in the `tokio` repo except in `loom`'s `num_cpus`.

So just to be clear, you're worried that in code like:

```rust

use sd_notify::NotifyState;
use tokio::net::TcpListener;

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let listener = TcpListener::bind("127.0.0.1:8080").await?;
    
    sd_notify::notify(&[NotifyState::Ready])?;

    loop {
        listener.accept().await?;

        tokio::spawn(async move {
            print!("New connection");
        });
    }
}
```

At the point where `sd_notify::notify` is called, `tokio` will access the environment?

tbu-:
I'm worried that tokio does not provide a guarantee that no environment access happens in that code at the point of `sd_notify::notify(&[NotifyState::Ready])?;`. This is different from the worry that `tokio` has environment access at this point in version 1.50.0.

lnicola:
Given that:

 - `tokio` takes great care to not access the environment (besides checking for `TOKIO_WORKER_THREADS` while building the runtime) in version 1.50
 - at that point, there is no work for the worker threads to do

I'm pretty confident that the code above is sound.

lnicola:
To add, if you were right, there would be no sound way to use either `sd-notify`, `dotenv` or `tracing_subscriber`'s env filters from under `#[tokio::main]`. This is something that `tokio` itself does in https://github.com/tokio-rs/tokio/blob/ee04abe7970b769439e2fecd4b1dd8790e4234e1/examples/chat.rs#L49. `hyper` does it too in https://github.com/hyperium/hyper/blob/master/examples/hello.rs. `axum` does it in https://github.com/tokio-rs/axum/blob/main/examples/tracing-aka-logging/src/main.rs. Most `tokio` apps I've seen initialize their logging like that (`hickory` being a notable exception).

PS: even docs.rs: https://github.com/rust-lang/docs.rs/blob/main/crates/bin/docs_rs_web/src/main.rs

tbu-:
To figure out whether it's safe, you'd have to ask the tokio folks for guarantees. And I think it'd be great if they could provide that guarantee. If you think this is a pattern that's supposed to be safe, please ask in tokio-related communication channels. I cannot provide that guarantee.

One thing I think is missing from your analysis is that _I think_ it's not guaranteed that the worker threads have started up completely by the point you're executing `sd_notify::notify(&[NotifyState::Ready])?;`. As far as I can tell, this means you'd need to have guarantees from the Rust standard library and the libc that nobody touches the environment on thread startup. I don't think the Rust standard library wants to give that guarantee, given the safety documentation at [`std::env::remove_var`](https://doc.rust-lang.org/std/env/fn.remove_var.html):
> All this makes it not practically possible for you to guarantee that no other thread will read the environment, so the only safe option is to not use `set_var` or `remove_var` in multi-threaded programs at all.

djc:
> To add, if you were right, there would be no sound way to use either `sd-notify`, `dotenv` or `tracing_subscriber`'s env filters from under `#[tokio::main]`. This is something that `tokio` itself does in https://github.com/tokio-rs/tokio/blob/ee04abe7970b769439e2fecd4b1dd8790e4234e1/examples/chat.rs#L49. `hyper` does it too in https://github.com/hyperium/hyper/blob/master/examples/hello.rs. `axum` does it in https://github.com/tokio-rs/axum/blob/main/examples/tracing-aka-logging/src/main.rs. Most `tokio` apps I've seen initialize their logging like that (`hickory` being a notable exception).
> 
> PS: even docs.rs: https://github.com/rust-lang/docs.rs/blob/main/crates/bin/docs_rs_web/src/main.rs

Are you arguing that these programs all modify the environment (via `set_var()` or `remove_var()`) inside `#[tokio::main]`? I don't think that's the case.

So to be clear, I'd argue that the problem here is with modifying the environment after multiple threads have been started. If that requirement was clearly documented before, I think it might be okay, but I definitely think marking these functions as `unsafe` is the better way to go. Even if most users would do it at the top of `#[tokio::main]` before worker threads are doing any work, I think it might be (have been) too easy to misuse in a way that is unsafe.

lnicola:
> Are you arguing that these programs all modify the environment

`tracing_subscriber` and the linked examples read, not modify, it, but `tokio` also doesn't explicitly guarantee that it doesn't modify the environment. `dotenvy` modifies it, of course. `sd-notify` modifies it if you ask for it.

> I definitely think marking these functions as unsafe is the better way to go

They are unsafe as of 0.5.0. Previous versions indeed don't mention the threading gotcha, but the parameter is documented as modifying the environment.

> I think it might be (have been) too easy to misuse

My feeling is that this is in part mitigated by the fact that you either call it on start-up, or you call it multiple times (which will be sound).

tbu-:
Thanks for creating the question issue for tokio. :)

> `dotenvy` modifies it, of course.

AFAICT, that will change in dotenvy 0.16 for the same reason it was changed in `sd-notify`: https://github.com/allan2/dotenvy/issues/112.

djc:
> AFAICT, that will change in dotenvy 0.16 for the same reason it was changed in sd-notify: allan2/dotenvy#112.

To be clear, that issue last made progress 2 years ago, so I'm not holding my breath.

djc:
> They are unsafe as of 0.5.0. Previous versions indeed don't mention the threading gotcha, but the parameter is documented as modifying the environment.

Sure. So the question here is, do we regard the previous behavior in 0.4.x, where `listen_fds()` (arguably the library's main entry point) silently modified the environment, as a potential footgun for unsuspecting callers? I'd say it's worth an advisory.

Darksonn:
I don't think Tokio's guarantees has any influence on whether or not an advisory should be filed. What Tokio says doesn't change anything here - for example what about `#[async_std::main]`? I'm sure there are many other crates like this.

lnicola:
> I don't think Tokio's guarantees has any influence on whether or not an advisory should be filed.

I think it was the original example and main concern put forward by tpu-.

> I'm sure there are many other crates like this.

A lot of things must be true to cause real issues:

 - you must enable the `unset_env` argument, meaning you only call the function once
 - you do this _after_ starting other threads
 - those threads must access the environment
 - those threads must do so from C code, which already requires `unsafe`

Given that, and since no advisory was released for the standard library (unlike `fs::remove_dir_all`, where other language implementations didn't even bother anyway) or other crates that are meant to be used in the same way, I don't think one is needed in this case.

But those are all the arguments I can make, and you don't seem to find them convincing, so :shrug:.

Darksonn:
I mean, I don't necessarily think an advisory is needed. :woman_shrugging:

djc:
Alright -- thanks for all the discussion everyone!

PR #2707, opened 2026-03-09
Add blog post urls to RUSTSEC-2026-{0033-0035}

drcaramelsyrup:
The blog adds more detail regarding how the vulnerabilties work for the pingora-core and pingora-cache RUSTSECs.

Mostly for completeness / informative value.

(I noticed that the GHSA advisory numbers had been added to the aliases, thanks!)


amousset:
Thanks for the update!

PR #2711, opened 2026-03-10
Add advisory for the malicious `chrono_anchor` crate

LawnGnome:
The blog post technically hasn't been published yet as I write this PR description, but it should be up shortly, I've seen a draft, and I'm comfortable linking to it.


amousset:
It seems to have been published.

PR #2724, opened 2026-03-17
Fix incomplete affected functions for RUSTSEC-2026-0041

Marcono1234:
See also https://github.com/PSeitz/lz4_flex/issues/213#issuecomment-4075896791

CC @PSeitz

(Question to the rustsec maintainers mostly) What exact effect does `affected.functions` have though, and can it cause false negatives? E.g. when someone uses an ancient lz4_flex version (quite unlikely though) where the API was different, would that be erroneously be considered safe? Maybe the implementation back then was actually not affected by this vulnerability, though might be better to err on the safe side and flag usage of such old versions as vulnerable?

### DIFF
--- a/crates/lz4_flex/RUSTSEC-2026-0041.md
+++ b/crates/lz4_flex/RUSTSEC-2026-0041.md
@@ -11,6 +11,14 @@ cvss = "CVSS:4.0/AV:N/AC:L/AT:P/PR:N/UI:N/VC:H/VI:N/VA:N/SC:N/SI:N/SA:N"
 [affected.functions]
 "lz4_flex::block::decompress_into" = ["<= 0.11.5", "= 0.12.0"]
 "lz4_flex::block::decompress_into_with_dict" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::block::decompress" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::block::decompress_with_dict" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::block::decompress_size_prepended" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::block::decompress_size_prepended_with_dict" = ["<= 0.11.5", "= 0.12.0"]
+# Deprecated re-exports
+"lz4_flex::decompress" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::decompress_into" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::decompress_size_prepended" = ["<= 0.11.5", "= 0.12.0"]
 
 [versions]
 patched = [">= 0.11.6, < 0.12.0", ">= 0.12.1"]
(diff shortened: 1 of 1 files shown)


crates/lz4_flex/RUSTSEC-2026-0041.md
```diff
@@ -28,13 +36,12 @@ the initialized portion of the output buffer.
 
 Two scenarios are affected:
 
-- Decompressing with the `unsafe` implementation (`safe-decode` feature flag disabled, which
-  is the default): can leak content of uninitialized memory as part of the decompressed result.
+- Decompressing with the `unsafe` implementation (`safe-decode` feature flag disabled; flag is
+  enabled by default): can leak content of uninitialized memory as part of the decompressed result.
```
Marcono1234, line 40: Reworded this to make it (hopefully) clearer; "which is the default" might have been ambiguous, e.g. whether it means "the flag is enabled by default" or "the flag being disabled is the default".

I hope that is ok.

crates/lz4_flex/RUSTSEC-2026-0041.md
```diff
@@ -11,6 +11,14 @@ cvss = "CVSS:4.0/AV:N/AC:L/AT:P/PR:N/UI:N/VC:H/VI:N/VA:N/SC:N/SI:N/SA:N"
...
 "lz4_flex::block::decompress_into" = ["<= 0.11.5", "= 0.12.0"]
 "lz4_flex::block::decompress_into_with_dict" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::block::decompress" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::block::decompress_with_dict" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::block::decompress_size_prepended" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::block::decompress_size_prepended_with_dict" = ["<= 0.11.5", "= 0.12.0"]
+# Deprecated re-exports
+"lz4_flex::decompress" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::decompress_into" = ["<= 0.11.5", "= 0.12.0"]
+"lz4_flex::decompress_size_prepended" = ["<= 0.11.5", "= 0.12.0"]
```
Marcono1234, line 21: I hope I did not overlook something here. If possible please double-check.

djc:
> (Question to the rustsec maintainers mostly) What exact effect does `affected.functions` have though, and can it cause false negatives? E.g. when someone uses an ancient lz4_flex version (quite unlikely though) where the API was different, would that be erroneously be considered safe? Maybe the implementation back then was actually not affected by this vulnerability, though might be better to err on the safe side and flag usage of such old versions as vulnerable?

I don't know how various tools consume this information.

Marcono1234:
@PSeitz, do you think this PR here is obsolete now due to your #2725?
Or should the re-exports be listed as well?

Side note: To the user these re-exports are not actually reported as deprecated due to https://github.com/rust-lang/rust/issues/30827.

Marcono1234:
Going to close this because as mentioned above this was mostly / completely addressed by #2725.

PR #2727, opened 2026-03-19
Add retrospective advisory for several malicious report by phylum in …

GrassedgeT:
These advisories are used to retrospectively document the 7 malware packages reported by phylum in 2023.
Related [blog](https://www.veracode.com/blog/rust-malware-staged-on-crates-io/).

Since these incidents occurred in 2023, and the largest RUSTSEC advisory number for 2023 was 96, I started numbering with 97.


djc:
These need to all be draft advisories named 0000-0000; they will automatically get an identifier after merge.

GrassedgeT:
Ok, done

PR #2777, opened 2026-04-11
Add unsoundness advisory for rand with a custom logger

dhardy:
See https://github.com/rust-random/rand/pull/1763


djc (approved):
Thanks for the report!

Can you add a header to address the linter concern?

PR #2785, opened 2026-04-14
Add webpki URI name constraints bug

ctz:


crates/rustls-webpki/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rustls-webpki"
+date = "2026-04-14"
+keywords = ["name constraints", "x509"]
+aliases = ["GHSA-965h-392x-2mh5"]
+
+[versions]
+patched = [">= 0.103.12"]
```
djc: I think we should clarify the situation of 0.104.x betas here.

crates/rustls-webpki/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
...
+
+# Name constraints for URI names were incorrectly accepted
+
+Name constraints for URI names were ignored and therefore accepted. 
+
+Note this library does not provide an API for asserting URI names, and URI name constraints are otherwise not implemented.  URI name constraints are now rejected unconditionally.
+
+Since name constraints are restrictions on otherwise properly-issued certificates, this bug is reachable only after signature verification and requires misissuance to exploit.
+
+This vulnerability is identified as [GHSA-965h-392x-2mh5](https://github.com/rustls/webpki/security/advisories/GHSA-965h-392x-2mh5). Thank you to @1seal for the report.
```
djc, line 21: I think like last time, the GHSA should start with our description rather than the original report.

djc (commented):
Same feedback for the other PR...

crates/rustls-webpki/RUSTSEC-0000-0000.md
ctz: I will blat the GHSA with the information here before publication.

djc:
Looks like the version ranges need some finessing...

ctz:
Do version ranges work with prereleases? The error message doesn't make sense to me.

ctz:
OK, apparently this edifice has a different interpretation of prereleases to cargo. So what was there before matches all prereleases.

djc:
> So what was there before matches all prereleases.

But that's not what we want, right?

crates/rustls-webpki/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,21 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "rustls-webpki"
+date = "2026-04-14"
+keywords = ["name constraints", "x509"]
+aliases = ["GHSA-965h-392x-2mh5"]
+
+[versions]
+patched = [">= 0.103.12, < 0.104.0-alpha.5", ">= 0.104.0-alpha.6"]
```
djc: Sorry -- I think this is still wrong?

This seems to claim that 0.104.0-alpha.{1-4} are patched, alpha 5 is not patched, then alpha 6+ are patched? I think you want to say that `>= 0.103.12, < 0.104.0-alpha.1` are patched, as well as `>= 0.104.0-alpha.6`.

ctz:
Thanks for the patience!

PR #2791, opened 2026-04-15
advisory linter requires affected functions to start with package rather than crate name

MartinNowak:
- allow to override crate name derived from crates.io package name
- fixes rustsec/rustsec#1575 (also see https://github.com/rustsec/rustsec/pull/1582)


djc:
I don't think this is the right approach and would prefer to wait for a mechanism that we can use to automatically get this fixed in the tooling rather than the advisories.

PR #2795, opened 2026-04-17
Update rand advisory to include an additional patched version.

nwalfield:
The fix was backported to the 0.8 branch and released as version 0.8.6.  Update the advisory to indicate that this version is also safe.

For details, please see https://github.com/rust-random/rand/pull/1772 and https://github.com/rust-random/rand/issues/1770 .


djc (approved):

nwalfield:
Thanks @djc for the quick reaction.

djc:
Thanks!

(Separately I'm curious to hear more about why you need OpenSSL support for Hickory in Sequoia -- a quick search found https://gitlab.com/sequoia-pgp/sequoia/-/work_items/1133 but that seems to be talking about the other direction...)

nwalfield:
The short answer is that some of our downstreams prefer OpenSSL to AWS-LC or ring.  @decathorpe documented the Fedora and RHEL position here quite well, I think: https://github.com/hickory-dns/hickory-dns/issues/3452 .  (I'm happy to continue the discussion, but perhaps this is not the right place.)

djc:
(I'm djc on GitLab as well if you want to tag me in an issue over there.)

xtqqczze:
> Affected versions of rand are >= 0.7, < 0.9.3 and 0.10.0.

This part needs updating too: https://github.com/rustsec/advisory-db/pull/2796

PR #2805, opened 2026-04-20
Add thin-vec GHSA-xphw-cqx3-667j

kpcyrd:
- https://github.com/mozilla/thin-vec/security/advisories/GHSA-xphw-cqx3-667j
- https://alioth-lists.debian.net/pipermail/pkg-rust-maintainers/2026-April/074118.html


djc (approved):
Thanks, looks good!

PR #2810, opened 2026-04-22
core2: unmaintained

### LINKED ISSUE #2788: `core2` unmaintained
Some context: https://hachyderm.io/@piecritic/116404790999620656

It seems all releases were yanked which is causing quite a bit of disruption.

The repo is archived: https://github.com/bbqsrc/core2/

Alternatives:
- [`embedded-io`](https://crates.io/crates/embedded-io) solves the same general problem
- [`no-std-io2`](https://crates.io/crates/no-std-io2) is a maintained fork

I'll note it's (currently) a hard dependency of `libflate` and appears to be part of the public API(?). So an advisory has the potential to be quite disruptive in addition to the yanking. Though with all the releases yanked, unless they get unyanked, it seems they will need to find a solution.

[Reddit thread](https://www.reddit.com/r/rust/comments/1slnga4/core2_yanked_millions_effected/)

djc:
Fixes #2788.

Not sure whether it's useful to mention the Reddit thread in here?

### DIFF
--- a/crates/core2/RUSTSEC-0000-0000.md
+++ b/crates/core2/RUSTSEC-0000-0000.md
@@ -1,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "core2"
+date = "2026-04-14"
+url = "https://hachyderm.io/@piecritic/116404790999620656"
+informational = "unmaintained"
+
+[versions]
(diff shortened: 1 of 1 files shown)


tarcieri (approved):
The Reddit thread (which was removed by moderators) is probably not helpful to mention

PR #2813, opened 2026-04-22
Add hickory-recursor report for GHSA-83hf-93m4-rgwq

cpu:
Add an advisory for the `hickory-recursor` crate, for [GHSA-83hf-93m4-rgwq](https://github.com/hickory-dns/hickory-dns/security/advisories/GHSA-83hf-93m4-rgwq).

Note that this is my first time filing a rustsec advisory myself, please let me know if I've messed anything up.

### DIFF
--- a/crates/hickory-recursor/RUSTSEC-0000-0000.md
+++ b/crates/hickory-recursor/RUSTSEC-0000-0000.md
@@ -1,0 +1,47 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "hickory-recursor"
+date = "2026-04-22"
+url = "https://github.com/hickory-dns/hickory-dns/security/advisories/GHSA-83hf-93m4-rgwq"
+categories = ["privilege-escalation"]
+keywords = ["dns", "cache-poisoning"]
+license = "CC-BY-4.0"
+[versions]
+patched = []
+```
+
+# Record cache accepts AUTHORITY section NS from sibling zone via parent-pool zone-context elevation
+
+The Hickory DNS project's experimental `hickory-recursor` crate's record cache
+(`DnsLru`) stores records from DNS responses keyed by each record's own (name,
+type), not by the query that triggered the response. `cache_response()` in
+`crates/recursor/src/lib.rs` chains `ANSWER`, `AUTHORITY`, and `ADDITIONAL`
+sections into one record iterator before insertion. The bailiwick filter it
+applies uses the zone context of the NS pool that serviced the lookup, not the
+zone being queried.
+
+This creates a cross-zone poisoning path. When Hickory builds the NS pool for
+`attacker.poc.` it uses the parent `poc.` `NS` pool (`ns.zone() = "poc."`). If
+the `poc.` nameserver under the attacker's control includes in its response's
+`AUTHORITY` section a record for a sibling zone like `victim.poc. NS
+ns.evil.poc.`, the bailiwick check `is_subzone("poc.", "victim.poc.")` passes
+(`victim.poc.` is a subdomain of `poc.`). The record is stored under
+`(victim.poc., NS)` in the shared cache.
+
+Subsequently, any client querying a name in `victim.poc`. causes Hickory to
+build its NS pool from the poisoned cache entry, routing queries to the
+attacker's nameserver (`ns.evil.poc.`) rather than to the legitimate nameserver
+for `victim.poc.`. The legitimate `NS` for that zone receives zero queries.
+
+This issue is fixed in `hickory-resolver` 0.26.0 with the `recursor` feature
+through an architectural change to response-level caching: responses are stored
+keyed by the originating query `(name, type)`. A response to `(attacker.poc.
+NS)` is stored only under that key and cannot affect the `(victim.poc., NS)`
+cache entry.
+
+We believe this issue has been present in all published versions of the
+experimental `hickory-recursor` crate, which has now been folded into the
+`hickory-resolver` crate under the non-default `recursor` feature flag. The
+`hickory-recursor` crate will not receive any updates going forward and all
+users should migrate to `hickory-resolver` with the `recursor` feature.


crates/hickory-recursor/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,33 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+
```
djc: Nit: please drop the empty lines between all the metadata fields.

crates/hickory-recursor/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,33 @@
...
+url = "https://github.com/hickory-dns/hickory-dns/security/advisories/GHSA-83hf-93m4-rgwq"
+
+categories = ["privilege-escalation"]
+
+keywords = ["dns", "cache-poisoning"]
+
+license = "CC-BY-4.0"
+
+[versions]
+patched = [">= 0.26.0"]
```
djc: Since there is no hickory-recursor 0.26.0, I think we should just leave this empty. The correct response is to migrate to the recursor that lives in hickory-resolver 0.26.0+ (should probably make that explicit).

crates/hickory-recursor/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,33 @@
...
+
+unaffected = ["< 0.22"]
```
djc: Where does this come from? hickory-recursor had a 0.1.0 version and the next version after that is 0.24.0.

crates/hickory-recursor/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+# Record Cache Accepts AUTHORITY-Section NS from Sibling Zone via Parent-Pool Zone-Context Elevation
```
djc: Nit: personally the Title Case Stuff Always Feels Pretty Shouty, but that might be a cultural thing?

crates/hickory-recursor/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,33 @@
...
+
+Prior to 0.26, Hickory DNS's experimental recursor crate's record cache (DnsLru) stores records from DNS responses keyed by each record's own (name, type), not by the query that triggered the response. cache_response() in crates/recursor/src/lib.rs chains ANSWER, AUTHORITY, and ADDITIONAL sections into one record iterator before insertion. The bailiwick filter it applies uses the zone context of the NS pool that serviced the lookup, not the zone being queried.
```
djc: I think most advisories hard-wrap at something like 80-100 columns, would a bit nicer. Maybe use `quoted` stuff for code identifiers etc?

djc (commented):
Thanks for picking this up!

crates/hickory-recursor/RUSTSEC-0000-0000.md
cpu: Ah, my mistake. I was looking at the first tag on the repo that had the recursor code but not the published crate metadata versions. I'll adjust here and in the GHSA.

crates/hickory-recursor/RUSTSEC-0000-0000.md
cpu: Done.

crates/hickory-recursor/RUSTSEC-0000-0000.md
cpu: Done, and added more explicit mention of the `hickory-recursor` crate not being supported going forward, pointing all users to `hickory-resover` w/ `recursor` feature.

crates/hickory-recursor/RUSTSEC-0000-0000.md
cpu: Removed both `patched` and `unaffected` since there are no unaffected published versions and no patched versions of `hickory-recursor` (only `hickory-resolver`)

crates/hickory-recursor/RUSTSEC-0000-0000.md
cpu: Removed the casing.

crates/hickory-recursor/RUSTSEC-0000-0000.md
cpu: Wrapped and added more quoting.

cpu:
```
  -> parse error: error parsing ./crates/hickory-recursor/RUSTSEC-0000-0000.md
  -> parse error: TOML parse error at line 1, column 1
  |
1 | [advisory]
  | ^
missing field `versions`
```

How should we handle this? There are no unaffected versions, and there's no patched version :thinking:

crates/hickory-recursor/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,45 @@
...
+
+This issue is fixed in `hickory-resolver` 0.26.0 with the `recursor` feature
+through an architectural change to response-level caching: responses are stored
+keyed by the originating query `(name, type)`. A response to `(attacker.poc.
+NS)` is stored only under that key and cannot affect the `(victim.poc., NS)`
+cache entry.
+
+We believe this issue has been present in all published versions of the
+experimental `hickory-recursor` crate, which has now been folded into the
+`hickory-resolver` crate under the non-defaut `resolver` feature flag. The
```
djc: "defaut" -> "default", the flag is called `recursor`.

djc (approved):

djc:
> ```
>   -> parse error: error parsing ./crates/hickory-recursor/RUSTSEC-0000-0000.md
>   -> parse error: TOML parse error at line 1, column 1
>   |
> 1 | [advisory]
>   | ^
> missing field `versions`
> ```
> 
> How should we handle this? There are no unaffected versions, and there's no patched version 🤔

IIRC we usually do `[versions] patched = []`.

crates/hickory-recursor/RUSTSEC-0000-0000.md
cpu: Fixed both. Thanks!

djc (approved):

PR #2823, opened 2026-04-25
RUSTSEC-2023-0071: Update tracking issue

kpcyrd:
This downgrades the old https://github.com/RustCrypto/RSA/issues/19 link to a reference. It has so many replies and references, it's hard to navigate to the followup issues and pull requests it links to.

The issue url is replaced with the current 2026 tracking issue https://github.com/RustCrypto/RSA/issues/626 (bigint fixed, but more problems from the padding implementation).


djc (approved):

djc:
Thanks!

PR #2824, opened 2026-04-25
RUSTSEC-2026-0078.md: remove reference link already set as URL

DanielEScherzer:
Otherwise results in the same link rendering twice


djc (approved):
Thanks!

PR #2827, opened 2026-04-27
Add two astral-tokio-tar advisories

woodruffw:
This adds two astral-tokio-tar advisories that were disclosed via GHSA today:

- https://github.com/astral-sh/tokio-tar/security/advisories/GHSA-xx64-wwv2-hcqq
- https://github.com/astral-sh/tokio-tar/security/advisories/GHSA-fp55-jw48-c537

~~I've numbered these as `...0000` and `...0001` until they're allocated real numbers 🙂~~


Skgland:
> I've numbered these as ...0000 and ...0001 until they're allocated real numbers 🙂

That doesn't work as the linter requires ...0000 as the prefix. You can add a differentiator in the filename after that e.g. ...0000.1.md and ...0000.2.md

djc (approved):

djc:
Thanks!

PR #2837, opened 2026-05-01
hickory-proto: adjust RUSTSEC-2026-0118 versions

cpu:
This is a bit of a non-standard situation where the `hickory-proto` crate _removed_ the affected type, migrating it to `hickory-net` (which we filed [RUSTSEC-2026-0120](https://rustsec.org/advisories/RUSTSEC-2026-0120.html) for separately). Adjust the affected version range so that `hickory-proto == 0.26.1`, and `hickory-net == 0.26.1` don't surface RUSTSEC-2026-0118.

### DIFF
--- a/crates/hickory-proto/RUSTSEC-2026-0118.md
+++ b/crates/hickory-proto/RUSTSEC-2026-0118.md
@@ -10,7 +10,7 @@ license = "CC-BY-4.0"
 
 [versions]
 patched = []
-unaffected = ["< 0.25.0-alpha.3"]
+unaffected = ["< 0.25.0-alpha.3", ">= 0.26.0-beta.1"]
 ```
 
 # NSEC3 closest-encloser proof validation enters unbounded loop on cross-zone responses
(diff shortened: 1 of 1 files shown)


djc:
I think we could more accurately exclude >= 0.26.0 from affected?

cpu:
> I think we could more accurately set affected to < 0.26.0?

Is `versions.affected` a thing? The [TOML example](https://github.com/RustSec/advisory-db#advisory-format) only has `versions.patched` and `versions.unaffected`, but maybe I'm misunderstanding what you're looking for.

cpu:
Are you asking for `patched` to be returned to `[]` and for `unaffected` to be changed to something like `["< 0.25.0-alpha.3", ">= 0.26"]`?

djc:
Yup! Or, for the lower bound, more accurately the specific 0.26 beta that moved that code out of proto?

cpu:
>  Yup! Or, for the lower bound, more accurately the specific 0.26 beta that moved that code out of proto?

Done :ballot_box_with_check: - I believe it was a370d6522 in v0.26.0-beta.1

djc (approved):

cpu:
Thanks!

PR #2849, opened 2026-05-11
Add advisory for vku VMABuffer::set_data unsoundness

sisy2020:
This PR adds an informational RustSec advisory for an unsound safe API in `vku`.

`VMABuffer::set_data` accepted an arbitrary offset and used it in unsafe pointer arithmetic before copying data into a mapped allocation. A safe caller could trigger an out-of-bounds write.

Upstream issue: https://github.com/ArrowMaxGithub/vku/issues/5
Fix commit: https://github.com/ArrowMaxGithub/vku/commit/ce02c19ec35e5ee84c00ec5005be9d6d44599b5f
Patched version: 0.4.0

### DIFF
--- a/crates/vku/RUSTSEC-0000-0000.md
+++ b/crates/vku/RUSTSEC-0000-0000.md
@@ -1,0 +1,26 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "vku"
+date = "2025-04-23"
+url = "https://github.com/ArrowMaxGithub/vku/issues/5"
+references = ["https://github.com/ArrowMaxGithub/vku/commit/ce02c19"]
+informational = "unsound"
+categories = ["memory-corruption"]
+
+[affected.functions]
+"vku::VMABuffer::set_data" = ["<= 0.3.0"]
+
+[versions]
+patched = [">= 0.4.0"]
+```
+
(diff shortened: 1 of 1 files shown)


djc:
Please obtain approval from the maintainer when submitting advisories.

ArrowMaxGithub:
While I feel humbled that my personal side-project gets any attention, I don't think this advisory is warranted.
It's very much WIP with [zero dependents](https://crates.io/crates/vku/reverse_dependencies) on crates.io and a clear warning to any hypothetical downstream users inside the readme.

djc:
> While I feel humbled that my personal side-project gets any attention, I don't think this advisory is warranted.
> It's very much WIP with [zero dependents](https://crates.io/crates/vku/reverse_dependencies) on crates.io and a clear warning to any hypothetical downstream users inside the readme.

I'm not sure those are the right reasons to advise against publishing an advisory? You published it on crates.io so people can use it, and it's not uncommon for Rust applications making use of libraries not to live on crates.io themselves. Since the cargo-audit/cargo-deny tooling will only warn people who have your crate in their dependency graph anyway, publishing an advisory for rarely used crates shouldn't be a problem.

ArrowMaxGithub:
Feel free to go ahead with this advisory.

djc (approved):

PR #2852, opened 2026-05-12
Add advisory for dynoxide-rs (DNS rebinding / CSRF)

hicksy:
Filing the dynoxide-rs counterpart to [GHSA-fvh2-gm75-j4j7](https://github.com/nubo-db/dynoxide/security/advisories/GHSA-fvh2-gm75-j4j7) / CVE-2026-42559 so cargo audit and cargo deny users get the alert.

The vulnerability lives in the MCP HTTP transport. dynoxide 0.9.13 ships the fix. Full write-up in the GHSA.

### DIFF
--- a/crates/dynoxide-rs/RUSTSEC-0000-0000.md
+++ b/crates/dynoxide-rs/RUSTSEC-0000-0000.md
@@ -1,0 +1,38 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "dynoxide-rs"
+date = "2026-05-12"
+url = "https://github.com/nubo-db/dynoxide/security/advisories/GHSA-fvh2-gm75-j4j7"
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H"
+keywords = ["dns-rebinding", "csrf", "mcp"]
+aliases = ["CVE-2026-42559", "GHSA-fvh2-gm75-j4j7"]
+related = ["GHSA-89vp-x53w-74fx"]
+
+[versions]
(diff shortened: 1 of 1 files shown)


crates/dynoxide-rs/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,38 @@
...
+dynoxide 0.9.13 closes both vectors:
+
+- Upgrades `rmcp` from 1.1.1 to 1.6.0 (which ships a default Host-header allowlist).
+- Sets explicit `allowed_hosts` and `allowed_origins` on `StreamableHttpServerConfig`.
+
+## References
+
+- GitHub Security Advisory: https://github.com/nubo-db/dynoxide/security/advisories/GHSA-fvh2-gm75-j4j7
+- Upstream rmcp advisory: https://github.com/modelcontextprotocol/rust-sdk/security/advisories/GHSA-89vp-x53w-74fx
+- dynoxide release: https://github.com/nubo-db/dynoxide/releases/tag/v0.9.13
```
djc: Let's keep these in the metadata, and not repeat them here. (There's already an alias for the GHSA, so no need to repeat it as a reference.)

I notice there has not been a RustSec advisory for the rmcp problem, maybe you can work with them to get advisories published?

djc:
Thanks!

crates/dynoxide-rs/RUSTSEC-0000-0000.md
hicksy: Done, thanks. 
Good point on the rmcp side. I'll reach out to the maintainers and see if they'd like an advisory published.

djc (approved):

PR #2853, opened 2026-05-12
Add advisory for fourq: OOB pointer arithmetic in Point::from_hash()

yaokunzhang:
## Affected crate(s)

- fourq (14 recent downloads on crates.io)

- https://github.com/982945902/fourq_rust/issues/1

## Severity

Undefined behavior via out-of-bounds pointer arithmetic and write. `Point::from_hash()` is a public safe API that performs unchecked pointer operations on short byte slices, leading to OOB memory writes and aliasing violations. Triggerable from safe code.

## Checklist

- [x] Advisory filename starts with `RUSTSEC-0000-0000` as the ID

### DIFF
--- a/crates/fourq/RUSTSEC-0000-0000.md
+++ b/crates/fourq/RUSTSEC-0000-0000.md
@@ -1,0 +1,27 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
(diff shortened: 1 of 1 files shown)


djc:
No response about whether publishing an advisory is appropriate.

PR #2868, opened 2026-05-13
Add advisories for libcrux-chacha20poly1305, libcrux-ml-dsa

jschneider-bensch:
These are advisories for bug fixes in the latest release of libcrux crates.

Please let me know if more context is necessary, or if the `informational = "notice"` one for ML-DSA is inappropriate.


djc (approved):
LGTM, thanks!

PR #2879, opened 2026-05-13
bump rustsec commit in workflows

Skgland:
as suggeseted in https://github.com/rustsec/rustsec/pull/1598#pullrequestreview-4285289734


djc (approved):

djc:
Thanks!

PR #2880, opened 2026-05-14
Add advisory for lettre for TLS hostname verification disabled in Boring TLS backend

paolobarbolini:
Reports https://github.com/lettre/lettre/security/advisories/GHSA-4pj9-g833-qx53

### DIFF
--- a/crates/lettre/RUSTSEC-0000-0000.md
+++ b/crates/lettre/RUSTSEC-0000-0000.md
@@ -1,0 +1,27 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "lettre"
+date = "2026-05-14"
+url = "https://github.com/lettre/lettre/security/advisories/GHSA-4pj9-g833-qx53"
+categories = ["crypto-failure"]
+cvss = "CVSS:4.0/AV:N/AC:L/AT:P/PR:N/UI:N/VC:H/VI:H/VA:N/SC:L/SI:L/SA:N"
+keywords = ["tls", "smtp", "mitm"]
+aliases = ["GHSA-4pj9-g833-qx53"]
+
+[versions]
+unaffected = ["< 0.10.1"]
+patched = [">= 0.11.22"]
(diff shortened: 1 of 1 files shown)


crates/lettre/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,27 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "lettre"
+date = "2026-05-14"
+url = "https://github.com/lettre/lettre/security/advisories/GHSA-4pj9-g833-qx53"
+categories = ["crypto-failure"]
+cvss_v4_0 = "CVSS:4.0/AV:N/AC:L/AT:P/PR:N/UI:N/VC:H/VI:H/VA:N/SC:L/SI:L/SA:N"
```
amousset:
```suggestion
cvss = "CVSS:4.0/AV:N/AC:L/AT:P/PR:N/UI:N/VC:H/VI:H/VA:N/SC:L/SI:L/SA:N"
```

paolobarbolini: Thanks. We fixed it at the same time :smile:

djc (approved):

djc:
Is there API to enable the boring backend or does that happen only via crate features? If there is, can you list an affected function?

paolobarbolini:
> Is there API to enable the boring backend or does that happen only via crate features? If there is, can you list an affected function?

By default lettre used native-tls. If you only enable the `boring-tls` feature and disable `native-tls`, then you are always vulnerable. If you enable `boring-tls` plus another TLS backend, then `boring-tls` basically doesn't exist unless you explicitly use it via the [builder](https://docs.rs/lettre/0.11.22/lettre/transport/smtp/client/struct.TlsParameters.html#method.new_boring) (unlikely I'd think)

djc:
Obligatory reminder that you should default to rustls. 😛

PR #2882, opened 2026-05-14
Add advisories: oneringbuf vmem double-free + unbounded-spsc transmute UAF

berkant-koc:
Two unsoundness advisories for memory-corruption bugs reachable from 100% safe Rust APIs.

## oneringbuf
`vmem` feature double-free / UAF when source `Box<[UnsafeSyncCell<T>]>` drops after `ptr::copy_nonoverlapping` into the `mmap`'d region. Reachable through `ConcurrentHeapRB::<String>::default(cap)` and any other constructor that funnels through `HeapStorage<T>::new` / `VmemStorage<T>::new`. Affects the actively maintained successor `oneringbuf` 0.7.0 AND the archived predecessor `mutringbuf` (all vmem-enabled versions up to 1.0.0).

## unbounded-spsc
Pointer-as-value `transmute` in `Sender::send` DISCONNECTED arm (`src/lib.rs:379-405`). Transmutes `*mut Producer<T>` (8 bytes) into `Consumer<T>` (which expects a real `Arc` pointer), yielding OOB read inside the `Sender` struct and a fake-Arc drop that hits `dealloc` on an address the allocator never returned. Reachable through `tx.send()` racing `rx.drop()` such that the sender's `connected` check wins but the receiver's `DISCONNECTED` store wins the subsequent `fetch_add`.

Both advisories are also being filed as private disclosures with the respective maintainers via email (no GHSA was available on either upstream repo at the time of submission).

Researcher: Berkant Koc <me@berkoc.com>
PGP: 0C588DFD76204987284213EA0AC529C41F8AA5D6

### DIFF
--- a/crates/oneringbuf/RUSTSEC-0000-0000.md
+++ b/crates/oneringbuf/RUSTSEC-0000-0000.md
@@ -1,0 +1,87 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "oneringbuf"
+date = "2026-05-14"
+url = "https://github.com/Skilvingr/rust-oneringbuf/security/advisories"
+categories = ["memory-corruption"]
+keywords = ["double-free", "use-after-free", "vmem", "drop", "safe-api-ub"]
+informational = "unsound"
+references = [
+    "https://github.com/Skilvingr/rust-mutringbuf/commit/030effa",
+    "https://github.com/Skilvingr/rust-oneringbuf/commit/53a9a0a",
+]
+related = []  # cross-link to the mutringbuf advisory once RUSTSEC IDs are issued
+
+[versions]
+patched = []
+unaffected = []
+
+[affected]
+functions = { "oneringbuf::VmemStorage::new" = ["*"] }
+```
+
+# Double-free / use-after-free in `vmem` storage reachable from 100% safe Rust
+
+When the `vmem` feature is enabled, `VmemStorage<T>::new(Box<[UnsafeSyncCell<T>]>)` (and every public constructor that funnels through it — `ConcurrentHeapRB::default(cap)`, `ConcurrentHeapRB::from(Vec<T>)`, `From<Box<[T]>>`, `From<Box<[UnsafeSyncCell<T>]>>`, `From<Vec<UnsafeSyncCell<T>>>`) bit-copies the input buffer into a freshly `mmap`'d region with `ptr::copy_nonoverlapping`, then lets the source `Box<[UnsafeSyncCell<T>]>` drop normally.
+
+Because `UnsafeSyncCell<T>` has a `Drop` impl that runs `assume_init_drop` on its inner `MaybeUninit<T>`, the source-side `T` values are dropped immediately at the end of `new`, while bitwise duplicates (containing the same heap pointers for `String`, `Box`, `Vec`, `Arc`, …) remain inside the `mmap` region. When the ring buffer is later destructed, the destructor's `drop_in_place` over the slice runs `UnsafeSyncCell::drop` a second time on every cell — a deterministic double-free of every heap-owning element.
+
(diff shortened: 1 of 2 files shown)


crates/oneringbuf/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,87 @@
...
+    let r = vmem_helper::new(&value);
+    let value = ManuallyDrop::new(value);
+    let raw = Box::into_raw(ManuallyDrop::into_inner(value));
+    let layout = core::alloc::Layout::array::<UnsafeSyncCell<T>>(len).unwrap();
+    unsafe { core::alloc::dealloc(raw as *mut u8, layout); }
+    Self { inner: r, len }
+}
+```
+
+Alternative — remove the `Drop` impl from `UnsafeSyncCell` entirely (matching the documented `MaybeUninit` contract that owners must call `assume_init_drop` explicitly), and have `HeapStorage::drop` remain the sole authority on when cells are dropped. This applies uniformly across the `vmem` and non-`vmem` paths.
```
djc, line 76: This is not useful for this advisory.

crates/oneringbuf/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Notes
+
+- The archived predecessor `mutringbuf` (https://github.com/Skilvingr/rust-mutringbuf, archived 2025-11-20) carries the same bug across all `vmem`-enabled releases up to `1.0.0`. A separate (or merged) RustSec advisory should name `mutringbuf` in parallel. The crate's README points users to `oneringbuf` as the maintained successor, so the primary remediation target is `oneringbuf`.
```
djc, line 80: Can you submit this?

crates/oneringbuf/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,87 @@
...
+- Practical exposure on crates.io is currently minimal: 0 reverse dependencies on either crate, ~10 downloads/month combined. The advisory is justified because the bug is a clear soundness defect on a public safe-Rust API, present in the maintained successor, and would survive any future popularity growth.
```
djc, line 81: This is not relevant to advisory readers.

crates/oneringbuf/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,87 @@
...
+- A coordinated GitHub Security Advisory has been drafted at https://github.com/Skilvingr/rust-oneringbuf/security/advisories.
+
+## Researcher
+
+Berkant Koc <me@berkoc.com>
+PGP: 0C588DFD76204987284213EA0AC529C41F8AA5D6
```
djc, line 87: We don't usually maintain this information as part of the advisory.

crates/oneringbuf/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,87 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "oneringbuf"
+date = "2026-05-14"
+url = "https://github.com/Skilvingr/rust-oneringbuf/security/advisories"
```
djc, line 6: There are no advisories here, so this doesn't seem like a useful URL.

djc (commented):
Suggest splitting this in two different PRs, since they're unrelated. Edit down the advisories, it looks like a lot of verbose slop right now.

berkant-koc:
Fair points — I will close this PR and re-open as two separate PRs (one per crate) with significantly trimmed bodies, more in line with what is typical in advisory-db. I will also drop the `Suggested patch` section and instead file the fix as an upstream PR on rust-oneringbuf (and rust-unbounded-spsc) and reference it from the advisory. Will ping you on the new PRs once they are up. Thanks for the quick review.

PR #2890, opened 2026-05-17
Withdraw RUSTSEC-2026-0132

### LINKED ISSUE #2878: RUSTSEC-2026-0132 does not describe any actual unsoundness
The advisory states:
> This can be triggered through safe code — modifying public `Context` fields and then calling `reset()` — with no unsafe required.

But this is not possible to do from a dependent crate, as `Context` is not public. As such, this is merely a code quality consideration (there isn't as much encapsulation between internal modules as there could be), and not an instance of unsoundness that its dependents need to watch out for. The author of the crate [has been informed of the *potential improvement*](https://github.com/rustysec/fuzzyhash-rs/issues/14). Because this advisory does not describe anything actually unsound about the current version of `ssdeep`, I believe it is appropriate to retract it.

kotauskas:
Closes #2878.


djc (approved):

djc:
Thanks!

PR #2912, opened 2026-05-27
Add advisory for inplace-vec-builder: `InPlaceVecBuilder` can corrupt the backing `Vec`

DiuDiu777:
## Affected crate(s)

- `inplace-vec-builder` (1,222,237 recent downloads per crates.io; 5,820,504 total downloads)

- https://github.com/rklaehn/inplace-vec-builder/issues/1

The issue was publicly reported upstream on 2026-03-09 and remains open with no maintainer response as of 2026-05-27.

## Severity

Informational soundness issue. Safe Rust code can leak `InPlaceVecBuilder` with `std::mem::forget` after it has entered an intermediate state, bypassing the `Drop` cleanup that restores the backing `Vec`.

This can expose the `Vec` with an invalid length and duplicated or otherwise invalid elements. Subsequent safe use or drop of the `Vec` can trigger undefined behavior, including double free and use-after-free. Miri reports a dangling pointer dereference during drop. No `unsafe` code is required from the caller.

## Checklist

The issue has been public for more than two months without upstream response. I am filing this advisory under RustSec's documented allowance for advisories after public disclosure with no upstream response.

### DIFF
--- a/crates/inplace-vec-builder/RUSTSEC-0000-0000.md
+++ b/crates/inplace-vec-builder/RUSTSEC-0000-0000.md
@@ -1,0 +1,75 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "inplace-vec-builder"
+date = "2026-03-09"
+url = "https://github.com/rklaehn/inplace-vec-builder/issues/1"
+informational = "unsound"
+categories = ["memory-corruption"]
+keywords = ["soundness", "double-free", "mem-forget", "set-len"]
+
+[versions]
+patched = []
+unaffected = []
+```
+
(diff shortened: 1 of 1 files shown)


djc:
I made an attempt to reach out to the maintainer.

DiuDiu777:
Thank you for reaching out; I’ll wait for any feedback and adjust the advisory if needed.

rklaehn:
I was on holiday and very afk. Taking a look now.

djc:
@rklaehn are you planning to release the fix? Do you think there is enough risk that it merits an advisory (perhaps with `informational = "unsound"`)?

PR #2913, opened 2026-05-27
Fix url

Jake-Shadle:
The PR URL in RUSTSEC-2026-0150 used a . instead of / to separate the domain from the path.


amousset:
Thanks!

PR #2914, opened 2026-05-27
Safe-code use-after-free via `IntoRef::into_ref`

thanasistrisp:
## Affected crate(s)

- `oneringbuf` — affected versions: `< 0.8.0`; patched in `0.8.0`

- Fix commit: https://github.com/skilvingr/rust-oneringbuf/commit/643a24b30914068416dff9021a069c12c865a316

- Maintainer confirmed the issue privately and said opening a RustSec advisory is appropriate.

## Severity

Safe-code heap-use-after-free. Affected versions exposed the obsolete `IntoRef::into_ref` method through the public `IntoRef` trait. For heap-backed ring buffers, this allowed safe external code to obtain and clone a `DroppableRef` handle. Extra clones were not reflected in the internal `alive_iters` counter, so one clone could free the backing allocation while another clone still existed. Dropping the remaining clone then accessed freed memory.

AddressSanitizer confirms this as a heap-use-after-free. The issue was fixed in `0.8.0` by removing the obsolete `into_ref` method.

## Checklist

### DIFF
--- a/crates/oneringbuf/RUSTSEC-0000-0000.md
+++ b/crates/oneringbuf/RUSTSEC-0000-0000.md
@@ -11,7 +11,7 @@ categories = ["memory-corruption"]
 keywords = ["use-after-free", "safe-code", "clone", "drop"]
 
 [versions]
-patched = [>= 0.8.0]
+patched = [">= 0.8.0"]
 unaffected = []
 ```
 
--- a/crates/oneringbuf/RUSTSEC-0000-0000.md
+++ b/crates/oneringbuf/RUSTSEC-0000-0000.md
@@ -1,0 +1,44 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "oneringbuf"
+date = "2026-05-27"
+url = "https://github.com/skilvingr/rust-oneringbuf/commit/643a24b30914068416dff9021a069c12c865a316"
+references = [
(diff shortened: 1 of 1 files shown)


crates/oneringbuf/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,44 @@
...
+date = "2026-05-27"
+url = "https://github.com/skilvingr/rust-oneringbuf/commit/643a24b30914068416dff9021a069c12c865a316"
+references = [
+    "https://github.com/skilvingr/rust-oneringbuf/commit/643a24b30914068416dff9021a069c12c865a316",
+]
+categories = ["memory-corruption"]
+keywords = ["use-after-free", "safe-code", "clone", "drop"]
+
+[versions]
+patched = [">= 0.8.0"]
```
djc, line 14: 0.8 has not actually been released though.

thanasistrisp: Thanks, good point.
@skilvingr, could you publish the fixed version (0.8.0) to crates.io when you get a chance?

skilvingr: Yeah, sorry for the delay. [Published.](https://crates.io/crates/oneringbuf/0.8.0)

djc (approved):

PR #2917, opened 2026-05-28
CVE-2018-25025: update patched version

mrdgo:
After thoroughly inspecting the vulnerability, it is present until 0.7.19, inclusive and only patched in the first 1.0.0 version.

## Affected crate(s)

- actix-web

### DIFF
--- a/crates/actix-web/RUSTSEC-2018-0019.md
+++ b/crates/actix-web/RUSTSEC-2018-0019.md
@@ -8,7 +8,7 @@ url = "https://github.com/actix/actix-web/issues/289"
 aliases = ["CVE-2018-25024", "CVE-2018-25025", "CVE-2018-25026", "GHSA-7x36-h62w-vw65", "GHSA-9qj6-4rfq-vm84", "GHSA-fgfm-hqjw-3265", "GHSA-w65j-g6c7-g3m4"]
 
 [versions]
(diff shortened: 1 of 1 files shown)


djc:
Please provide more context on what research led you to this conclusion.

mrdgo:
If you follow the links below, you can see the vulnerable code:

v0.7.14: https://github.com/actix/actix-web/blob/cd9901c928bfb7b016484f8c0c81c3629eca3e9f/src/server/h1writer.rs#L236

v0.7.19: https://github.com/actix/actix-web/blob/cc6e0c6d04e868e284c45b9ec38f1cc1fe98ac39/src/server/h1writer.rs#L236

The problem is undefined behavior because of mutable access to the original data while a mutable reference exists, see the unsafe section three lines below.
You can see that the pattern is exactly the same: `reserve` is called, while `buf` - a mutable reference into `buffer` - lives.
An exploit is only possible if the backend re-orders code in an unfortunate way, but then an attacker could gain an arbitrary write primitive to the heap, where the `dyn HttpHandler`s, and therefore vtables, live.

Edit: Between the versions, there was commit `782eeb5`. But it only specialized a large unsafe section into multiple smaller ones that actually perform the unsafe operations.

djc (approved):

PR #2921, opened 2026-05-28
Add advisory for solana_rbpf: OOB pointer arithmetic in `invoke_function`

DiuDiu777:
## Affected crate(s)

- `solana_rbpf` (362,603 recent downloads per crates.io; 4,761,947 total downloads)

The upstream repository is archived, so I cannot report this through the normal upstream issue tracker.

- https://github.com/solana-labs/rbpf

## Severity

Informational soundness issue. Safe Rust code can construct an `EbpfVm` and call the safe method `EbpfVm::invoke_function`, which performs out-of-bounds pointer arithmetic internally.

The affected implementation casts `self` to `*mut u64` and applies a randomized offset from `get_runtime_environment_key()` using `ptr::offset`. This can move the pointer outside the allocation containing the `EbpfVm`, causing undefined behavior before the supplied builtin function is invoked. Miri reports an in-bounds pointer arithmetic violation. No `unsafe` code is required from the caller.

## Checklist

### DIFF
--- a/crates/solana_rbpf/RUSTSEC-0000-0000.md
+++ b/crates/solana_rbpf/RUSTSEC-0000-0000.md
@@ -1,0 +1,43 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "solana_rbpf"
+date = "2026-05-28"
+url = "https://github.com/solana-labs/rbpf"
+references = [
+    "https://github.com/anza-xyz/sbpf/pull/151",
+    "https://crates.io/crates/solana-sbpf",
+]
+informational = "unsound"
+categories = ["memory-corruption"]
+keywords = ["soundness", "pointer-arithmetic", "out-of-bounds"]
+
+[affected.functions]
+"solana_rbpf::vm::EbpfVm::invoke_function" = [">= 0.8.0, <= 0.8.5"]
+
+[versions]
+patched = []
+unaffected = ["< 0.8.0"]
+```
+
+# `EbpfVm::invoke_function` performs out-of-bounds pointer arithmetic
+
+Affected versions of `solana_rbpf` expose the safe method
+`EbpfVm::invoke_function`. This method computes an obfuscated VM pointer by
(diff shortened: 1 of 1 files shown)


djc:
@Lichtso do you think it makes sense to publish this, with a note attached to clarify that this crate is unmaintained?

DiuDiu777:
Hi, gentle ping on this advisory PR. Is there anything I can provide or update to help move the review forward? Thanks!

djc:
I reached out to someone privately, let's see if that helps.

nagisa:
It should be fine to land this. The issue is fixed in solana-sbpf (https://github.com/anza-xyz/sbpf/pull/151) which is the maintained incarnation of the crate this report is against. Maybe if there's a way to nudge people towards a different crate, we should do that. Otherwise LGTM!

crates/solana_rbpf/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,104 @@
...
+The resulting pointer arithmetic is performed with `ptr::offset`, which
+requires the computed pointer to remain within the same allocation. In practice,
+the randomized offset can move the pointer far outside the allocation
+containing the `EbpfVm`, causing undefined behavior before the supplied builtin
+function is invoked.
+
+The upstream repository is archived, and no patched version is currently
+available.
+
+## Example
```
djc: Let's drop the example and Miri output, and instead add a section noting that this crate is unmaintained, and pointing users towards the sbpf crate.

djc:
> It should be fine to land this. The issue is fixed in solana-sbpf ([anza-xyz/sbpf#151](https://github.com/anza-xyz/sbpf/pull/151)) which is the maintained incarnation of the crate this report is against. Maybe if there's a way to nudge people towards a different crate, we should do that. Otherwise LGTM!

Thanks!

DiuDiu777:
Done~ :)

djc (approved):

PR #2931, opened 2026-06-02
Add advisories for metacall

thanasistrisp:
## Affected crate

- metacall (2,066 recent downloads)

## Links to upstream issue

https://github.com/metacall/core/issues/809

## Severity

I am reporting two different categories of memory vulnerabilities. 
The first is more critical since every time the `MetaCallException` is created, when it is dropped, it leads to a bad-free. This can be triggered through the safe public API `MetaCallException::new()`, with no `unsafe` required from the caller. 
The second one refers to many unsoundness issues found in `Clone` and `new_raw` methods of multiple traits.

## Checklist

### DIFF
--- a/crates/metacall/RUSTSEC-0000-0000.1.md
+++ b/crates/metacall/RUSTSEC-0000-0000.1.md
@@ -1,0 +1,44 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "metacall"
+date = "2026-06-01"
+url = "https://github.com/metacall/core/issues/809"
+categories = ["memory-corruption"]
+keywords = ["bad-free"]
+
+[versions]
+patched = []
+```
+
+# Bad-free in `MetaCallException::new`
+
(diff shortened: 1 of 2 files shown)


djc (approved):
LGTM, but I'd like to include patched versions if they're coming soon.

djc:
Actually, are these issues independent from the existing issues reported in

https://rustsec.org/advisories/RUSTSEC-2026-0139.html

?

thanasistrisp:
> Actually, are these issues independent from the existing issues reported in
> 
> https://rustsec.org/advisories/RUSTSEC-2026-0139.html
> 
> ?

`MetaCallException::new(...)` is a completely different case and much more severe, since it is not about unsoundness at all. (RUSTSEC-0000-0000.1.md)

The other ones (unsoundness) are part of previously reported items (but the same two methods are also exposed to many more classes), but the underlying bug is a bit different; I explain more inside the RUSTSEC-0000-0000.2.md‎. But if you think they are very similar, we could include them to the prior RUSTSEC ID.

Feel free to correct me.

djc (approved):

PR #2945, opened 2026-06-05
Another Diesel Advisory

weiznich:
This fills another advisory for another Diesel issue recently uncovered by LLM/AI.

## Affected crate(s)

- Diesel

https://github.com/diesel-rs/diesel/commit/1bc2ea46d9840e8d9af844239d3c84f37fe7d84b
https://github.com/diesel-rs/diesel/pull/5056

## Severity

The API provided by Diesel allowed an unsound usage of `SqliteConnection::deserialize_readonly_database` as it did not ensure that the passed buffer is not dropped before the created connection is dropped. This could lead to an use-after-free issue, but only if the user actively keeps around the connection for longer than the buffer. Whether or not that's an issue depends on the users code calling this function, therefore I filled this as informal unsound advisory rather than as "real" vulnerability.

## Checklist

### DIFF
--- a/crates/diesel/RUSTSEC-0000-0000.md
+++ b/crates/diesel/RUSTSEC-0000-0000.md
@@ -1,0 +1,28 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "diesel"
+date = "2026-06-05"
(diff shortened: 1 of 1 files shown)


djc:
There is only a merge commit in your branch apparently?

weiznich:
Ah right, I should actually add the advisory before pushing :facepalm:

djc (approved):

djc (approved):

PR #2950, opened 2026-06-08
Update RUSTSEC-2026-0173, RUSTSEC-2024-0370 with proc-macro-error3 as alternative

gamma0987:
## Affected crate(s)

- proc-macro-error (40,369,775)
- proc-macro-error2 (35,084,645)

## Affected Advisories

RUSTSEC-2024-0370
RUSTSEC-2026-0173

proc-macro-error3 is a maintained fork of the proc-macro-error2 crate, which also was a fork and drop-in replacement for proc-macro-error.


djc:
We don't recommend new crates.

PR #2956, opened 2026-06-11
Add advisory for triton-vm

jan-ferdinand:
## Affected crate(s)

- [triton-vm](https://crates.io/crates/triton-vm)

None; [here](https://github.com/TritonVM/triton-vm/commit/17c7ba0a19f31793f1c807a58afaf9504342aac5)'s the commit of the fix, though.

> In principle, we do not publish advisories for issues that have not been reported upstream.

The report did not happen through public channels.

## Severity

Any Triton VM application using instruction `sponge_absorb_mem` to hash memory data can be given a proof for a forged hash that doesn't correspond to the actual memory. This breaks soundness of Triton VM's proof system.

## Checklist

### DIFF
--- a/crates/triton-vm/RUSTSEC-0000-0000.md
+++ b/crates/triton-vm/RUSTSEC-0000-0000.md
@@ -1,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

PR #2958, opened 2026-06-11
Report memory exposure in PyO3 `nth_back` iterator methods

davidhewitt:
## Affected crate(s)

- `pyo3` (40M recent downloads per crates.io)

https://github.com/PyO3/pyo3/pull/6086

## Severity

Code that uses `BoundListIterator` and `BoundTupleIterator`'s `nth_back` methods with attacker-controlled N are vulnerable to out-of-bound reads if the iterators have already yielded items.

My assessment is that this is a relatively niche iterator method with a couple of conditions to make it a problem, but code meeting the vulnerable combination could easily go undetected and deserves both a fix and an advisory.

## Checklist

### DIFF
--- a/crates/pyo3/RUSTSEC-0000-0000.md
+++ b/crates/pyo3/RUSTSEC-0000-0000.md
@@ -1,0 +1,36 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

PR #2960, opened 2026-06-11
Report thread safety issue with PyO3's `PyCFunction::new_closure`

davidhewitt:
## Affected crate(s)

- `pyo3` (40M recent downloads per crates.io)

https://github.com/PyO3/pyo3/pull/6096

## Severity

`PyCFunction::new_closure` missed a `Sync` bound on the closure to be wrapped as a Python callable. Older Python releases would have been less likely to observe data races due to partial serial execution imposed by the Global Interpreter Lock. As the Python ecosystem transitions to free-threaded Python, the `Sync` bound is critical for thread safety.

This API is not the primary mechanism by which user code produces Python callables (typically done by `#[pyfunction]` and `#[pymethods]` proc macros), however it seems highly likely that users will have written thread-unsafe code using this API. PyO3 happened to have such a case in its own test suite! (Was using a `RefCell` inside the closure in question.)

## Checklist

### DIFF
--- a/crates/pyo3/RUSTSEC-0000-0000.md
+++ b/crates/pyo3/RUSTSEC-0000-0000.md
@@ -1,0 +1,34 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "pyo3"
+date = "2026-06-11"
+url = "https://github.com/PyO3/pyo3/pull/6096"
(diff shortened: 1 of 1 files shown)


LawnGnome (approved):

LawnGnome:
Thanks!

PR #2971, opened 2026-06-15
Add a new Wasmtime advisory

alexcrichton:
Mirroring https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-3p27-qvp9-27qf to here.


LawnGnome (approved):
I do wonder if we should be reproducing a bit more information into the RustSec advisories than just the title, but that's a process question for us. Approving and merging, since this in line with existing practice. Thanks!

PR #2976, opened 2026-06-17
Add advisory for jomini BufferWindow unsoundness

sisy2020:
## Affected crate(s)

- `jomini`

- Upstream issue: https://github.com/rakaly/jomini/issues/178
- Fix PR: https://github.com/rakaly/jomini/pull/179
- Fix commit: https://github.com/rakaly/jomini/commit/1f68766666aecdb3cfb99527240f2e4305b7104b

## Severity

This is proposed as an informational unsound advisory.

Although `BufferWindow` was crate-internal, it was used internally by public 
reader APIs, and its safe methods performed unsafe pointer and slice operations 
that depended on internal invariants. If those invariants were broken, 
these operations could cause undefined behavior.

No concrete malformed-input path has been identified, so the advisory is
intentionally scoped conservatively. The issue was fixed in `jomini` 0.29.0.

I would appreciate feedback on whether this crate-internal pointer-invariant
issue is appropriate for an informational RustSec advisory.

## Checklist

### DIFF
--- a/crates/jomini/RUSTSEC-0000-0000.md
+++ b/crates/jomini/RUSTSEC-0000-0000.md
@@ -1,0 +1,38 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "jomini"
+date = "2025-04-24"
+url = "https://github.com/rakaly/jomini/issues/178"
+references = [
+    "https://github.com/rakaly/jomini/pull/179",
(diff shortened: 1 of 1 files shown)


djc:
If there is no public API path to the problem by default I don't think this merits a RustSec advisory, although I would be willing to change my mind if the maintainer thinks there is a reason to.

sisy2020:
Thank you very much for taking the time to review this. I checked with the upstream maintainer, and I agree that the PR should remain closed.
I really appreciate you looking into it.

PR #2980, opened 2026-06-17
Add advisory for capnp unsoundness

Waan2y:
## Affected crate(s)

- capnp

- https://github.com/capnproto/capnproto-rust/issues/667

## Severity
- Unsoundness of safe public API

## Checklist

## Comment
This issue has the same underlying cause as RUSTSEC-2025-0143, but it was found in a different API that is not mentioned in the existing advisory.

I would appreciate guidance from the maintainer on whether this would be more appropriately handled as an update to the existing advisory, or as a separate advisory linked through related.

Based on my current understanding of the guidance, I have marked this issue as related by adding `related = ["RUSTSEC-2025-0143"]`, but I would be happy to adjust it if the maintainers recommend a different way to handle this .

### DIFF
--- a/crates/capnp/RUSTSEC-0000-0000.md
+++ b/crates/capnp/RUSTSEC-0000-0000.md
@@ -1,0 +1,35 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "capnp"
+date = "2026-06-15"
+url = "https://github.com/capnproto/capnproto-rust/issues/667"
+categories = ["memory-corruption"]
+keywords = ["unsoundness", "undefined-behavior"]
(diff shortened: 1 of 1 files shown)


djc:
I've been trying to limit the amount of "noisy" advisories for things that are low risk of hitting inadvertently:

This feels like it has pretty low risk of being hit by a user, what do you think? @dwrensha thoughts?

dwrensha:
I think this is low risk of being hit by a user. It is not intended that a downstream user should ever construct a `RawEnumSchema`.

djc:
Okay, in that case I'm going to close this.

Waan2y:
I understand the history and policy you mentioned.
I sincerely appreciate your helpful input and guidance.

PR #2982, opened 2026-06-17
CVE-2019-16760.md: update gist link

DanielEScherzer:
CC @emilyalbini
@djc should links to individual gists to replaced with copies so that they don't break in the future?


djc (approved):
Thanks. I don't think this is common enough that we should mirror gists.

(A little surprising that GitHub doesn't have redirects in place for this like they do for repos? Seems straightforward...)

PR #2986, opened 2026-06-21
Add advisory for hftbacktest DataPtr::at unsoundness

sisy2020:
## Affected crate(s)

- `hftbacktest` (597 recent downloads per crates.io)

- Upstream issue: https://github.com/nkaz001/hftbacktest/issues/212
- Fix commit: https://github.com/nkaz001/hftbacktest/commit/69474f8c684db0c304cc885240c8ba5722af68b2

## Severity

Soundness issue in a safe public API. `DataPtr::at` accepted an arbitrary index and used it in unsafe pointer arithmetic via `ptr.add(index)`. Safe callers could pass an out-of-bounds index and trigger undefined behavior without using `unsafe`.

The issue was fixed upstream by marking `DataPtr::at` as `unsafe`. The fix is included in `hftbacktest` version `0.8.1`.

## Checklist

### DIFF
--- a/crates/hftbacktest/RUSTSEC-0000-0000.md
+++ b/crates/hftbacktest/RUSTSEC-0000-0000.md
@@ -1,0 +1,24 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "hftbacktest"
+date = "2025-04-24"
+url = "https://github.com/nkaz001/hftbacktest/issues/212"
+references = ["https://github.com/nkaz001/hftbacktest/commit/69474f8c684db0c304cc885240c8ba5722af68b2"]
+informational = "unsound"
+
+[affected.functions]
+"hftbacktest::backtest::data::DataPtr::at" = ["< 0.8.1"]
+
+[versions]
+patched = [">= 0.8.1"]
+```
+
+# `DataPtr::at` may create out-of-bounds pointers from safe code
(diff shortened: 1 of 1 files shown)


djc:
It doesn't seem like this is worthy of an advisory.

- Few downloads
- No reverse dependencies
- Looks like using a `DataPtr` already requires using `unsafe` API

sisy2020:
Thanks again for the review. I have no objection to closing this.

For completeness, I re-checked the versions published on crates.io. It looks like `DataPtr::at` was a safe `pub fn` from `0.3.0` through `0.7.0`, and was changed to `unsafe` in `0.8.0`, so the affected range would be `< 0.8.0`.

On the “requires unsafe” point, I also found that `DataPtr::new` was a safe public constructor, so safe code could obtain a `DataPtr` and reach the unchecked `ptr.add` through `DataPtr::at`:
```rust
// DataPtr::new was a safe public constructor:
// pub fn new(size: usize) -> Self { ... managed: true }
use hftbacktest::backtest::data::DataPtr;
fn main() {
    let dp = DataPtr::new(16);     // allocates 16 bytes — safe
    let _p = dp.at(1_000_000);     // ptr.add(1_000_000) — out of bounds
}
```

Just adding this for accuracy and for the public record, not to argue for reopening the advisory. Thanks again for taking the time to review it.

PR #2990, opened 2026-06-22
Add unsound pointer arithmetic advisory for memmap2.

de-vri-es:
## Affected crate(s)

- memmap2 (56,337,220 recent downloads per crates.io)

https://github.com/RazrFalcon/memmap2-rs/issues/169
https://github.com/RazrFalcon/memmap2-rs/issues/170

## Severity

Unchecked pointer arithmetic (`pointer::offset` and `pointer::add`) in a safe public API can lead to undefined behavior. The pointer is not dereferenced, but it is passed to `madvise` and `msync` syscalls (and Windows equivalent). 

## Checklist

### DIFF
--- a/crates/memmap2/RUSTSEC-0000-0000.md
+++ b/crates/memmap2/RUSTSEC-0000-0000.md
@@ -1,0 +1,40 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "memmap2"
+date = "2026-06-20"
+url = "https://github.com/RazrFalcon/memmap2-rs/issues/169"
+references = ["https://github.com/RazrFalcon/memmap2-rs/pull/170"]
+informational = "unsound"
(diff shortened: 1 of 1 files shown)


LawnGnome (approved):

LawnGnome:
Thanks!

djc:
Given this comment from one of the upstream issues:

> Probably best, although I do think the real-world risk is limited. I don't expect there are applications that take the offset or range passed to `flush_range` or `advise_rage` from user input. Accidental UB from a bug is somewhat more likely.

I'm not totally sure that an advisory really helps here.

de-vri-es:
Well, thats my guess. But it is still UB in the safe API of a crate with 1810 direct dependents and 56 million downloads in the last 90 days.

So I also dont think my guess on how it's used is good enough ^_^

PR #2999, opened 2026-06-26
Add advisory for polkavm `write_memory` and `zero_memory` corruption

sorpaas:
## Affected crate(s)

- `polkavm`

* https://github.com/paritytech/polkavm/commit/0ec2c443aad4
* https://github.com/paritytech/polkavm/compare/v0.33.0...v0.33.1

## Severity

High/critical as this crate's consumers commonly use it in blockchain consensus. Anyone who has not yet updated from the yanked versions risk consensus splits.

## Checklist

- [X] Advisory filename(s) starts with `RUSTSEC-0000-0000` as the ID
- [X] `date` field is set to the public disclosure date
- [X] Contains a concise and descriptive title after advisory metadata
- [X] Fix already in releases.

### DIFF
--- a/crates/polkavm/RUSTSEC-0000-0000.md
+++ b/crates/polkavm/RUSTSEC-0000-0000.md
@@ -1,0 +1,23 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "polkavm"
+date = "2026-06-26"
+url = "https://github.com/paritytech/polkavm/commit/0ec2c443aad4"
+references = ["https://github.com/paritytech/polkavm/compare/v0.33.0...v0.33.1"]
+categories = ["memory-exposure", "denial-of-service"]
+cvss = "CVSS:4.0/AV:N/AC:L/AT:P/PR:N/UI:N/VC:L/VI:H/VA:N/SC:N/SI:H/SA:L"
+keywords = ["interpreter", "memory", "consensus", "vm"]
+
+[versions]
+patched = [">= 0.33.1"]
+unaffected = ["< 0.32.0"]
+```
+
+# Interpreter guest memory corruption in `write_memory` and `zero_memory`
+
+In `polkavm` `0.32.0` and `0.33.0` the interpreter has a new memory implementation. It mishandles host-initiated memory operations (`write_memory` and `zero_memory`) when the target range already lies within the allocated backing buffer (`rw_data` / `aux`). Such operations fell through to a resize-based code path that placed bytes at the wrong location (or dropped them), and left regions that should have been zeroed un-cleared.
+
+The result is that host writes into guest memory can be silently misplaced or lost, and zeroing can be incomplete.
+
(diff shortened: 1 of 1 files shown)


djc:
> ## Checklist
> 
> * [x]  Advisory filename(s) starts with `RUSTSEC-0000-0000` as the ID
> * [x]  `date` field is set to the public disclosure date
> * [x]  Contains a concise and descriptive title after advisory metadata
> * [x]  Fix already in releases.

Don't change the checklist. I'd still like to see confirmation from @koute whether an advisory makes sense here.

sorpaas:
I added the item back.

Right now 0.32.0 and 0.33.0 still receives 5-10 downloads per day, which is dangerous -- a lot of the consumers probably use it in consensus critical code. A private fork of Substrate, etc that they maintain themselves and update dep versions manually. It would be great if they can get notified in time.

koute:
> I'd still like to see confirmation from @koute whether an advisory makes sense here.

@djc Maintainer here.

Fundamentally I'm fine either way. However, few points:

- I don't believe the code in question was ever used by anyone in production, and the crate itself pretty much has only a single actual production user (us) who'd care about its security.
- If you read our README you'll see a big "Do not use it in production!" warning, which (I hope) further detracts any possible third-party use (especially in a security-sensitive context).
- The description of the issue is mostly wrong. Basically, the issue here is that calling `write_memory`/`zero_memory` from the host and when using the interpreter backend might clear parts of read-write memory of the guest program running inside the VM.
- The practical benefit of issuing this advisory is pretty much zero.

Please note that the author of this issue is acting in bad faith, and his main objective here is to generate negative publicity for us rather than engage in good faith. The fact that he did not ping me, or that he overexaggerates the actual impact ("commonly use" where it's pretty much only used by us, 5-10 downloads per day "dangerous" where it's most definitely automated bots, etc.) should clue you in to this fact. (I can expand on this with _much_ more, but I doubt you're interested in our drama. Regardless of whether this security advisory is published he *will* publish another hit piece/Twitter post criticizing us, accomplishing his goals.)

djc:
Thanks for the context -- just going to close this.

sorpaas:
I'm don't think personal attack is appropriate here. We just disagree with polkavm's security practices.

* `polkavm` is itself used in production (deployed live in Polkadot Hub). That alone overwrote some third-party teams' judgements about its security (even with the big warning in README). This means that it deserves to have its own security advisories, regardless of author's claim about the library's stability.
* From 0.32.0 (first vulnerable release) to 0.33.1 (the patch), there's a ~3.5 week long period. Anyone who incorporated polkavm during that time risk being pinned to 0.32.0 or 0.33.0. (And the download does show this. I don't think they can be attributed to "bots".)
* We can't ignore third-party users and users developing in private. I don't think that's appropriate. We know there are small teams (like Rialo) that have already been using `polkavm`, but depended on old versions.

It seems that the maintainer already decided to take a side. We of course appreciate reconsideration, but we won't further budge.

In any case, we do want to keep the records clear.

PR #3005, opened 2026-06-29
Add advisory for stackvector public length field unsoundness

sisy2020:
## Affected crate(s)

- `stackvector` (16,059 recent downloads per crates.io)

- Upstream issue: https://github.com/Alexhuszagh/rust-stackvector/issues/3
- Fix PR: https://github.com/Alexhuszagh/rust-stackvector/pull/6
- Fix commit: https://github.com/Alexhuszagh/rust-stackvector/commit/02b947afdeeb1be95ec0888354aa76afdd9d0357

## Severity

This advisory classifies the issue as informational `unsound`. In affected versions, `StackVec::length` is public, so safe Rust code can set it to a value larger than the backing array capacity. Safe methods such as `remove`, `pop`, and `truncate` then rely on this corrupted length before performing unsafe pointer operations, which can lead to out-of-bounds access and undefined behavior.

The fix has been merged upstream, but at the time of writing, a patched release does not appear to have been published to crates.io.

## Checklist

### DIFF
--- a/crates/stackvector/RUSTSEC-0000-0000.md
+++ b/crates/stackvector/RUSTSEC-0000-0000.md
@@ -7,17 +7,20 @@ url = "https://github.com/Alexhuszagh/rust-stackvector/issues/3"
 references = [
     "https://github.com/Alexhuszagh/rust-stackvector/pull/6",
     "https://github.com/Alexhuszagh/rust-stackvector/commit/02b947afdeeb1be95ec0888354aa76afdd9d0357",
+    "https://github.com/Alexhuszagh/rust-stackvector/issues/5",
 ]
 informational = "unsound"
 
 [versions]
-patched = []
+patched = [">= 2.0.0"]
 ```
 
-# stackvector's public `length` field allows memory safety violations from safe code
+# Multiple soundness issues in `stackvector`
 
-Affected versions of `stackvector` expose the `StackVec::length` field as public. Safe Rust code can set `length` to a value larger than the backing array capacity. Other safe methods, including `remove`, `pop`, and `truncate`, rely on `length` to check bounds before performing unsafe pointer operations (`ptr::read`, `ptr::copy`, `offset`/`add`).
+Affected versions of `stackvector` contained multiple soundness issues that could allow safe Rust code to trigger undefined behavior.
 
-If `length` is corrupted by safe code, these methods may perform out-of-bounds pointer arithmetic, reads, writes, or copies. This makes the safe API unsound, since callers could trigger undefined behavior without using `unsafe`.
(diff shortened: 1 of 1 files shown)


djc:
Waiting for maintainer confirmation if an advisory is appropriate.

Alexhuszagh:
The advisory is appropriate. I'll file an update when the new version is released and the advisory no longer applies.

djc:
> The advisory is appropriate. I'll file an update when the new version is released and the advisory no longer applies.

Do you have an indication of how long you think that'll take?

Alexhuszagh:
> > The advisory is appropriate. I'll file an update when the new version is released and the advisory no longer applies.
> 
> Do you have an indication of how long you think that'll take?

I can have it in 3 hours (at work but I can do this on my break).  I was planning on waiting for a release for a broader refactor but life got in the way so having an immediate release is likely a better idea and wait for the refactor after.

sisy2020:
Thanks for the confirmation. I’m happy to update the advisory once a patched release is published.

Alexhuszagh:
Version 1.2.0 has been released with these patches:
https://crates.io/crates/stackvector/1.2.0

There are a few issues that still need to be resolved, specifically, the use of [`mem::uninitialized`](https://doc.rust-lang.org/std/mem/fn.uninitialized.html) in `StackVec::from_vec_unchecked` which is called in `from_vec` (see Alexhuszagh/rust-stackvector#5), which I can get to later.

There's also some other Miri violations that likely need to be fixed. Specifically, a overall migration to `MaybeUninit` in general instead of other logic (you can validate this by running `cargo +nightly miri test` and the borrow stack rules are broken due a misuse of `MaybeUninit` (and would require a breaking change). So as the repository stands right now, an advisory would still apply (for other reasons).

UPDATE: It looks like the soundness fixes shouldn't be too involved and I can have a release that will fix all soundness issues.

djc:
> UPDATE: It looks like the soundness fixes shouldn't be too involved and I can have a release that will fix all soundness issues.

Let's hold the advisory for that, then. Thanks for your work on fixing these issues!

Alexhuszagh:
Version 2.0.0 has been released which fixes all the soundness issues:
https://crates.io/crates/stackvector/2.0.0

Also miri has been added to CI/CD to help mitigate any future soundness.

crates/stackvector/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,23 @@
...
+date = "2025-10-23"
+url = "https://github.com/Alexhuszagh/rust-stackvector/issues/3"
+references = [
+    "https://github.com/Alexhuszagh/rust-stackvector/pull/6",
+    "https://github.com/Alexhuszagh/rust-stackvector/commit/02b947afdeeb1be95ec0888354aa76afdd9d0357",
+]
+informational = "unsound"
+
+[versions]
+patched = []
```
djc: Please add the fixed version.

djc (approved):

PR #3017, opened 2026-06-30
Add advisory for grubbnet NetworkBuffer drain unsoundness

sisy2020:
## Affected crate(s)

- `grubbnet` (143 recent downloads per crates.io)

- Upstream issue: https://github.com/Dooskington/grubbnet/issues/4
- Fix commit: https://github.com/Dooskington/grubbnet/commit/32562b9dca028357b27a9d317d62316be22035bd

## Severity

This advisory is proposed as informational `unsound`. Affected versions exposed the safe public method `NetworkBuffer::drain`, which used a caller-controlled `count` in unsafe pointer arithmetic and `ptr::copy` without checking. If `count` was too large, safe callers could trigger integer underflow and out-of-bounds memory copying, causing undefined behavior.

The issue was fixed in `0.2.0` by validating `count` before performing the copy.

## Checklist

### DIFF
--- a/crates/grubbnet/RUSTSEC-0000-0000.md
+++ b/crates/grubbnet/RUSTSEC-0000-0000.md
@@ -1,0 +1,26 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "grubbnet"
+date = "2025-05-21"
+url = "https://github.com/Dooskington/grubbnet/issues/4"
+references = [
+    "https://github.com/Dooskington/grubbnet/commit/32562b9dca028357b27a9d317d62316be22035bd"
+]
+informational = "unsound"
+
+[affected.functions]
+"grubbnet::buffer::NetworkBuffer::drain" = ["< 0.2.0"]
+
+[versions]
+patched = [">= 0.2.0"]
+```
+
+
+# `NetworkBuffer::drain` can cause undefined behavior from safe code
+
+Affected versions of `grubbnet` exposed a safe public method `NetworkBuffer::drain` that accepted a caller-provided `count` and used it in unsafe pointer operations without validating that `count` was no larger than the number of bytes currently stored in the buffer (`self.offset`).
+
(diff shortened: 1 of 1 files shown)


sisy2020:
One additional note: I noticed that this crate has relatively few recent downloads, so I’m opening this as a draft first to ask whether this kind of advisory is appropriate for inclusion in RustSec.

My understanding is that the issue is directly reachable through a safe public API: `NetworkBuffer::drain` is safe and public, and safe callers can trigger the invalid copy without writing any `unsafe` code. That said, I’m happy to close this PR if the project would prefer not to include advisories for crates with this level of usage.

We also have a few other low-download soundness cases. If preferred, I can collect them into a single issue for review first, or open them as draft PRs so you can review each concrete advisory before it is finalized.

Thanks for your guidance!

djc:
Given that it looks like most users would want/need to call this API and it actually triggers undefined behavior, I think an `unsound` advisory for this is probably okay.

> We also have a few other low-download soundness cases. If preferred, I can collect them into a single issue for review first, or open them as draft PRs so you can review each concrete advisory before it is finalized.

I think collecting them into a single issue for review would be useful. I think these axes make for useful comparison:

- How popular is the crate
- How likely is that a user will (accidentally) trigger the issue (especially from network/untrusted file input)
- What is the impact once triggered

Suggest ranking issues along these lines.

sisy2020:
Hi, just a gentle ping on this PR as it's been inactive for a while. Is there anything I can update or do to help move this forward? Thanks!

djc:
We recently had a discussion about notability criteria, and I think grubbnet fails them:

> at least one [crates.io](http://crates.io/) crate (from a different author) that depends on it, or >1k recent downloads or 100k all time downloads

So suggest we drop this one.

sisy2020:
Thanks, that makes sense. I agree that this one doesn't meet the notability criteria, so I'll close the PR. Thanks for pointing me to the discussion!

PR #3021, opened 2026-07-01
Add advisory for quick-xml: unbounded namespace-declaration allocation (memory DoS)

qifan-sailboat:
This adds a RustSec advisory for [quick-xml#970](https://github.com/tafia/quick-xml/issues/970).

- **Crate:** `quick-xml`
- **Affected:** `< 0.41.0`
- **Patched:** `0.41.0` (released 2026-06-29)
- **Category:** denial-of-service (uncontrolled memory allocation)

`NsReader` calls `NamespaceResolver::push` for every `Start`/`Empty` event *before* returning it to the caller. `push` allocated one `NamespaceBinding` (plus prefix/value bytes) per `xmlns` / `xmlns:*` attribute with no upper bound, so a start tag with `N` namespace declarations drove ~3× the tag's byte size in resolver heap *inside* quick-xml — allocation a consumer cannot observe or bound even if it caps its input size. On untrusted XML, and especially with several `NsReader`s running concurrently, this can exhaust process memory (OOM). Only `NsReader` consumers are affected; a plain `Reader` without namespace resolution is not. Fixed in 0.41.0 (commit [`7ca2526`](https://github.com/tafia/quick-xml/commit/7ca25266e94987210daa864889ab15c9332c8a2a)), which adds a configurable per-element declaration cap (default 256) and `NamespaceError::TooManyDeclarations`.

Reported by me; the `quick-xml` maintainer asked that I file the RustSec advisory. The ID is left as the `RUSTSEC-0000-0000` placeholder for assignment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/quick-xml/RUSTSEC-0000-0000.md
+++ b/crates/quick-xml/RUSTSEC-0000-0000.md
@@ -1,0 +1,61 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "quick-xml"
+date = "2026-06-29"
+url = "https://github.com/tafia/quick-xml/issues/970"
+references = [
+    "https://github.com/tafia/quick-xml/commit/7ca25266e94987210daa864889ab15c9332c8a2a",
+]
+categories = ["denial-of-service"]
+keywords = ["xml", "parser", "dos", "memory", "namespace"]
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

PR #3031, opened 2026-07-04
Add advisory for unbounded memory allocation in geozero WKB decoding

LucaCappelletti94:
This adds a denial-of-service advisory for `geozero`. Decoding untrusted WKB or EWKB through `to_geo` reserves a `Vec` sized from a 32-bit element count read directly from the input, before the coordinate bytes are read, so a 38-byte blob can declare a count near `u32::MAX` and drive an allocation of roughly 47 GB that aborts the process. The reproduction in the advisory is verified against `geozero` 0.15.1 with no other crates involved. A fix that bounds the reservation is proposed upstream in georust/geozero#297 (open since 2026-06-17). No fixed release exists yet, so `patched` is empty.

### DIFF
--- a/crates/geozero/RUSTSEC-0000-0000.md
+++ b/crates/geozero/RUSTSEC-0000-0000.md
@@ -46,7 +46,7 @@ fn main() {
 
 ## Patches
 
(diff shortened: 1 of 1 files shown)


djc:

LucaCappelletti94:
Done #3032

PR #3032, opened 2026-07-04
Add advisory for unbounded memory allocation in geozero WKB decoding

LucaCappelletti94:
## Affected crate(s)

- `geozero` (226,232 recent downloads per crates.io)

- Proposed fix: georust/geozero#297 (open since 2026-06-17)

The aforementioned PR has now been merged, updated crate is yet to be released as possibly other PRs need to be merged first.

## Severity

Remote denial-of-service via unbounded memory allocation (CWE-770). Decoding untrusted WKB or EWKB through `to_geo` reserves a `Vec` from a 32-bit element count read directly from the input, before the coordinate bytes are read, so a 38-byte blob can declare a count near `u32::MAX` and drive an allocation of roughly 47 GB that aborts the process. Verified against `geozero` 0.15.1 with no other crates involved.

## Checklist

### DIFF
--- a/crates/geozero/RUSTSEC-0000-0000.md
+++ b/crates/geozero/RUSTSEC-0000-0000.md
@@ -46,7 +46,7 @@ fn main() {
 
 ## Patches
 
-No fixed release is available yet. A fix that bounds the reservation against the bytes remaining in the input (or grows fallibly with `Vec::try_reserve`) is proposed in [georust/geozero#297](https://github.com/georust/geozero/pull/297).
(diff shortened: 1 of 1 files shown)


LucaCappelletti94:
The proposed fix has now been merged. The updated crate has not yet been released.

djc:
> The proposed fix has now been merged. The updated crate has not yet been released.

If the fix will be released, shortly, I'd prefer to hold off on publishing the advisory. Ask the maintainer(s) about their planning?

PR #3034, opened 2026-07-04
Add advisory for fulgur: blank-page denial of service via non-painting replaced elements

mitsuru:
Adds a denial-of-service advisory for the `fulgur` crate (an HTML/CSS to PDF converter). This is a maintainer-published advisory — the corresponding GitHub Security Advisory GHSA-4rf6-qx84-q9fv has already been published by the fulgur project.

## Affected crate(s)

- `fulgur` (1,797 recent downloads per crates.io)

- GitHub Security Advisory (maintainer-published): https://github.com/fulgur-rs/fulgur/security/advisories/GHSA-4rf6-qx84-q9fv
- Upstream fix: https://github.com/fulgur-rs/fulgur/pull/575 (merged; released in 0.26.0)

## Severity

Denial of service via uncontrolled resource consumption (CWE-400). In `fulgur` < 0.26.0, a non-painting replaced element — an unresolved `src` (the common offline-first case), a `visibility:hidden` image, an undecodable image format, or an empty `<svg>` — bypassed the childless-collapse defense and amplified a few bytes of HTML into roughly 10,000 (`MAX_PAGES`) blank PDF pages, exhausting CPU and memory via the per-page allocation and render loop. A trailing-sibling variant of the same gap was also open. Fixed in 0.26.0 by removing the tag-only "replaced content" gate so any pathologically tall childless box collapses.

The amplification is bounded (no infinite loop), so it is scored CWE-400 only. CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H (7.5); `fulgur` performs no network I/O of its own, so the Attack Vector is assessed as Network per the CVSS v3.1 User Guide §3.7, as explained in the advisory body. The unbounded pre-0.19.0 variant is tracked separately as GHSA-j5cx-ph8g-95v3.

## Checklist

- [x] Asked maintainer(s) if publishing an advisory is appropriate (maintainer-published; GHSA-4rf6-qx84-q9fv was filed by the fulgur project itself)

### DIFF
--- a/crates/fulgur/RUSTSEC-0000-0000.md
+++ b/crates/fulgur/RUSTSEC-0000-0000.md
@@ -1,0 +1,53 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "fulgur"
+date = "2026-07-05"
+url = "https://github.com/fulgur-rs/fulgur/security/advisories/GHSA-4rf6-qx84-q9fv"
+categories = ["denial-of-service"]
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+keywords = ["dos", "pagination", "pdf"]
+aliases = ["GHSA-4rf6-qx84-q9fv"]
+
+[versions]
+patched = [">= 0.26.0"]
+```
+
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

PR #3042, opened 2026-07-06
Add advisory for invalid pointer dereference in fmt::Pointer impl for crossbeam_epoch::{Atomic,Shared}

taiki-e:
## Affected crate(s)

- crossbeam-epoch (recent downloads per crates.io: 112.9M)

- https://github.com/crossbeam-rs/crossbeam/pull/1273 (incomplete fix in 0.9.19)
- https://github.com/crossbeam-rs/crossbeam/pull/1276 (full fix in 0.9.20)

## Severity

- I confirmed that null pointer dereference can be triggered without unsafe code when the opt-level is low or when the pointer is to a DST.
  - 0.9.19's fix addressed this.
- I guess use-after-free is also possible when combined with other unsafe code.
  - 0.9.20's fix addressed this.
    - Another case addressed in 0.9.20 is an unsoundness that can likely be triggered in Miri or sanitizer but is not considered to occur on real hardware. (See the PR 1276 mentioned above for details.)

## Checklist

### DIFF
--- a/crates/crossbeam-epoch/RUSTSEC-0000-0000.md
+++ b/crates/crossbeam-epoch/RUSTSEC-0000-0000.md
@@ -1,0 +1,17 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "crossbeam-epoch"
+date = "2026-07-06"
+url = "https://github.com/crossbeam-rs/crossbeam/pull/1276"
+keywords = ["NULL-pointer-dereference"]
+aliases = []
+
+[versions]
+patched = [">= 0.9.20"]
+unaffected = ["< 0.9.0"]
+```
+
+# Invalid pointer dereference in `fmt::Pointer` impl for `Atomic` and `Shared` when the underlying pointer is invalid
+
+Affected versions of `fmt::Display` dereference the underlying pointer. This causes a invalid pointer dereference e.g., when a pointer created with `Atomic::null` or `Shared::null`. `fmt::Debug` impls and pre-0.9 `fmt::Display` impls, which do not dereference pointers, are not affected by this issue.


crates/crossbeam-epoch/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,20 @@
...
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "crossbeam-epoch"
+date = "2026-07-06"
+url = "https://github.com/crossbeam-rs/crossbeam/pull/1276"
+keywords = ["NULL-pointer-dereference"]
+aliases = []
+
+[affected]
+functions = { "<crossbeam_epoch::Atomic as core::fmt::Pointer>::fmt" = [">= 0.9.0, < 0.9.20"], "<crossbeam_epoch::Shared as core::fmt::Pointer>::fmt" = [">= 0.9.0, < 0.9.20"] }
```
taiki-e: Hmm...
```
error: RustSec error: parse error: error parsing ./crates/crossbeam-epoch/RUSTSEC-0000-0000.md
Caused by:
  -> parse error: error parsing ./crates/crossbeam-epoch/RUSTSEC-0000-0000.md
  -> parse error: TOML parse error at line 10, column 15
   |
10 | functions = { "<crossbeam_epoch::Atomic as core::fmt::Pointer>::fmt" = [">= 0.9.0, < 0.9.20"], "<crossbeam_epoch::Shared as core::fmt::Pointer>::fmt" = [">= 0.9.0, < 0.9.20"] }
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
parse error: invalid character ' ' in identifier: 'Atomic as core'

  -> TOML parse error at line 10, column 15
   |
10 | functions = { "<crossbeam_epoch::Atomic as core::fmt::Pointer>::fmt" = [">= 0.9.0, < 0.9.20"], "<crossbeam_epoch::Shared as core::fmt::Pointer>::fmt" = [">= 0.9.0, < 0.9.20"] }
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
parse error: invalid character ' ' in identifier: 'Atomic as core'
```

taiki-e: Removed.

LawnGnome: Just for the rustsec folks: I don't think we have an exact issue for allowing trait methods here, but rustsec/rustsec#940 touches on it. We should probably allow this and look at pulling something in to replace the ad hoc `FromStr` impl for `FunctionPath`, but I don't think it's a blocker for this particular PR.

LawnGnome (approved):
Thanks!

PR #3047, opened 2026-07-09
Add advisory for bump-scope: wrong documentation potentially leading to use-after-free

bluurryy:
I'm the maintainer of the affected crate.

## Affected crate(s)

- bump-scope (88,910 recent downloads per crates.io)

- [GHSA-xhx6-q96x-pf9c](https://github.com/bluurryy/bump-scope/security/advisories/GHSA-xhx6-q96x-pf9c)
- https://github.com/bluurryy/bump-scope/commit/db9da1967d5d8a22b93eb05ec65f6078a61289f7

## Severity

Soundness issue in safety invariants of `BumpAllocatorCore` trait potentially leading to use-after-free bugs.

This does not affect safe code of the library. It only affects niche, unlikely to be useful or used, safety invariants.

## Checklist

### DIFF
--- a/crates/bump-scope/RUSTSEC-0000-0000.md
+++ b/crates/bump-scope/RUSTSEC-0000-0000.md
@@ -1,0 +1,31 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "bump-scope"
+date = "2026-07-07"
+url = "https://github.com/bluurryy/bump-scope/security/advisories/GHSA-xhx6-q96x-pf9c"
+informational = "unsound"
+categories = ["memory-corruption", "memory-exposure"]
+keywords = ["use-after-free", "unsound"]
+aliases = ["GHSA-xhx6-q96x-pf9c"]
+
+[versions]
+patched = [">= 2.3.2"]
+```
+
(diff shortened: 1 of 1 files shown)


djc:
> This does not affect safe code of the library. It only affects niche, unlikely to be useful or used, safety invariants.

If so, I don't think we need an advisory for it.

bluurryy:
I do believe it is very unlikely that someone encounters that bug. Though I was thinking that for a soundness bug such as this it would be better safe than sorry, given that if someone was bitten by that bug it could hurt a lot. 

I think an `informational = "unsound"` advisory is warranted. The description of `informational = "unsound"` in the [`MAINTAINERS_GUIDE.md`](https://github.com/rustsec/advisory-db/blob/main/MAINTAINERS_GUIDE.md) fits this bug well:

> `informational = "unsound"` is used for [soundness issues](https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#soundness-of-code--of-a-library) that can only be triggered by a programmer (as opposed to e.g. a malicious input), and/or require very contrived code to trigger. They will be surfaced as warnings instead of hard errors by `cargo audit`.

djc:
Please have a look at

bluurryy:
I've read through it and do think that this advisory would have a very low signal to noise ratio. In that sense I'd no longer like to add this advisory.

Thanks for the review!

PR #3065, opened 2026-07-22
Update alternative recommendation for RUSTSEC-2023-0085

Haaremy:
## Summary

Update the recommended alternatives in `RUSTSEC-2023-0085`.

`fluke-hpack` 0.3.1 still reproduces the panic described in the advisory,
whereas `loona-hpack` 0.4.3 returns an error instead.


djc (approved):
Thanks.

PR #3066, opened 2026-07-22
smallstr and gumdrop are unmaintained

djc:
## Affected crate(s)

- smallstr (2.2M recent downloads)
- gumdrop (720k recent downloads)

I contacted the author privately to inquire about the status of gumdrop. This is their response:

> Unfortunately, I don't have time to maintain my open source software projects anymore. So, yes, it would be fair to list gumdrop and any other crates I've uploaded as unmaintained. Thanks for reaching out.

See [their crates](https://crates.io/users/murarth?sort=recent-downloads).

The most popular crate by far is assert_matches. Given that this was just stabilized in std in 1.97, I think now would not be a good time to notify people that is is unmaintained? gumdrop and smallstr are both pretty popular and have popular alternatives available. Beyond that, the remainder of the crates is much less popular.

## Checklist

### DIFF
--- a/crates/gumdrop/RUSTSEC-0000-0000.md
+++ b/crates/gumdrop/RUSTSEC-0000-0000.md
@@ -1,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "gumdrop"
+date = "2026-07-22"
+informational = "unmaintained"
+url = "https://github.com/murarth/gumdrop/issues/63"
+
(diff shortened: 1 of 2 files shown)


LawnGnome (approved):
Makes sense to me.

I'm neutral on the `assert_matches` front: [crates.io already recommends the std macro](https://crates.io/crates/assert_matches), although practically I'm not sure that there's much concern in its maintenance status.

PR #3076, opened 2026-07-27
fix web publishing when we have a cache hit

alex:
Since d9c89372 in rustsec/rustsec, we've been failing to build whenever the admin binary is cached.


djc:
Should merge 

- https://github.com/rustsec/rustsec/pull/1651

and get it deployed here.

alex:
Ah, that's better, missed that there was an open PR for it. Closing in favor of that.

PR #3077, opened 2026-07-27
update rustsec version in publish-web.yml

alex:


djc:
Also need to update the cache key on line 24. Do you want to take care of the other workflows as well? I usually do a workspace-wide search and replace.

alex:
gah, this is what I get for doing this in the web UI. yeah, give me a sec I'll clean it up.

alex:
ok, updated. I'm going to do a follow up PR to refactor all this into a reusable action so there's only one thing to update

djc (approved):

PR #3082, opened 2026-07-28
Add advisory for fontdue: invalid char constructed from untrusted font data

s1ko:
## Affected crate(s)

- `fontdue` (3,236,119 recent downloads, 7,970,747 all-time per crates.io)

- https://github.com/mooman219/fontdue/issues/172 (filed 2026-07-06, open, no maintainer response)

## Severity

Soundness issue — undefined behavior reachable from a safe public API with only
attacker-controlled input.

`Font::from_bytes`, the crate's only font constructor, transmutes each raw `u32`
codepoint from the font's `cmap` table into a `char` without checking that it is
a Unicode scalar value:

```rust
char_to_glyph.insert(unsafe { mem::transmute::<u32, char>(codepoint) }, mapping);
```

`ttf-parser`'s cmap iterators do not filter invalid codepoints: format-4
subtables yield surrogates (`0xD800..=0xDFFF`) and format-12 subtables yield
values above `0x10FFFF`. A crafted font therefore makes `from_bytes` construct an
invalid `char`, which is immediate undefined behavior. The invalid value is kept
as a live `HashMap<char, NonZeroU16>` key and handed back to safe callers by the
public `chars()` accessor.

Confirmed with Miri against the published crates.io releases, using a 240-byte
font with one format-4 segment `startCode = endCode = 0xD800`,
`idRangeOffset = 0`, `idDelta = 1`:

```
error: Undefined Behavior: constructing invalid value of type char:
       encountered 0x0000d800, but expected a valid unicode scalar value
       (in `0..=0x10FFFF` but not in `0xD800..=0xDFFF`)
   --> fontdue-0.9.3/src/font.rs:269:59
```

The UB fires during parsing, before any rasterization, so any application that
loads untrusted or user-supplied fonts is exposed. No memory corruption or code
execution was demonstrated.

Affected from 0.6.4 onward: that is the release in which `char_to_glyph` changed
from `HashMap<u32, NonZeroU16>` to `HashMap<char, NonZeroU16>` and the transmute
was introduced. Reproduced under Miri on both 0.6.4 (with `ttf-parser` 0.12.3)
and 0.9.3 (with `ttf-parser` 0.21.1). Releases up to 0.6.3 key the map by `u32`
and are unaffected. 0.9.3 is the latest release (2025-02-12) and is still
affected, as is current `master`; the crate is marked
`maintenance = "experimental"`, so `patched` is empty.

## Checklist

On the last point: the crate has no security policy and private vulnerability
reporting is disabled, so the report was made on the public issue tracker on
2026-07-06. It has had no maintainer response in over three weeks. This is filed
under the CONTRIBUTING provision for advisories that receive no response
following public disclosure on the repository issue tracker. Happy to hold the
advisory if you would rather wait longer.

Reported by s1ko (s1ko@riseup.net, https://github.com/s1ko).

### DIFF
--- a/crates/fontdue/RUSTSEC-0000-0000.md
+++ b/crates/fontdue/RUSTSEC-0000-0000.md
@@ -1,0 +1,99 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "fontdue"
+date = "2026-07-06"
+url = "https://github.com/mooman219/fontdue/issues/172"
+references = [
+    "https://docs.rs/fontdue/0.9.3/src/fontdue/font.rs.html#269",
+]
+informational = "unsound"
+categories = ["memory-corruption"]
+cvss = "CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:L/A:L"
+keywords = ["font", "cmap", "transmute", "undefined-behavior"]
+
+[affected.functions]
+"fontdue::Font::from_bytes" = [">= 0.6.4"]
+
+[versions]
+patched = []
+unaffected = ["< 0.6.4"]
+```
+
+# `fontdue::Font::from_bytes` constructs an invalid `char` from untrusted font data
+
+`Font::from_bytes` builds the codepoint-to-glyph map by transmuting each raw `u32`
+codepoint read from the font's `cmap` table directly into a `char`, with no
+validity check (`src/font.rs:269` in 0.9.3):
+
+```rust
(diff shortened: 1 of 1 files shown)


djc:
I've reached out to the maintainer via private email, let's wait (for up to a week) to see if that helps.

djc:
Seems like a fix for this issue has been published; the maintainer seems to think this doesn't merit an advisory. If you disagree, suggest to pursue that discussion upstream. Thanks for the report!

PR #3089, opened 2026-07-31
Add advisory for event-listener

Ollie-Pearce:
## Affected crate(s)

- event-listener (recent downloads per crates.io: 92645331)

https://github.com/smol-rs/event-listener/pull/163

## Severity

Unsound `Send`/`Sync` impl on `StackSlot<'_, T>` (stack-allocated listener
from the `listener!` macro) lets a `!Send` tag set via `Event::with_tag`
cross thread boundaries and get accessed via `StackSlot::wait`, causing a
data race in safe code.

## Checklist

- [X] Advisory filename(s) starts with `RUSTSEC-0000-0000` as the ID
- [X] `date` field is set to the public disclosure date
- [X] Contains a concise and descriptive title after advisory metadata

### DIFF
--- a/crates/event-listener/RUSTSEC-0000-0000.md
+++ b/crates/event-listener/RUSTSEC-0000-0000.md
@@ -1,0 +1,24 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "event-listener"
+date = "2026-07-13"
+url = "https://github.com/smol-rs/event-listener/pull/163"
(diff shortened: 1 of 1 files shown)


djc:
@taiki-e do you think an advisory makes sense for this? Or is it unlikely to be a problem in practice?

taiki-e (approved):

djc (approved):

PR #3113, opened 2026-08-06
Add advisory for ithmb-core < 1.9.5 (progressive-JPEG OOM + PhotoDB infinite loop)

B67687:
## Advisory

New advisory for `ithmb-core` (published by `B67687/Ithmb-Codec`): two denial-of-service vulnerabilities reachable from untrusted input, fixed in 1.9.5.

1. **CWE-400**: crafted progressive JPEG (SOF2 65535x65535) triggers ~8 GiB coefficient allocation before any size check → deterministic SIGABRT. Root cause is in `jpeg_decoder` 0.3.2 (allocates during first SOS, before its buffer-limit check at EOI); fix is a `read_info()` pre-flight rejection using a `w*h*11` byte budget.
2. **CWE-835**: crafted PhotoDB container with `mhii total_len==0` makes the chunk walker spin at 100% CPU forever; fix is a `child_end <= pos` bail-out.

Both were found via security research with working PoCs (166-byte and 24-byte fixtures). Filed by the crate author.

## Note

All crates.io versions prior to 1.9.5 were permanently deleted from the registry, so the practical exposure is limited to existing lockfiles and vendored copies — but `cargo audit` should still flag them.

### DIFF
--- a/crates/ithmb-core/RUSTSEC-0000-0000.md
+++ b/crates/ithmb-core/RUSTSEC-0000-0000.md
@@ -1,0 +1,46 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "ithmb-core"
+date = "2026-08-06"
+url = "https://github.com/B67687/Ithmb-Codec"
+categories = ["denial-of-service", "memory-exposure"]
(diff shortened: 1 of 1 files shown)


djc:
You didn't use the pull request template, and this crate doesn't seem to be listed on crates.io at all.

PR #3114, opened 2026-08-06
Add advisory for viperjs: BigInt division panics on a ceiling-exact divisor

MerlijnW70:
Filed by the crate author. The vulnerability was reported privately, fixed, released and
disclosed before this PR — the prerequisites in CONTRIBUTING.md are all met, and step 3's
coordinated disclosure was followed.

- **Advisory:** [GHSA-6976-qm5m-7mcj](https://github.com/MerlijnW70/viperjs/security/advisories/GHSA-6976-qm5m-7mcj) (published)
- **Affected:** `<= 0.2.1` — every version ever published
- **Patched:** [0.2.2](https://github.com/MerlijnW70/viperjs/releases/tag/v0.2.2), on crates.io
- **CVE:** requested through GitHub as CNA; I will follow up with an `aliases` entry once it is assigned

`viperjs` is a JavaScript engine meant to run untrusted script inside a host application, so a
panic reachable from script text is a denial of service in the embedder's process rather than a
robustness nit. A `BigInt` divisor landing exactly on the engine's internal limb ceiling indexed
out of bounds; two neighbouring operations on the same value returned wrong results in silence,
which is why the description names them as well.

The crate is `#![forbid(unsafe_code)]` with no runtime dependencies, so this is a panic and not
memory unsafety.

Happy to adjust the category, the CVSS vector or the wording — this is my first submission here.

### DIFF
--- a/crates/viperjs/RUSTSEC-0000-0000.md
+++ b/crates/viperjs/RUSTSEC-0000-0000.md
@@ -1,0 +1,60 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "viperjs"
+date = "2026-08-06"
+url = "https://github.com/MerlijnW70/viperjs/security/advisories/GHSA-6976-qm5m-7mcj"
+references = ["https://github.com/MerlijnW70/viperjs/releases/tag/v0.2.2"]
+categories = ["denial-of-service"]
+keywords = ["panic", "bigint", "untrusted-input"]
+aliases = ["GHSA-6976-qm5m-7mcj"]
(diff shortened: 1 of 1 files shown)


MerlijnW70:
Following the optional step in CONTRIBUTING.md: **0.2.0 and 0.2.1 are now yanked** on crates.io, so a fresh resolution of `viperjs = "0.2"` picks 0.2.2. Those were the only two affected releases, and there are no dependent crates.

The advisory as filed is unchanged — `patched = [">= 0.2.2"]` still describes it correctly, since a yank does not remove a version from an existing lockfile.

djc (approved):

PR #3120, opened 2026-08-07
Added advisory for sevenz-rust

VLAN-RID:
## Affected crate(s)

- sevenz-rust (495,037 recent downloads per crates.io)

Repository for library has been deleted and project has been abandoned.

## Severity
Arbitrary file write from no path verification in decompress_impl potentially allowing for remote command execution.

## Checklist

- [x] Asked maintainer(s) if publishing an advisory is appropriate (unable to find maintainer contact due to author's github being deleted)


djc:
Thanks for reporting!

I think this should be two advisories, one `unmaintained` and one non-`informational`.

For the security issue, it would be nice to have more details in the advisory given the lack of an upstream issue or PR.

VLAN-RID:
@djc I have updated the advisory with more details and explicitly specified absolute path traversal in the advisory heading.
What would be the difference between a `unmaintained` and a non-informational advisory beyond the toml field?

djc:
> @djc I have updated the advisory with more details and explicitly specified absolute path traversal in the advisory heading. What would be the difference between a `unmaintained` and a non-informational advisory beyond the toml field?

One of them is about the vulnerability, declaring all existing versions as vulnerable. The other is for declaring the crate to be unmaintained. That will be reflected on crates.io.

crates/sevenz-rust/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,60 @@
...
+                "normal.txt")
+```
+2. Build the rust program with `sevenz-rust` as a dependency.
+```rust
+fn main() {
+    sevenz_rust::decompress_file("slip.7z", "sample").expect("complete");
+}
+```
+3. Run the rust program in the same directory as the archive.
+4. View `sample/normal.txt`, `relative.txt` adjacent to `sample`, and `C:/Temp/absolute.txt` which will be now written to disk.
```
djc: Let's drop this section, I think the details section is enough.

VLAN-RID: @djc Removed the section 👍

VLAN-RID:
@djc I removed the informational tag on this branch and will make a new pr with an advisory simply stating that the crate is unmaintained.

VLAN-RID:
Created informational advisory at #3132

djc (approved):

PR #3124, opened 2026-08-09
Document withdrawal of dcrypt 2.0.0

NoCentralHub:
## Summary

dcrypt 2.0.0 has been withdrawn and yanked because it violates the project's
zero-unsafe, zero-native-code, and zero-FFI implementation policy. Version
1.2.3 has also been yanked and is not a safe fallback. No maintainer-supported
patched release is currently available.

This follow-up:

- adds a prominent withdrawal warning to all four dcrypt advisories;
- replaces the deleted-release reference with the durable upstream withdrawal
  notice; and
- removes recommendations to deploy 2.0.0 while preserving the technically
  accurate vulnerability ranges and `patched >= 2.0.0` metadata for these four
  specific flaws.

The withdrawal is a separate implementation-policy correction. It does not
assert that these four named vulnerabilities persist in 2.0.0, and the
immutable v2.0.0 tag remains available as historical provenance.

Upstream notice:
https://github.com/ioi-foundation/dcrypt/blob/master/docs/security/V2.0.0-WITHDRAWAL.md

## Validation

- `rustsec-admin lint`
- `git diff --check`

### DIFF
--- a/crates/dcrypt-algorithms/RUSTSEC-0000-0000.md
+++ b/crates/dcrypt-algorithms/RUSTSEC-0000-0000.md
@@ -6,7 +6,7 @@ date = "2026-08-09"
 url = "https://github.com/ioi-foundation/dcrypt/security/advisories/GHSA-h9f2-fgp8-vc4h"
 references = [
     "https://github.com/ioi-foundation/dcrypt/commit/c99cc86f0ee353010cd202cbcd2c310371b0bbb8",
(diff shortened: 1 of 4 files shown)


djc:
Needs a rebase after ID assignment.

Please avoid relying more on advisory-db review until you're more certain all the issues have been resolved.

(Personally I would suggest using graviola instead of reimplementing all this crypto stuff.)

Issue #3126, opened 2026-08-09
Malformed advisory breaks database loading for all consumers: RUSTSEC-0000-0000 filed under crates/gettext-rs but declares package = "gettext-sys"


fastrevmd-lab:
`cargo audit` currently fails to load the advisory database entirely, so it scans nothing and exits with an error rather than a verdict.

```
error: error loading advisory database: git operation failed:
expected RUSTSEC-0000-0000 to be in gettext-sys directory
(instead of "/home/.../.cargo/advisory-db/crates/gettext-rs")
```

## Cause

Commit e11d6b33 ("Add advisory for gettext-rs (unsound setlocale)", 2026-08-06) added:

`crates/gettext-rs/RUSTSEC-0000-0000.md`

whose front matter declares a different package:

```toml
[advisory]
id = "RUSTSEC-0000-0000"
package = "gettext-sys"
```

The loader requires the containing directory to match `package`, so `gettext-sys` in `crates/gettext-rs/` is rejected — and because the check aborts the load, **one malformed entry takes down the whole database** for every consumer, not just anyone depending on gettext.

The placeholder `RUSTSEC-0000-0000` id suggests this was merged before assignment.

## Impact

Any project running `cargo audit` against a freshly fetched database gets an error instead of a scan. In CI that surfaces as a failing supply-chain job with no findings, which is easy to misread as "a vulnerability was found" or, worse, to wave through as flaky. It is neither: the gate did not run.

Reproduced with `cargo-audit 0.22.2` on a clean `~/.cargo/advisory-db`.

## Suggested fix

Either move the file to `crates/gettext-sys/`, or change `package` to `gettext-rs`, depending on which crate the advisory is actually about — the affected functions are named `gettextrs::setlocale` and `gettextrs::TextDomain::init`, while `package` says `gettext-sys`, so those two are worth reconciling as part of the fix. Assigning a real `RUSTSEC-YYYY-NNNN` id would also stop the placeholder recurring.

A CI check that each advisory's `package` matches its directory would prevent a single bad entry from breaking every downstream consumer again.

djc:
Working on it. We have CI checks but they didn't fire correctly.

- https://github.com/rustsec/advisory-db/pull/3125

PR #3119, opened 2026-08-06
Add advisory for gettext-rs (unsound setlocale)

Minoru:
## Affected crate(s)

- gettext-rs (about 5000 downloads per day)

## Links to upstream issue(s) or PR(s)

- https://github.com/gettext-rs/gettext-rs/issues/64

## Severity

Unsoundness can lead to memory corruption and crashes. (I could not reproduce the crash, but the linked issue claims that the binary segfaults and I believe it to be a likely outcome.)

## Checklist

- [x] Advisory filename(s) starts with `RUSTSEC-0000-0000` as the ID
- [x] `date` field is set to the public disclosure date -- I used today's date, although the fixes were released yesterday and the issue was filed 5 years ago
- [x] Contains a concise and descriptive title after advisory metadata
- [x] Asked maintainer(s) if publishing an advisory is appropriate -- I am the maintainer

Fixes #3073.

PR #3128, opened 2026-08-09
Move RUSTSEC-2026-0244 to correct crate

Minoru:
In https://github.com/rustsec/advisory-db/pull/3119 I specified the wrong `package` ("gettext-sys" instead of "gettext-rs"). This caused problems for downstream users (https://github.com/rustsec/advisory-db/issues/3126). https://github.com/rustsec/advisory-db/pull/3125 attempted a fix but got it wrong (updated everything to align with the wrong `package`). This PR is my attempt to fix it properly.

/cc @djc

### DIFF
--- a/crates/gettext-rs/RUSTSEC-2026-0244.md
+++ b/crates/gettext-rs/RUSTSEC-2026-0244.md
@@ -8,8 +8,8 @@ informational = "unsound"
 categories = ["memory-corruption", "thread-safety"]
 
 [affected.functions]
-"gettext-rs::setlocale" = ["< 0.8.0"]
-"gettext-rs::TextDomain::init" = ["< 0.8.0"]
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

Minoru:
Sorry about the mess I've created.

AlastairHolmes:
If you are seeing this error:
```
error: error loading advisory database: parse error: duplicate advisory ID: RUSTSEC-2026-0244
```

I think you need to delete your existing advisories db (~/.cargo/advisory-db) or specifically the `gettext-sys` folder, as cargo-audit will not delete the existing wrong RUSTSEC-2026-0244 files, and so errors with duplicate ID.

Just putting this here, if anyone else has the same problem.

PR #3130, opened 2026-08-09
Add note re hyphens in crate names

Minoru:
In https://github.com/rustsec/advisory-db/pull/3128 I had to try a few different ways to mangle "gettext-rs". This commit ensures that future contributors don't have to re-discover that knowledge.

I realise that turning hyphens into underscores is Cargo convention, but I got confused anyway. Perhaps the fact that gettext-rs defines a library name of "gettextrs" contributed to that.

I'd open an issue first to discuss this, but it felt like too small a change to warrant that. Please don't feel pressured to merge this!

### DIFF
--- a/EXAMPLE_ADVISORY.md
+++ b/EXAMPLE_ADVISORY.md
@@ -20,6 +20,7 @@ keywords = ["example", "freeform", "keywords"]
 #os = ["windows"]
 
 #[affected.functions]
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Yup, we have this issue:

- https://github.com/rustsec/rustsec/issues/1575

PR #3148, opened 2026-08-13
Add advisory for tiberius: panics on hostile server input

north-echo:
## Affected crate(s)

- `tiberius` (622,315 recent downloads per crates.io, 97 reverse dependencies)

- https://github.com/prisma/tiberius/issues/424
- https://github.com/prisma/tiberius/issues/425

Both filed 2026-07-29 and neither has received a maintainer response in the fifteen days since. A private advisory was filed the same day through the repository's private vulnerability reporting; it remains unpublished, which is why no resolvable GHSA or CVE alias is listed.

## Severity

Remote denial of service. Eleven decode paths respond to server-controlled values with `panic!`, `unimplemented!`, `todo!` or `.unwrap()` on `None` rather than returning `Err`. Three are in the PRELOGIN exchange, which is the first message of a TDS connection and the one that negotiates encryption, so they are reachable before authentication and before TLS. A single server-controlled byte turns a recoverable connection error into a process-level panic; under `panic = "abort"` that is a hard abort, and in pooled deployments it surfaces inside a pool worker rather than as an error the application can catch.

All eleven were confirmed against the published 0.12.3 crate in a **release** build, with `debug_assertions` off, driving only the public `Client::connect` API against a listener speaking TDS. Every site is still present on `main` at `a6b4fcd`, verified 2026-08-13.

No fixed version exists. The most recent release, 0.12.3, is from 2024-07-19, so `cargo audit` visibility is the only practical warning mechanism for the 97 dependents.

## On the qualifying criteria

Stated plainly rather than stretched: **tiberius makes no panic-free claim**, so the main clause of that bullet is not satisfied. The case rests on the parenthetical and on one other point.

The panics are useful for network denial of service. They are unauthenticated and remote, three are pre-auth and pre-TLS, and several fire on a single byte. One is a bare `assert!` rather than `debug_assert!`, so it is live in release builds and covers 255 of 256 byte values.

Separately, three sites are reached by values the crate's own `try_from` conversions **accept as valid TDS types**. `ColInfo`, `Udt` and `SSVariant` are real protocol types that are simply unimplemented, so a conforming server triggers them without sending anything malformed. Genuinely unknown bytes are already rejected correctly with `Error::Protocol`. The README lists "a perfect implementation of the TDS protocol" as a goal, so this is a deviation from stated intent rather than a hardening wish.

Happy to convert this to `informational` if you prefer that framing, and happy to adjust the CVSS. Scoring the malicious-or-MITM-server precondition as `AC:H` gives 5.9 and is equally defensible; I used `AC:L` on the basis of an application connecting to a SQL Server it does not control.

RUSTSEC-2020-0010 already exists for this crate but is an `unmaintained` informational scoped `unaffected = ["> 0.3.2"]`, covering the original 0.3 line rather than the current fork, so there is no overlap.

## Checklist

- [x] `date` field is set to the public disclosure date (2026-07-29, when #424 and #425 were filed)
- [x] Asked maintainer(s) if publishing an advisory is appropriate — reported upstream publicly and privately on 2026-07-29; no response in fifteen days, which follows the documented fallback

This advisory was researched and written with AI assistance; I verified every site against the published crate in a release build and reviewed the text before submitting.

### DIFF
--- a/crates/tiberius/RUSTSEC-0000-0000.md
+++ b/crates/tiberius/RUSTSEC-0000-0000.md
@@ -1,0 +1,94 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "tiberius"
+date = "2026-07-29"
+url = "https://github.com/prisma/tiberius/issues/424"
+references = [
+    "https://github.com/prisma/tiberius/issues/424",
+    "https://github.com/prisma/tiberius/issues/425",
+]
+categories = ["denial-of-service"]
+keywords = ["tds", "mssql", "sql-server", "parser", "panic", "dos"]
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+# aliases = []   # a GHSA was filed privately on 2026-07-29 but the maintainer has
+#                # not published it, so there is no resolvable alias or CVE yet
+
+[versions]
+patched = []            # no fixed release exists; latest is 0.12.3 (2024-07-19)
+```
+
+# tiberius panics on malformed or unexpected responses from a SQL Server (denial of service)
+
+`tiberius` decodes the TDS protocol from the server without treating server bytes as
+untrusted input. Eleven code paths respond to unexpected server-controlled values with
+`panic!`, `unimplemented!`, `todo!`, or `.unwrap()` on `None` rather than returning `Err`.
+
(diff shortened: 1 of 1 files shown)


djc:
Dropped a comment in

- https://github.com/prisma/tiberius/issues/427

Please ping me in a week if there's been no movement.

djc:
Going to give the new maintainers some (more) time to get a release out.

PR #3163, opened 2026-08-20
add advisories for arone, aronenao and tinymember

ubiratansoares:
## Affected crate(s)

- `arone`
- `aronenao`
- `tinymember`

N/A

## Severity

During tha [arrayref attack](https://blog.rust-lang.org/2026/08/20/supply-chain-attack-on-arrayref) these three were flagged as suspicious. They were owned by the same user.

## Checklist


djc (approved):

djc:
Thanks!

PR #3167, opened 2026-08-21
Add advisory for simd-json uncontrolled recursion during Value materialization

dngen:
## Affected crate(s)

- `simd-json` 0.17.3 (confirmed affected)

- Upstream report and maintainer confirmation:
  https://github.com/simd-lite/simd-json/issues/473
- Fix commit:
  https://github.com/simd-lite/simd-json/commit/abb59c3bebf8cde1f0b974133a960ab3a968891d

The maintainer confirmed that the issue is fixed in version 0.18.

date = "2026-08-13"

[versions]
patched = [">= 0.18.0"]

## Severity

Deeply nested JSON passed through `to_owned_value` or
`to_borrowed_value` can cause uncontrolled recursion during tape-to-Value
materialization, resulting in native stack exhaustion and process abort.

The confirmed impact is availability only at the library level. No specific
downstream network service is claimed. The proposed severity is Low.

## Checklist

- [x] Advisory filename starts with `RUSTSEC-0000-0000`
- [x] Asked maintainer whether publishing an advisory is appropriate

### DIFF
--- a/crates/simd-json/RUSTSEC-0000-0000.md
+++ b/crates/simd-json/RUSTSEC-0000-0000.md
@@ -2,12 +2,12 @@
 [advisory]
 id = "RUSTSEC-0000-0000"
 package = "simd-json"
-date = "2026-08-13"
+date = "2026-08-14"
 categories = ["denial-of-service"]
 keywords = ["json", "parser", "dos", "stack-overflow", "recursion"]
 
 [versions]
-patched = []
+patched = [">= 0.18.0"]
 ```
 
 # Unbounded recursion in `to_owned_value` / `to_borrowed_value` allows stack overflow from deeply nested JSON
@@ -62,8 +62,3 @@ representative harness.
 Bound recursion depth in the tape to `Value` materialization (return an error past a default limit;
 for reference serde_json's default recursion limit is 128), or make the materialization iterative
 with an explicit stack like stage2.
-
-## Duplicate check
-
-Not RUSTSEC-2019-0008 (that is an out-of-bounds read in string parsing, fixed in 0.1.15). No other
-RustSec/CVE advisory found, and no matching upstream issue or PR for this mechanism as of 2026-08-13.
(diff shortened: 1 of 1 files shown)


crates/simd-json/RUSTSEC-0000-0000.md
```diff
@@ -0,0 +1,69 @@
...
+## Suggested fix
+
+Bound recursion depth in the tape to `Value` materialization (return an error past a default limit;
+for reference serde_json's default recursion limit is 128), or make the materialization iterative
+with an explicit stack like stage2.
+
+## Duplicate check
+
+Not RUSTSEC-2019-0008 (that is an out-of-bounds read in string parsing, fixed in 0.1.15). No other
+RustSec/CVE advisory found, and no matching upstream issue or PR for this mechanism as of 2026-08-13.
```
djc: Let's drop all this crap? I don't think it adds value for consumers (and seems like LLM slop).

djc (commented):
Thanks for the report!

I'd like to see explicit signoff from the maintainer on whether publishing an advisory for this is a good idea.

You haven't put the patched version in the advisory metadata.

crates/simd-json/RUSTSEC-0000-0000.md
dngen: Agreed, removed It is review provenance rather than actionable information for advisory consumers.

dngen:
> Thanks for the report!
> 
> I'd like to see explicit signoff from the maintainer on whether publishing an advisory for this is a good idea.
> 
> You haven't put the patched version in the advisory metadata.

Thanks. I’ve added >= 0.18.0 as the patched version and removed the duplicate-check section. I’ve also asked the upstream maintainer for explicit confirmation that publishing a RustSec advisory is appropriate. I’ll update this PR once they respond.

djc (commented):
I actually think we should remove all sections except the first since I don't think any of them are very actionable (but apparently GitHub butchered my selection).

dngen:
> ered my selection).

thanks for clarifying. I reduce body to the opening section only.

PR #3171, opened 2026-08-24
Add advisory for zkboo-ecc: incorrect fixed-base scalar multiplication when the comb window width divides the scalar width

sg495:
Filed by the crate author.

`zkboo-ecc`'s `Curve::mul_secret_scalar` recodes a secret scalar into one signed digit per comb window, carrying a residual that the final window consumes whole. The window count was `width.div_ceil(w)`, which leaves the final window no headroom when the window width `w` divides the scalar width — the residual can then exceed what the final window's `w - 1` index bits can hold, its top bit is dropped, and the routine silently returns the wrong curve point.

For the 256-bit secp256k1 scalar this affects `w` in {2, 4, 8, 16}. Measured against a native reference, 5 of 14 boundary scalars were wrong at both `w = 4` and `w = 8`.

The default window width is 5, which does not divide 256, so every convenience entry point in `zkboo-ecc` and in the downstream `zkboo-bip32` is unaffected and bit-identical across the fix. The width is only reachable by callers supplying their own `WindowTables`.

Scope note: this is a correctness failure in a cryptographic primitive, not a soundness break of the underlying zero-knowledge proof system — prover and verifier run the same circuit, so an affected proof still verifies, it simply attests to a different statement than intended (e.g. a wrong derived public key or address).

Fixed in 1.0.1 (`zkboo/zkboo-ecc@7b9f17e`); 0.1.0 predates comb multiplication and is unaffected.

### DIFF
--- a/crates/zkboo-ecc/RUSTSEC-0000-0000.md
+++ b/crates/zkboo-ecc/RUSTSEC-0000-0000.md
@@ -1,0 +1,64 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "zkboo-ecc"
+date = "2026-08-24"
+url = "https://github.com/zkboo/zkboo-ecc/commit/7b9f17ec0003d65d254d58d73b31440f72de9ad6"
+categories = ["crypto-failure"]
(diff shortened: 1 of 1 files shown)


djc:

PR #3175, opened 2026-08-26
Add advisory for physis_pro: remote error body can permanently disable a shared LLM provider

ilPez00:
Filed by the crate author, which per CONTRIBUTING waives the upstream-report and coordinated-disclosure prerequisites. The embargo is lifted: the fix, the affected range and the mechanism are already public in the crate's own repository and changelog.

**Summary.** `physis_pro` routes completions through a cascade of LLM providers. In `0.1.6`–`0.1.8` a provider was removed from the cascade by substring-matching the *text* of a failed HTTP response. OpenAI-compatible endpoints quote the offending request back inside `error.message`, and that request carries the caller's prompt — so a prompt containing a phrase like `does not exist` that provokes any 4xx would match. The removal was permanent (`mark_unavailable()` wrote a flag only `Provider::new()` ever cleared) and the cascade is process-wide, so the loss hit every caller rather than the one who caused it. Walking the cascade reaches `No provider available` for the life of the process.

Availability only — no disclosure or modification of data. Rated `AC:H` because success depends on the upstream provider echoing request content in its error bodies: common among OpenAI-compatible gateways, but a property of the vendor rather than of this crate.

**Fixed in `0.1.9`.** Detection now reads only the machine-written `error.code` / `error.type` fields against a fixed set of vendor codes, never the message prose, and skips bodies that are not JSON. The park became a bounded 15-minute cooldown rather than a permanent latch.

**Versions.** `patched = [">= 0.1.9"]`, `unaffected = ["< 0.1.6"]`. The lower bound was taken from the published tarballs rather than from git history — `0.1.5` carries only a pre-existing `401`/`403` latch, which keys on a status code rather than on caller-influenced text, so it is not vulnerable. `0.1.6`, `0.1.7` and `0.1.8` have been yanked.

`rustsec-admin lint` reports `Linted ok` for this file.

Full write-up: https://github.com/ilPez00/physis-pro/blob/main/security/advisories/physis_pro-provider-availability.md

### DIFF
--- a/crates/physis_pro/RUSTSEC-0000-0000.md
+++ b/crates/physis_pro/RUSTSEC-0000-0000.md
@@ -1,0 +1,90 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "physis_pro"
+date = "2026-08-26"
+url = "https://github.com/ilPez00/physis-pro/blob/main/security/advisories/physis_pro-provider-availability.md"
+categories = ["denial-of-service"]
+cvss = "CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:H"
+keywords = ["dos", "availability", "multi-tenant", "llm"]
+
+[versions]
+patched = [">= 0.1.9"]
+unaffected = ["< 0.1.6"]
+```
+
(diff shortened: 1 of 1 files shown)


djc:
Please use the pull request template, and don't have your agent write the PR description.

PR #3177, opened 2026-08-27
Add unsound advisory for hibit_tree

tooson9010-spec:
## Affected crate(s)
- `hibit_tree` (123 recent downloads per crates.io)

https://github.com/tower120/hibit_tree/pull/28

## Severity
`BlockPtr::insert_impl` set the occupancy bit in `mask` before running the child constructor and wrote the index mapping only afterwards, so a panicking `Default::default()` left the two halves of the metadata disagreeing — two logical indices resolved to the same physical slot and `Drop for Tree` destroyed it twice. Use-after-free / double-free reachable from safe Rust via `Tree::get_or_insert`, confirmed under AddressSanitizer. Fixed in 0.1.1-beta.1.

## Checklist
- [x] Asked maintainer(s) if publishing an advisory is appropriate (asked on the PR; the maintainer responded by releasing the fix as 0.1.1-beta.1)

### DIFF
--- a/crates/hibit_tree/RUSTSEC-0000-0000.md
+++ b/crates/hibit_tree/RUSTSEC-0000-0000.md
@@ -1,0 +1,52 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "hibit_tree"
+date = "2026-08-27"
+url = "https://github.com/tower120/hibit_tree/pull/28"
+categories = ["memory-corruption"]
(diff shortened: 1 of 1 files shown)


djc:
At 126 recent, 4k all-time downloads and no dependents, I think this fails our recent notability criteria:

tooson9010-spec:
@djc 
That's fair, thanks. Happy to withdraw this one; the maintainer already
shipped the fix in 0.1.1-beta.1, so users are covered either way.

PR #3178, opened 2026-08-27
Add unsound advisory for stack_dst

tooson9010-spec:
## Affected crate(s)
- `stack_dst` (5,257 recent downloads per crates.io)

https://github.com/thepowersgang/stack_dst-rs/issues/16

## Severity
Three removal/replacement paths destroy an initialized value before committing the metadata change that removes it. A panicking `T::drop()` skips the commit, so the container still treats the destroyed value as live and drops it again — use-after-free / double-free reachable from safe Rust, confirmed under AddressSanitizer on all three paths. Separate from RUSTSEC-2021-0033, which covered `push_cloned`. Fixed in 0.8.2.

## Checklist
- [x] Asked maintainer(s) if publishing an advisory is appropriate (asked on the issue; the maintainer approved with a 👍 reaction)

### DIFF
--- a/crates/stack_dst/RUSTSEC-0000-0000.md
+++ b/crates/stack_dst/RUSTSEC-0000-0000.md
@@ -1,0 +1,56 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "stack_dst"
+date = "2026-08-27"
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

PR #3181, opened 2026-08-28
Add unmaintained advisory for `lasso`

odilf:
Author unresponsive since 2024.

## Affected crate(s)

- `lasso` (871k recent downloads per crates.io)

https://github.com/Kixiron/lasso/issues/53

## Checklist

Regarding claiming implicit unmaintenance, the maintenance status issue has only been open since May, which AFAIU would generally not be long enough to be considered unmaintained. However
- a [non-trivial issue](https://github.com/Kixiron/lasso/issues/48) (with proposed solution) has been opened since Jun 3rd, 2024, with no maintainer response
- author has not had any GitHub activity at all since Aug 19th, 2024 

It is mostly the fact that there is no signs of the maintainer for two years that makes me think it's ok to mark the crate as unmaintained. Their website linked on GitHub is also defunct, so it's hard to contact them.

Of course, I would be happy to give it a few months more for a response before adding it to the advisory. This is my first time reporting something, so feel free to let me know what/whether it is appropriate :)

### DIFF
--- a/crates/lasso/RUSTSEC-0000-0000.md
+++ b/crates/lasso/RUSTSEC-0000-0000.md
@@ -1,0 +1,20 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "lasso"
+date = "2026-08-28"
+url = "https://github.com/Kixiron/lasso/issues/53"
+informational = "unmaintained"
+
+[versions]
+patched = []
+```
+
+# `lasso` is unmaintained
+
+The author of `lasso` is unresponsive, with no GitHub activity since 2024 and outstanding issues such as [one from Jun 2024](https://github.com/Kixiron/lasso/issues/48).
+
+Maintained alternatives:
+
+- [`string-interner`](https://crates.io/crates/string-interner)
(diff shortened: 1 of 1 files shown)


djc:
Have reached out to the maintainer via private email, going to see if they respond.

If there is no action (from the maintainer or me) in this PR or the upstream issue in the next week, feel free to ping me.

odilf:
pinging, was there any response?

djc:
Now we're going to wait for ~270 days:

> To justify the "implicitly unmaintained" status, where a crate author is unreachable, the following criteria must be met:
>
> * Stale repository: no recent maintenance activity, including any of the following: recent commits, responses from the author on open issues, crate releases, or other publicly visible activity by the author. Inactivity over a period of 1 year or more is the preferred threshold.
> * Contact attempts with the author made with no response. Ideally these attempts are made via a public GitHub issue, so that issue can be cited in an unmaintained crate advisory if need be. Unresponsiveness by the author over a period of 270 days (or 60 days of unresponsiveness after being notified of a vulnerability) is the minimum before a crate will be considered unmaintained.

PR #3190, opened 2026-08-31
Add information about GHSA-f42v-x7gq-phc8

alexanderkjall:
## Affected crate(s)

- sudo-rs (recent downloads per crates.io)

@squell is it ok if this is published in rustsec?

information copied from:

https://github.com/trifectatechfoundation/sudo-rs/security/advisories/GHSA-f42v-x7gq-phc8

## Severity

local privilege escalation

## Checklist

### DIFF
--- a/crates/sudo-rs/RUSTSEC-0000-0000.md
+++ b/crates/sudo-rs/RUSTSEC-0000-0000.md
@@ -1,0 +1,54 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "sudo-rs"
+date = "2026-08-31"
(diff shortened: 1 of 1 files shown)


squell:
> @squell is it ok if this is published in rustsec?

I have no objection to this! RUSTSEC is important, and I do believe one other of our advisories is also in RUST-SEC. On the other hand, some others aren't, so if you take this you might consider adding the others as well.

I don't know whether it is "appropriate" for you though--sudo-rs as published on crates.io isn't meant to be a dependency of other crates; i.e. it's not something anyone would add as "cargo install sudo-rs" (that wouldn't even work properly). I don't know if that falls under the scope of RUSTSEC, to be honest.

djc:
We do consider binary crates in scope for the advisory database.

PR #3203, opened 2026-09-08
Update CVSS RUSTSEC-2026-0279.md

AidenMohan:
just adding a cvss string for it to be consistent with a cve cna-lr request or github GHSA
also added credits for the same purpose
thanks!


djc (approved):

AidenMohan:
thanks!

PR #3204, opened 2026-09-09
Add advisory for aligned_box: panic-safety double-free in realloc_with_default

tooson9010-spec:
## Affected crate(s)

- `aligned_box` (19,224 recent downloads per crates.io, 3 reverse dependencies)

Reported and fixed in [michaellass/aligned_box#6](https://github.com/michaellass/aligned_box/pull/6), released in 0.3.1 on 2026-09-08.

## Severity

Panic-safety unsoundness when shrinking an `AlignedBox<[T]>` via `realloc_with_default`: ownership of the buffer is taken out of `self.container` before the tail elements are dropped, and only committed back afterwards. A panicking element `Drop` skips the commit and leaves `self.container` pointing at the partially destroyed buffer, which the destructor then drops again and deallocates — a double free (CWE-415) / use-after-free (CWE-416) reachable from safe Rust, confirmed with glibc's `double free detected in tcache 2`. Fixed in 0.3.1.

## Checklist

- [x] Asked maintainer(s) if publishing an advisory is appropriate (maintainer approved in the PR thread)

### DIFF
--- a/crates/aligned_box/RUSTSEC-0000-0000.md
+++ b/crates/aligned_box/RUSTSEC-0000-0000.md
@@ -1,0 +1,38 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "aligned_box"
+date = "2026-09-09"
+url = "https://github.com/michaellass/aligned_box/pull/6"
+categories = ["memory-corruption"]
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

PR #3205, opened 2026-09-09
Add advisory for owned-alloc: panic-safety double-free in OwnedAlloc::drop_in_place

tooson9010-spec:
## Affected crate(s)

* `owned-alloc` (736,101 recent downloads per crates.io)

Reported in [bzim/owned-alloc#1](https://gitlab.com/bzim/owned-alloc/-/issues/1) (2026-09-08). Same maintainer and account as `lockfree`, where a public issue and a private email have both gone unanswered; no release since 2018.

## Severity

Panic-safety unsoundness in `OwnedAlloc::drop_in_place`: the contained value is destroyed by hand, and the `mem::forget(self)` that commits the ownership transfer lives inside `into_raw` and so runs only afterwards. A panicking `T::drop` skips the commit, so the still-live `OwnedAlloc` unwinds and its destructor drops the same `T` a second time before deallocating — a double free (CWE-415). The destructor also reads the destroyed value through `Layout::for_value` on the way (CWE-416). Confirmed under AddressSanitizer (`attempting double-free`).

`MaybeUninitAlloc::drop_in_place` delegates to the same function, so both public entry points are affected. Storing a value whose `Drop` can panic and calling either is enough; no `unsafe` on the caller's side.

## Checklist

- [x] Asked maintainer(s) if publishing an advisory is appropriate (reported upstream)

### DIFF
--- a/crates/owned-alloc/RUSTSEC-0000-0000.md
+++ b/crates/owned-alloc/RUSTSEC-0000-0000.md
@@ -1,0 +1,39 @@
+```toml
+[advisory]
+id = "RUSTSEC-0000-0000"
+package = "owned-alloc"
+date = "2026-09-09"
+url = "https://gitlab.com/bzim/owned-alloc/-/issues/1"
+categories = ["memory-corruption"]
+keywords = ["panic-safety", "memory-safety", "use-after-free", "double-free"]
+
+[affected]
+[affected.functions]
+"owned_alloc::OwnedAlloc::drop_in_place" = ["<= 0.2.0"]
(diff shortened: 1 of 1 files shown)


tooson9010-spec:
@djc 

This crate has the same maintainer and account as lockfree (#3174, already merged), with no release since 2018, so it seems effectively unmaintained. Could you let me know how you'd like me to proceed with this one?

djc (approved):
Thanks for the ping; merging.

Can you additionally do an unmaintained advisory for this?

tooson9010-spec:
@djc 

Thanks for merging! Opened the unmaintained advisory in #3239.

PR #3208, opened 2026-09-12
RUSTSEC-2023-0071: record 0.9.10 / 0.10-rc still unpatched

SebTardif:
`patched = []` stays empty. There is still no fixed rsa release.

This updates the human text and references so cargo-audit readers are not left on the old "work is underway to migrate to crypto-bigint" story:

- 0.9.10 (latest stable) and 0.10.0-rc.18 (latest) are still affected
- RustCrypto/RSA#394 did not close Marvin
- Tracking remains RustCrypto/RSA#626
- In-progress: RustCrypto/RSA#680 (implicit rejection), RustCrypto/RSA#702 (default-path blinding)


djc (commented):
I don't think we should talk about a particular otherwise unrelated crate in this advisory.

SebTardif:
@djc

> I don't think we should talk about a particular otherwise unrelated crate in this advisory.

Agreed. This advisory is for rsa. I dropped the jsonwebtoken paragraph. Downstream JWT backend advice does not belong here.

djc (approved):

PR #3214, opened 2026-09-14
Mark chrono and friends as unmaintained

djc:
chrono and friends are now unmaintained. They have been soft-deprecated since January:

- https://github.com/chronotope/chrono/issues/1768


weiznich:
Would it be possible to postpone this until jiff issued an 1.0 release?

They seem to be rather close to this https://github.com/BurntSushi/jiff/pull/622 and I guess this would prevent quite a bit of ecosystem churn of not having people now migrating to jiff 0.2 and then again to 1.0.

djc:
> Would it be possible to postpone this until jiff issued an 1.0 release?

Sounds like that might make sense.

> They seem to be rather close to this https://github.com/BurntSushi/jiff/pull/622 and I guess this would prevent quite a bit of ecosystem churn of not having people now migrating to jiff 0.2 and then again to 1.0. 

@BurntSushi what's your 1.0 planning?

BurntSushi:
I would hold off on this until 1.0 is out. I don't have an ETA yet unfortunately.

djc:
> I would hold off on this until 1.0 is out. I don't have an ETA yet unfortunately.

Can you say if it's more like O(weeks) or O(1-5 months) or...?

BurntSushi:
Months.

LawnGnome:
I think we should wait for `jiff` 1.0 as well, and not just because I'm personally dreading having to internalise a new, non-`chrono` API. :slightly_smiling_face:

PR #3218, opened 2026-09-15
Add advisory for fixed-queue VecDeque::clear double free

tooson9010-spec:
## Affected crate(s)

- `fixed-queue` (967 recent downloads per crates.io)

Reported and fixed in https://github.com/rise0chen/fixed-queue/pull/1, merged
Aug 10, released in 0.5.2.

## Severity

Panic-safety unsoundness in `VecDeque::clear`: the elements are dropped before
`end` / `start` / `is_full` are reset, so a panicking element `Drop` leaves the
metadata stale. `Drop for VecDeque` calls `clear()` again, so dropping the queue
during unwinding revisits the already-dropped slots — a double free (CWE-415) /
use-after-free (CWE-416) reachable from safe Rust, confirmed under
AddressSanitizer. Fixed in 0.5.2.

## Checklist

- [x] Asked maintainer(s) if publishing an advisory is appropriate — asked twice
      on the merged PR (Aug 11 and again about two weeks ago), no response


tooson9010-spec:
Recent downloads are 967, just under the 1,000 threshold, with 3 reverse
dependencies. Happy to withdraw if that's below the bar.

djc:
Yeah, I think this fails notability (because the dependents are from the same author).

tooson9010-spec:
@djc 

That makes sense. Thanks for checking and the guidance. I'll keep the notability bar (and same-author dependents) in mind for future submissions.

PR #3226, opened 2026-09-18
Unsoundness in iceoryx2-bb-container

dkroenke:
## Affected crate(s)

- iceoryx2-bb-container (373,431 recent downloads per crates.io)

https://github.com/eclipse-iceoryx/iceoryx2/security/advisories/GHSA-8mq4-3mwq-qvg6

## Severity

The String API in iceoryx2-bb-container exposes its contents as mutable bytes through safe APIs, while String::as_str() converts those bytes into a Rust string slice without validating UTF-8. 
An application can therefore create an invalid &str and trigger undefined behavior using entirely safe Rust.
Fixed with version 0.10.0

## Checklist


dkroenke:
FYI @elBoberido

djc (approved):
Thanks!

PR #3230, opened 2026-09-21
Add advisory for ringbuf: panic-safety double-free in Consumer::skip/…

tooson9010-spec:
## Affected crate(s)

- `ringbuf` (5,140,045 recent downloads per crates.io)

Fixed in [agerasev/ringbuf#60](https://github.com/agerasev/ringbuf/pull/60), released in 0.5.2.

## Severity

Panic-safety unsoundness in `Consumer::skip` / `clear`: the consumed elements are dropped before the read index is advanced, so a panicking element `Drop` leaves the read index pointing at already-dropped slots. The `LocalRb` destructor then drops them again — a double free (CWE-415) / use-after-free (CWE-416) reachable from safe Rust, confirmed under AddressSanitizer. Fixed in 0.5.2.

## Checklist

- [x] Asked maintainer(s) if publishing an advisory is appropriate (maintainer approved in the PR thread)


djc (approved):

djc:
Thanks!

PR #3246, opened 2026-09-22
Add advisory for stack_collections

OverwrittenCode:
## Affected crate(s)

- stack_collections (crates.io: https://crates.io/crates/stack_collections)

- https://github.com/OverwrittenCode/stack_collections/pull/4

## Severity

`StackVec::retain` committed its new length only after its internal loop
completed. If the retain predicate or a removed element's `Drop` panicked
before the loop finished, unwinding proceeded with the vector's length
still equal to the original length, leaving either a duplicated or an
already-destroyed element inside `0..len`. `StackVec`'s own `Drop` then
revisited that slot, causing a double-drop - and for any `T` owning a
heap allocation or other resource, a double-free.

Requires `T: Drop` with a predicate or destructor capable of panicking,
on builds with unwinding enabled (`panic = "unwind"`); not exploitable
under `panic = "abort"`.

I am the maintainer of stack_collections; this advisory is self-filed
following the fix merged in the linked PR.

## Checklist


LawnGnome (approved):
Thanks!

PR #3249, opened 2026-09-23
Add advisory for bun_collections: `SinglyLinkedList::remove` dereferences a null link

lwz23:
## Affected crate(s)

- `bun_collections` 0.2.1 (3,176 recent downloads per crates.io)

- https://github.com/putao520/bao/issues/46

## Severity

In versions before 0.2.1, `SinglyLinkedList::remove` is safe and walks the intrusive list with an unchecked dereference. On an empty list, or when `node` is not in the list, `(*current_elm).next` reads a null pointer. That is undefined behavior. The list head is a raw `*mut Node<T>`, and safe code can construct the empty list.

## Checklist


djc:
> * Asked maintainer(s) if publishing an advisory is appropriate

Please confirm with the maintainer whether they think an advisory makes sense for this (this goes for all of the advisories you submitted, especially because you're submitting in bulk).

Also note that we've recently adopted notability criteria:

- At least 1000 recent downloads, or
- At least 100.000 all-time downloads, or
- At least one reverse dependency from a different maintainer (team)

This seems to just barely pass those, at 3.2k recent downloads.

lwz23:
Asked the upstream maintainer whether publishing this advisory is appropriate: https://github.com/putao520/bao/issues/46

PR #3251, opened 2026-09-23
Add advisory for rpi-plugin-sdk: `register_entrypoint` was a safe function that derefer...

lwz23:
## Affected crate(s)

- `rpi-plugin-sdk` 0.1.24 (658 recent downloads per crates.io)

- https://github.com/bigfish1913/pi-rust/issues/7

## Severity

From 0.1.4 through 0.1.22, `register_entrypoint` was `pub fn` and took `api: *const PluginApiVt`. After a null check it did `&*api`. A non-null dangling or misaligned pointer is still undefined behavior, and safe Rust can pass one.

## Checklist


djc:
> * `rpi-plugin-sdk` 0.1.24 (658 recent downloads per crates.io)

Fails notability.

PR #3253, opened 2026-09-23
Add advisory for asm-lsp: `get_str` and `get_string` build a `str` without checking UTF-8

lwz23:
## Affected crate(s)

- `asm-lsp` 0.10.1 (3,617 recent downloads per crates.io)

- https://github.com/bergercookie/asm-lsp/issues/187

## Severity

`get_str` and `get_string` are safe. In 0.10.1 they still call `str::from_utf8_unchecked` on a caller-supplied `&[u8]` or `Vec<u8>`. A safety comment tells the caller to pass UTF-8, but the functions are not `unsafe`, so safe Rust can pass invalid bytes.

## Checklist


djc:
> `get_str` and `get_string` are safe. In 0.10.1 they still call `str::from_utf8_unchecked` on a caller-supplied `&[u8]` or `Vec<u8>`. A safety comment tells the caller to pass UTF-8, but the functions are not `unsafe`, so safe Rust can pass invalid bytes.

Since these guarantees are at least documented, I don't think this is worthy of an advisory.

PR #3254, opened 2026-09-23
Add advisory for autograd: Random-array constructors write into an uninitialized `Vec`

lwz23:
## Affected crate(s)

- `autograd` 2.0.0-rc3 (637 recent downloads per crates.io)

- https://github.com/raskr/rust-autograd/issues/65

## Severity

`gen_random_array` allocates `Vec::with_capacity(size)` and then, for every index, writes with `get_unchecked_mut` before `set_len`. Those slots are not initialized, so the write is undefined behavior. Public helpers such as `random_normal` call this function.

## Checklist


djc:
Fails notability criteria.

PR #3256, opened 2026-09-23
Add advisory for rllama: `data_u16_to_gpu` builds a slice from a raw pointer

lwz23:
## Affected crate(s)

- `rllama` 0.3.0 (10 recent downloads per crates.io)

- https://github.com/Noeda/rllama/issues/19

## Severity

`data_u16_to_gpu` is a safe public method. It takes `data: *const u16` and `nitems`, then calls `slice::from_raw_parts(data, nitems)` and uploads that slice.

## Checklist


djc:
Fails notability criteria.

PR #3260, opened 2026-09-23
Add advisory for capnp: `set_capability_pointer` dereferenced a raw pointer in a safe f...

lwz23:
## Affected crate(s)

- `capnp` 0.27.2 (2,451,496 recent downloads per crates.io)

- https://github.com/capnproto/capnproto-rust/issues/534
- https://github.com/capnproto/capnproto-rust/pull/563

## Severity

`set_capability_pointer` lived in `capnp::private` and was `pub fn`. It wrote through `reff: *mut WirePointer` with no null check. The module is public because generated code calls it, so downstream crates could call the safe function too.

## Checklist


djc:
> `set_capability_pointer` lived in `capnp::private` and was `pub fn`. It wrote through `reff: *mut WirePointer` with no null check. The module is public because generated code calls it, so downstream crates could call the safe function too.

Given that this is squared away within `private` and likely mostly used by generated code from the same workspace, I don't think this is worthy of an advisory.

PR #3263, opened 2026-09-23
Add advisory for j4rs: `to_rust_string` and `drop_c_string` were safe public FFI helpers

lwz23:
## Affected crate(s)

- `j4rs` 0.25.2 (20,801 recent downloads per crates.io)

- https://github.com/astonbitecode/j4rs/issues/138
- https://github.com/astonbitecode/j4rs/issues/140

## Severity

`to_rust_string` was `pub fn` and called `CStr::from_ptr` on a caller-supplied `*const c_char`. `drop_c_string` was `pub fn` and called `CString::from_raw`. Safe Rust could pass a null, dangling, or non-owned pointer.

## Checklist


djc:
Needs maintainer confirmation.

lwz23:
Asked the upstream maintainer whether publishing this advisory is appropriate: https://github.com/astonbitecode/j4rs/issues/138

lwz23:
Asked the upstream maintainer whether publishing this advisory is appropriate: https://github.com/astonbitecode/j4rs/issues/140

astonbitecode:
As discussed in the `j4rs` thread, the confusion was that `drop_c_string` was not marked as unsafe. The function was not exposed publicly and all required validations and correctness (for null pointers etc) were enforced by the calling code. 
We did not identify any potential wrong usage that could produce UB.

However, now things are in place, so, if there are other reasons this needs to be published, I am fine.

djc:
> As discussed in the `j4rs` thread, the confusion was that `drop_c_string` was not marked as unsafe. The function was not exposed publicly and all required validations and correctness (for null pointers etc) were enforced by the calling code.
> We did not identify any potential wrong usage that could produce UB.

Sounds like an advisory would not be appropriate, then. Thanks!

PR #3271, opened 2026-09-23
Add CVE alias for librsvg's RUSTSEC-2026-0305

federicomenaquintero:
Just an update now that the CVE has been assigned.


djc (approved):
Thanks!

Usually this happens automatically after a while:

https://github.com/rustsec/advisory-db/blob/main/.github/workflows/sync-ids.yml#L32

